Issue #5, opened 2020-01-02
Sentry properties multiple apps module in same project


nicolasjafelle:
I am testing Sentry crash report with the proguard plugin in our Android Apps.
My main concern if it is possible to have different `sentry.properties` in the same project.

Our project has two apps module but it seems the `sentry.properties` goes in the project root level:
```
project
--> sentry.properties
    --> app1
    --> app2
```

Is it possible to be like this?

```
project
    --> app1
    --> sentry.properties
    --> app2
    --> sentry.properties
```

Thanks in advance!

marandaneto:
Closing the issue as a part of large repository cleanup, due to it being inactive and/or outdated. There's a large chance that this issue has been solved since.
Please do not hesitate to ping me if it is still relevant, and I will happily reopen and work on it.
Cheers!

PR #33, opened 2020-10-28
Allow module level sentry properties file

MatthewTPage:
## Problem
The sentry gradle plugin does not seem to support multiple application modules in the same gradle project very cleanly. I read https://github.com/getsentry/sentry-android-gradle-plugin/issues/5 but the solution suggested there did not seem ideal.

## Solution
Allow the sentry.properties file to exist in all currently supported locations, as well as the module's directory.

Ie, both of the following paths would be supported:
```
.
├── app-1
├── app-2
└── sentry.properties

.
├── app-1
│   └── sentry.properties
└── app-2
    └── sentry.properties
```

## Verification

I built a snapshot of the gradle plugin and applied it to my own project (already set up to use sentry). I built release APKs of the app with `sentry.properties` in the root gradle directory, and another with it in the application gradle module directory Crashed the app intentionally on both APKs built, and in the sentry web UI I was able to see the crash events with the crashes properly de-obfuscated.

### DIFF
--- a/src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
+++ b/src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
@@ -25,7 +25,7 @@ class SentryPlugin implements Plugin<Project> {
      */
     static String getSentryCli(Project project) {
         // if a path is provided explicitly use that first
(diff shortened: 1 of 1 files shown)


marandaneto:
sounds good, thanks for the contribution @MatthewTPage 
we'll release this in the next version :)

PR #41, opened 2021-02-12
Ensure that the sentry debug properties are added to final APK

### LINKED ISSUE #37: Race condition when creating and packaging sentry-debug-meta.properties
_Android Gradle Plugin:_
**4.1.0**

_Gradle:_
**6.7**

_Optimizer:_
- [X]  R8

- [X]  Yes -> in project root

_SDKs:_
- [X]  sentry-android -> 3.1.3

---
I have the following issue:

Android crash reports are not deobfuscated on Sentry.io.
App log shows:
> I/Sentry: sentry-debug-meta.properties file was not found.

On clean build, the file is correctly created in `build/intermediates/merged_assets` folder but it's missing from the final apk.
On rebuild, the file is included in the apk.

**Steps to reproduce:**
- Run `gradlew clean assembleRelease` and open the final apk

**Actual result:**
- `sentry-debug-meta.properties` doesn't get packaged into apk's assets folder

**Expected result:**
- `sentry-debug-meta.properties` gets packaged into apk's assets folder

ansman:
To ensure that the asset file is added it needs to be available before
the proguard task runs. So instead of relying on the CLI to generate the
UUID there is now a separate task that creates it and --uuid is used in
the CLI to match the mappings to the build.

This fixes #37

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,8 @@ repositories {
 }
 
 compileGroovy {
-    sourceCompatibility = '1.6'
-    targetCompatibility = '1.6'
+    sourceCompatibility = '1.8'
+    targetCompatibility = '1.8'
 }
 
 dependencies {
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,8 @@ repositories {
 }
 
 compileGroovy {
-    sourceCompatibility = '1.8'
-    targetCompatibility = '1.8'
+    sourceCompatibility = '1.6'
+    targetCompatibility = '1.6'
 }
 
 dependencies {
--- a/src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
+++ b/src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
@@ -1,13 +1,18 @@
 package io.sentry.android.gradle;
 
 import org.gradle.api.DefaultTask;
+import org.gradle.api.Task;
+import org.gradle.api.Transformer;
+import org.gradle.api.file.Directory;
 import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.file.RegularFile;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.specs.Spec;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.OutputDirectory;
 import org.gradle.api.tasks.TaskAction;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.BufferedWriter;
 import java.io.FileOutputStream;
@@ -29,10 +34,21 @@ abstract class GenerateSentryProguardUuidTask extends DefaultTask {
     protected abstract Property<UUID> getOutputUuidInternal();
 
     @Internal
-    private final Provider<RegularFile> outputFile = getOutputDirectory().map(dir -> dir.file("sentry-debug-meta.properties"));
+    private final Provider<RegularFile> outputFile = getOutputDirectory().map(new Transformer<RegularFile, Directory>() {
+        @NotNull
+        @Override
+        public RegularFile transform(@NotNull Directory dir) {
+            return dir.file("sentry-debug-meta.properties");
+        }
+    });
 
     public GenerateSentryProguardUuidTask() {
-        getOutputs().upToDateWhen(spec -> false);
+        getOutputs().upToDateWhen(new Spec<Task>() {
+            @Override
+            public boolean isSatisfiedBy(Task spec) {
+                return false;
+            }
+        });
         setDescription("Generates a unique build ID");
     }
 
@@ -46,9 +62,12 @@ abstract class GenerateSentryProguardUuidTask extends DefaultTask {
         UUID uuid = UUID.randomUUID();
         getOutputUuidInternal().set(uuid);
 
-        try (Writer writer = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(outputFile.get().getAsFile())))) {
+        Writer writer = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(outputFile.get().getAsFile())));
+        try {
             writer.write("io.sentry.ProguardUuids=");
             writer.write(uuid.toString());
+        } finally {
+            writer.close();
         }
     }
 }
--- a/src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
+++ b/src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
@@ -1,0 +1,54 @@
+package io.sentry.android.gradle;
+
+import org.gradle.api.DefaultTask;
+import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.file.RegularFile;
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.Internal;
+import org.gradle.api.tasks.OutputDirectory;
+import org.gradle.api.tasks.TaskAction;
+
+import java.io.BufferedWriter;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.util.UUID;
+
+abstract class GenerateSentryProguardUuidTask extends DefaultTask {
+    @OutputDirectory
+    public abstract DirectoryProperty getOutputDirectory();
+
+    @Internal
+    public Provider<UUID> getOutputUuid() {
+        return getOutputUuidInternal();
+    }
+
+    @Internal
+    protected abstract Property<UUID> getOutputUuidInternal();
+
+    @Internal
+    private final Provider<RegularFile> outputFile = getOutputDirectory().map(dir -> dir.file("sentry-debug-meta.properties"));
+
+    public GenerateSentryProguardUuidTask() {
+        getOutputs().upToDateWhen(spec -> false);
+        setDescription("Generates a unique build ID");
+    }
+
+    @Internal
+    public Provider<RegularFile> getOutputFile() {
+        return outputFile;
+    }
+
+    @TaskAction
+    protected void generateProperties() throws IOException {
+        UUID uuid = UUID.randomUUID();
+        getOutputUuidInternal().set(uuid);
+
+        try (Writer writer = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(outputFile.get().getAsFile())))) {
+            writer.write("io.sentry.ProguardUuids=");
+            writer.write(uuid.toString());
+        }
+    }
+}
--- a/src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
+++ b/src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
@@ -1,5 +1,6 @@
 package io.sentry.android.gradle
 
+import com.android.Version
 import com.android.build.gradle.AppPlugin
 import com.android.build.gradle.LibraryPlugin
 import com.android.build.gradle.api.ApplicationVariant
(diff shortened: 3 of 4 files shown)


build.gradle
```diff
@@ -13,8 +13,8 @@ repositories {
 }
 
 compileGroovy {
-    sourceCompatibility = '1.6'
-    targetCompatibility = '1.6'
+    sourceCompatibility = '1.8'
+    targetCompatibility = '1.8'
```
ansman, line 17: This isn't strictly needed but should be fine and it makes the java code tidier

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
```diff
@@ -0,0 +1,54 @@
...
+import org.gradle.api.tasks.TaskAction;
+
+import java.io.BufferedWriter;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.util.UUID;
+
+abstract class GenerateSentryProguardUuidTask extends DefaultTask {
```
ansman, line 19: I wrote this in java in preparation for a Kotlin rewrite which feels like a well spent effort.

build.gradle
marandaneto: true, I'd keep as it is though due to the usage of our older Android SDK that runs on Java 6/7, removing the lambdas should be already enough.
We'd go fully with Kotlin in the new Plugin version, targeting Java 8 of course.

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
```diff
@@ -274,60 +243,48 @@ class SentryPlugin implements Plugin<Project> {
...
-                            name: persistIdsTaskName,
-                            type: Exec) {
-                        description "Write references to proguard UUIDs to the android assets."
-                        workingDir project.rootDir
-
-                        def propsFile = getPropsString(project, variant)
+                    def generateUuidTask = project.tasks.create(
+                            name: "generateSentryProguardUuid${variant.name.capitalize()}${variantOutput.name.capitalize()}",
+                            type: GenerateSentryProguardUuidTask) {
+                        outputDirectory.set(project.file("build/generated/assets/sentry/${variant.name}"))
```
marandaneto, line 249: previously we're not appending `/sentry`, wondering if this wouldn't break reading the file from the APK

build.gradle
ansman: Fair enough, though this only affects the build scripts and nothing else.

I'll revert 👍

Btw, I reached out on your personal Gmail address regarding the new plugin. Did you receive it or did it go to spam?

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @TaskAction
+    protected void generateProperties() throws IOException {
+        UUID uuid = UUID.randomUUID();
+        getOutputUuidInternal().set(uuid);
+
+        try (Writer writer = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(outputFile.get().getAsFile())))) {
+            writer.write("io.sentry.ProguardUuids=");
+            writer.write(uuid.toString());
+        }
```
marandaneto, line 52: I'm wondering if this would cause some problems like:

my mapping file has changed (not the same hash), but we didn't generate a new UUID and we've overwritten the mapping file on the server (if the server allows doing so).

or just the other way around, we've generated a new UUID, but the mapping file is still the same, so today `sentry-cli` takes care of it by computing the hash of the file, but right now we'd be doing this here instead, although not checking the hash at all.

https://github.com/getsentry/sentry-cli/blob/762ff34376a380789b3de731764836d50c7e0106/src/commands/upload_proguard.rs#L117-L131

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
```diff
@@ -274,60 +243,48 @@ class SentryPlugin implements Plugin<Project> {
...
-                                cli,
-                                "upload-proguard",
-                                "--write-properties",
-                                debugMetaPropPath,
-                                mappingFile
-                        ]
-
-                        if (!extension.autoUpload) {
-                            args << "--no-upload"
+                    if (VersionNumber.parse(Version.ANDROID_GRADLE_PLUGIN_VERSION) >= new VersionNumber(3, 3, 0, null)) {
```
marandaneto, line 256: this plugin should work across multiple Gradle and AGP versions, not sure if with the new changes, this still holds true, as the bug that this PR tries to solve starts happening on Gradle >= 6.7 or AGP >= 4.0, it'd be fine to bump a minor version and do the breaking change, people still could use older versions if they don't have the newest AGP/Gradle versions, but we'd need to be testing this fix across Gradle >= 6.7 and AGP >= 4.0, AGP probably up to 4.2, as 7.0 is very alpha.

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Internal
+    protected abstract Property<UUID> getOutputUuidInternal();
+
+    @Internal
+    private final Provider<RegularFile> outputFile = getOutputDirectory().map(dir -> dir.file("sentry-debug-meta.properties"));
+
+    public GenerateSentryProguardUuidTask() {
+        getOutputs().upToDateWhen(spec -> false);
```
marandaneto, line 35: does it mean each build would generate a new UUID?

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
```diff
@@ -274,60 +243,48 @@ class SentryPlugin implements Plugin<Project> {
...
+                            getSentryProject().set(buildTypeProperties.get(SENTRY_PROJECT_PARAMETER).toString())
                         }
-
-                        project.logger.info("args executed.")
-
-                        enabled true
                     }
 
+                    variant.register(uploadSentryProguardMappingsTask)
+                    project.android.sourceSets[variant.name].assets.srcDir(generateUuidTask.outputDirectory)
```
marandaneto, line 286: I guess we can ignore https://github.com/getsentry/sentry-android-gradle-plugin/pull/41/files#r577592017 because this outputs exactly to the root assets folder, no `sentry`

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
ansman: No, because the path is hooked up so that it uses the output directory.

In fact, you should have unique output directories/files for each task since it's used in up-to-date checking and if they overlap you can get cache issues. This is the same pattern KAPT uses for example:
```
build/generated/kapt/release
```

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
ansman: The first issue has already been addressed by the `outputs.upToDateWhen { false }`. This means that the task will never be up to date and every build will generate a new UUID.

The second part I can't answer since it depends on how the CLI/backend behaves. If things break when uploading the same mappings with a different UUID we can work around that by copying the mappings file and appending a comment that includes the UUID to ensure that the hash would be different.

ansman: However, reading the documentation for `--uuid` it does not say that you cannot use the same mappings for multiple builds.

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
ansman: Yeah, I agree. In general it feels better to support a smaller set of versions and just let people use the old versions. But for this change I did not want to break anything, though AGP 3.3 is pretty damn old.

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
ansman: Yeah, exactly. By default Gradle does not consider tasks up to date. However, if you add any `@Input[Directory|File]` or `@Output[Directory|File]` then Gradle will assume your task is up to date if the input and output is the same as in the last run. For this task we always want it to run when a new build is performed so we effectively "disable" the up to date checks.

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
ansman: I'm not sure what you mean. We are using the `outputDirectory` directory from the `generateUuidTask` task which is set to `project.file("build/generated/assets/sentry/${variant.name}")` so they will always be the same.

This is how new plugins and tasks should be written, using properties. By using properties you ensure that any changes propagate to all dependents.

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
marandaneto: I'll double-check with the `sentry-cli` team, reading Rust is not that easy hehe

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
marandaneto: I meant that we use the `project.android.sourceSets[variant.name].assets` directory (no `/sentry`), but I asked before about the `/sentry` dir cus I've written that comment before this one, so just ignore it, all good :)

build.gradle
marandaneto: hey, yep, I've gotten it and forwarded it to my Sentry email but I didn't have the time just yet, gonna do it within the next days, sorry about that, just wanna give more like a concrete direction instead of a short reply :)

marandaneto:
@ansman that's great, thanks for the contribution and figuring this out, I'll clear up the other questions with the sentry-cli team and see if we are fine, moving forward, I'd like to fix this bug asap, I'll run a few tests too.

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
ansman: Ah, I see. So `AndroidSourceSet.assets` is is not a directory but rather an `AndroidSourceDirectorySet` which is like a collection of directories so I'm rather adding the output directory to the source set.

In general when working with Gradle you don't want to add files to the APK/JAR/whatever by just adding files to the directory that some other task reads (such as the merged assets directory). The right way to do it is to add it to the source set and make the merge task depend on the task that generates the files.

ansman:
No problem! I talked a bit to the Android tooling team over at Google and they confirmed that there is no supported way to add assets after R8/proguard runs. The reason for this is that the assets are actually an input for the R8 task (for service loader special handling) so by the same R8 runs it is too late too add new assets. You might be able to make it work but it would be fragile and unsupported.

build.gradle
ansman: Ah, take your time! I just wanted to make sure you received it.

src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy
marandaneto: that makes total sense, thanks

src/main/groovy/io/sentry/android/gradle/GenerateSentryProguardUuidTask.java
marandaneto: @ansman this is currently blocked by https://github.com/getsentry/sentry-cli/issues/914
the situation is: we'll generate new UUIDs (for every build), if the mapping file has not changed (the hash itself), the UUID won't exist on our backend (because we won't upload it again, so the new UUID won't be valid) hence stack traces won't be demangled, we'll get somebody into it asap.

cortinico:
Closing as I've cherry-picked your commits in #62 @ansman 👍

PR #43, opened 2021-02-17
Add some basic functional tests

ansman:
This based on #41

This will serve as a base for a rewrite to ensure the behavior is still the same.


ansman:
There should probably be a Github workflow that runs this but I can't really do that on my branch.

build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+    compileOnly(localGroovy())
+    compileOnly("com.android.tools.build:gradle:4.1.0")
+
+    add("funcTestImplementation", gradleTestKit())
+    add("funcTestImplementation", kotlin("test"))
+    add("funcTestImplementation", kotlin("test-junit"))
+    add("funcTestImplementation", "org.jetbrains:annotations:19.0.0")
+}
+
+gradlePlugin {
```
ansman, line 124: By using the java gradle plugin instead of just including the metadata you get additional validations as well as better test support.

cortinico:
Closing as I've cherry-picked your commits in #62 @ansman 👍

PR #47, opened 2021-02-27
Added a CI Setup & Moved to Composite Build

cortinico:
I'm converting this project to use a composite build. The setup is inspired from:
https://github.com/cortinico/kotlin-gradle-plugin-template/

This will introduce a couple of nice features that will make handling the project a bit easier:
- Added two projects that test the plugin: `:examples:android-gradle` and `:examples:android-gradle-kts` that test the plugin setup on AGP 4.1
- Added a Github Actions workflow that will run on win/mac/linux and will try to run `persistSentryProguardUuidsForReleaseRelease` + all the checks/tests
- Added a check inside the `buildSrc` to automatically download the CLI only the first time you open the project.

The original project is untouched and just moved inside the `plugin-build` folder.

The missing piece here is the `sentry.properties` files that should be populated correctly with an API key. Do you have a test API key @marandaneto or can I ask you to set it up using a Github Actions secret?


examples/android-gradle-kts/sentry.properties
```diff
@@ -0,0 +1,3 @@
+defaults.project=your-project
+defaults.org=your-org
```
marandaneto, line 2:
```suggestion
defaults.project=sentry-android
defaults.org=sentry-sdks
```

examples/android-gradle/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+    buildTypes {
+        release {
+            minifyEnabled true
+        }
+    }
+}
+
+sentry {
+    autoProguardConfig = true
+    autoUpload = true
```
marandaneto, line 34: here could be omitted, it's enabled by default

examples/android-gradle-kts/sentry.properties
```diff
@@ -0,0 +1,3 @@
...
+auth.token=YOUR_AUTH_TOKEN
```
marandaneto, line 3: maybe instead of using a token, we could use the dryRun which is setting `autoUpload = false`
I'll check if I can set a token using Github Actions secret, I need to check with the security team.
cc @bruno-garcia

marandaneto: you could install the CLI locally and authenticate using https://docs.sentry.io/product/cli/configuration/

marandaneto:
I've created a branch `feat/gradle-plugin-v2` and targeted this PR into it, let's make this branch the base of all the other PRs, and later we do a merge without squash to keep history :)

examples/android-gradle/build.gradle
marandaneto: let's set it to `false` right till we figure out this -> https://github.com/getsentry/sentry-android-gradle-plugin/pull/47#discussion_r584158468

marandaneto:
> I'm converting this project to use a composite build. The setup is inspired from:
> https://github.com/cortinico/kotlin-gradle-plugin-template/

nice, I like this template pretty much.

> This will introduce a couple of nice features that will make handling the project a bit easier:
> 
> * Added two projects that test the plugin: `:examples:android-gradle` and `:examples:android-gradle-kts` that test the plugin setup on AGP 4.1
> * Added a Github Actions workflow that will run on win/mac/linux and will try to run `persistSentryProguardUuidsForReleaseRelease` + all the checks/tests
> * Added a check inside the `buildSrc` to automatically download the CLI only the first time you open the project.

nice :)

> The original project is untouched and just moved inside the `plugin-build` folder.
> 
> The missing piece here is the `sentry.properties` files that should be populated correctly with an API key. Do you have a test API key @marandaneto or can I ask you to set it up using a Github Actions secret?

suggestion is here -> https://github.com/getsentry/sentry-android-gradle-plugin/pull/47#discussion_r584158468

marandaneto (approved):
left a few comments on how we could go around the auth token for now, other than that, LGTM, thanks for doing this.

examples/android-gradle/build.gradle
cortinico: Updated 👍

marandaneto:
@cortinico feel free to merge when it's approved :)

examples/android-gradle-kts/sentry.properties
marandaneto: @cortinico I've created the `SENTRY_AUTH_TOKEN` GH repository secret, we'll need to change the value of `auth.token` before executing the tests

PR #50, opened 2021-03-03
Feat: Gradle plugin v2

marandaneto:


marandaneto:
@cortinico looks like https://github.com/getsentry/sentry-android-gradle-plugin/blob/feat/gradle-plugin-v2/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginTest.kt#L103 is failing for windows, we don't need to get blocked by this, just be aware :)

marandaneto:
@cortinico CI is not happy because of 

> Cannot set the value of extension 'sentry' property 'autoUpload' of type java.lang.Boolean using an instance of type java.lang.String.

examples/android-gradle/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+        release {
+            minifyEnabled true
+            proguardFiles getDefaultProguardFile(
+              'proguard-android-optimize.txt')
+        }
+    }
+}
+
+sentry {
+    autoUpload = System.getenv("AUTO_UPLOAD") ?: false
```
marandaneto, line 35: needs to convert to a boolean here probably

marandaneto, line 35:
```suggestion
    autoUpload = System.getenv("AUTO_UPLOAD").toBoolean() ?: false
```

marandaneto:
> throw new Error(`Cache upload failed because file read failed with ${error.message}`);

cortinico:
> throw new Error(`Cache upload failed because file read failed with ${error.message}`);

Those are false negatives from the caching action. I've retriggered them and they're green now.
I'll keep an eye on those if they happen again.

cerisier:
@marandaneto @cortinico Can I be of any help to get this merged :)

I'm running into issues because of current version of the plugin executing the mapping upload work as part of the configuration closure, effectively trying to upload mappings at every single builds for every single variants.

I was about to create a PR to fix just that but stumbled upon this epic PR which effectively solves this as well :)

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,5 @@
+object BuildPluginsVersion {
+    val AGP = System.getenv("VERSION_AGP") ?: "4.1.0"
```
marandaneto, line 2:
```suggestion
    val AGP = System.getenv("VERSION_AGP") ?: "4.1.3"
```

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -0,0 +1,5 @@
...
+    const val KOTLIN = "1.4.21"
```
marandaneto, line 3:
```suggestion
    const val KOTLIN = "1.4.32"
```

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Thu Oct 15 09:46:23 CEST 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-all.zip
```
marandaneto, line 3: should we try 7.0.0?

.github/workflows/test-publish-dry-run.yaml
```diff
@@ -0,0 +1,73 @@
...
+        name: ["stable"]
+        include:
+          - agp: "4.0.0"
+            gradle: "6.1.1"
+            experimental: false
+            name: AGP-4.0.0
+          - agp: "4.1.2"
+            gradle: "6.5"
+            experimental: false
+            name: AGP-4.1.2
```
marandaneto, line 30: lets add an entry for gradle 7.0.0

marandaneto:
> @marandaneto @cortinico Can I be of any help to get this merged :)
> 
> Right now, the plugin eagerly creates tasks immediatly for all variants, I was about to create a PR to fix just that but stumbled upon this epic PR which effectively solves this as well :)

hey @cerisier sorry late reply, we're about to release an alpha version by the end of the next week :)
it'd be awesome to get some validation, I'll keep you posted.

plugin-build/build.gradle.kts
```diff
@@ -0,0 +1,101 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    kotlin("jvm") version BuildPluginsVersion.KOTLIN
+    id("java-gradle-plugin")
+    id("maven-publish")
```
marandaneto, line 6: what's about using https://github.com/vanniktech/gradle-maven-publish-plugin
the benefit is, it handles sources, javadoc, signing and the usage of `closeAndReleaseRepository` so we don't have to click buttons

plugin-build/gradle.properties
```diff
@@ -6,4 +6,19 @@ version = 1.7.37-SNAPSHOT
...
+# publication pom properties
+POM_NAME=Sentry Android Gradle Plugin
+POM_DESCRIPTION=Sentry Android Gradle Plugin
+POM_URL=https://github.com/getsentry/sentry-android-gradle-plugin
+POM_SCM_URL=https://github.com/getsentry/sentry-android-gradle-plugin
+POM_SCM_CONNECTION=scm:git:git://github.com/getsentry/sentry-android-gradle-plugin.git
+POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/getsentry/sentry-android-gradle-plugin.git
+
+POM_LICENCE_NAME=MIT
+POM_LICENCE_URL=http://www.opensource.org/licenses/mit-license.php
```
bruno-garcia, line 20: What about changing the default branch to `main` when swap to the new code base?
Good chance to do that.

Meanwhile: 
```suggestion
POM_LICENCE_URL=https://github.com/getsentry/sentry-android-gradle-plugin/blob/master/LICENSE
```

cortinico: Yup absolutely 👌

marandaneto:
if it's green, we can merge to master :)

PR #51, opened 2021-03-04
Replicate .github/workflow on master branch

cortinico:
I'm replicating the GH Actions workflow on the `master` branch as apparently Github Actions is not picking up the workflows if they live only in a branch.


cortinico:
PreMerge checks will fail for `master`. I suggest we merge it as it is and they'll be green once we merge the other branch.

marandaneto:
> PreMerge checks will fail for `master`. I suggest we merge it as it is and they'll be green once we merge the other branch.

wfm

marandaneto (approved):

PR #52, opened 2021-03-04
Add a workflow to test publishing of mapping files

cortinico:
I'm adding a workflow to test end-to-end that the publishing works fine.
This workflow relies on the API token secret and works only on merge and PRs within the upstream repo.

For external PRs it won't run (as the secret won't be available) as it would just fail.

I've done a test run on my fork with a dummy token here: https://github.com/cortinico/sentry-android-gradle-plugin/runs/2034278777?check_suite_focus=true

### DIFF
--- a/.github/workflows/test-publish.yaml
+++ b/.github/workflows/test-publish.yaml
@@ -1,0 +1,40 @@
+name: Test Publishing UUID
+
+on:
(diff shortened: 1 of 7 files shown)


marandaneto (approved):
LGTM

PR #58, opened 2021-03-07
Setup a CI build matrix

cortinico:
I'm adding a `Test Publishing Dry Run` workflow that contains a matrix of AGP/Gradle versions.
This allows to test end to end that the two example projects are building fine on the specified Gradle/AGP version.

### DIFF
--- a/.github/workflows/test-publish-dry-run.yaml
+++ b/.github/workflows/test-publish-dry-run.yaml
@@ -1,0 +1,76 @@
+name: Test Publishing Dry Run
+
+on:
+  push:
+    branches:
+      - master
+      - feat/gradle-plugin-v2
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  publish-dry-run:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        agp: ["4.1.0"]
+        gradle: ["6.8.3"]
+        experimental: [false]
+        name: ["stable"]
+        include:
+          - agp: "3.4.3"
+            gradle: "6.1.1"
+            experimental: false
+            name: AGP-3.4.3
+          - agp: "3.5.4"
+            gradle: "6.1.1"
+            experimental: false
+            name: AGP-3.5.4
+          - agp: "3.6.4"
+            gradle: "6.1.1"
+            experimental: false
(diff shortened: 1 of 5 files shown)


cortinico:
Here the CI run: https://github.com/cortinico/sentry-android-gradle-plugin/actions/runs/629753202

build.gradle.kts
```diff
@@ -1,6 +1,8 @@
 plugins {
+    val agpVersion = System.getenv("VERSION_AGP") ?: BuildPluginsVersion.AGP
```
marandaneto, line 2: this line is duplicated across build scripts, lets create (Dependencies.kt) a method that does the same check and just call it from here, wdyt?

.github/workflows/test-publish-dry-run.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+jobs:
+  publish-dry-run:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        agp: ["4.1.0"]
+        gradle: ["6.8.3"]
+        experimental: [false]
```
marandaneto, line 20: what would be the use-case for this flag? I quite didn't understand.
if it's just because of `continue-on-error`, maybe we should use a different naming, like `continueOnError`

marandaneto (approved):
left 2 comments, but they are rather a suggestion and a question, feel free to merge it.
LGTM

build.gradle.kts
cortinico: Yup let's do it.

.github/workflows/test-publish-dry-run.yaml
cortinico: This allows you to define a matrix entry like this:
```
      matrix:
        agp: ["7.0.0-alpha09"]
        gradle: ["7.0-20210309230237+0000"]
        experimental: [true]
```

That will run on the latest AGP/Gradle version but will run as optional (so you can monitor what's happening on the latest versions).
It would be useful to enable it when Gradle 7 enters beta/RC cycle.

marandaneto: oh got it, so we allow experimental builds to fail and CI won't break, nice, thanks for explaining

PR #59, opened 2021-03-09
Fix failing Gradle tests on Windows 

cortinico:
Just a small test as tests are failing on Windows due to a path separator issue.
Here the run on my fork: https://github.com/cortinico/sentry-android-gradle-plugin/actions/runs/634024107


marandaneto (approved):
good catch

PR #61, opened 2021-03-12
Remove the autoProguard capability

cortinico:
Since the `.proguard` configuration is provided by the Sentry Android SDK `.aar`, there is no need to add an extra task to configure the proguard configuration. Hence I'm removing it.


bruno-garcia (approved):
👍

PR #62, opened 2021-03-13
Ensure that the sentry debug properties are added to final APK

### LINKED ISSUE #37: Race condition when creating and packaging sentry-debug-meta.properties

cortinico:
Original commit message:

```
To ensure that the asset file is added it needs to be available before
the proguard task runs. So instead of relying on the CLI to generate the
UUID there is now a separate task that creates it and --uuid is used in
the CLI to match the mappings to the build.
```

I've cherry picked the commits in both #41 and #43 plus I've converted most of the code to Kotlin.
As a side I've added some Unit and Integration tests for the two tasks class.

Here the CI run on my fork: https://github.com/cortinico/sentry-android-gradle-plugin/actions/runs/649050843

I noticed some flakyness with `ZipException: invalid stored block lengths` mostly related to how the APK is zipped/unzipped in the integration tests (only on the macos runner) I'd say we merge this and we follow up on a subsequent PR if it keeps on happening.

Fixes #37 
Closes #41 
Closes #43


.github/workflows/test-publish-dry-run.yaml
```diff
@@ -72,5 +72,14 @@ jobs:
...
 
-      - name: Run the persistSentryProguardUuids task
-        run: ./gradlew assembleRelease persistSentryProguardUuidsForReleaseRelease
+      - name: Run the uploadSentryProguardMappings task
+        run: ./gradlew assembleRelease uploadSentryProguardMappingsReleaseRelease
+
+      - name: Check sentry-debug-meta.properties inside APKs
+        run: |
+          unzip ./examples/android-gradle/build/outputs/apk/release/android-gradle-release-unsigned.apk -d output
+          cat output/assets/sentry-debug-meta.properties
```
marandaneto, line 81: does that mean if `cat` fails (probably file not found), the build fails? :) awesome

cortinico: Yes :)

marandaneto (approved):
LGTM

marandaneto:
> I noticed some flakyness with ZipException: invalid stored block lengths mostly related to how the APK is zipped/unzipped in the integration tests (only on the macos runner) I'd say we merge this and we follow up on a subsequent PR if it keeps on happening.

@cortinico sounds good, let's merge it and see how it goes, thanks

PR #63, opened 2021-03-14
Kotlinize the plugin Extension

cortinico:
I'm just updating the `SentryPluginExtension` to use Kotlin the Gradle Lazy API.

This is going to be a breaking change for consumer of the plugin on Gradle Kotlin DSL (as they'll have to `.set(true)` instead of ` = true`. Users on Gradle Groovy DSL won't be affected.


marandaneto (approved):
LGTM

PR #64, opened 2021-03-14
Fix flaky test failures for `ZipFile`

cortinico:
It seems that tests are failing on the `ZipFile` when trying to read an asset file from the APK.
I'm updating them to use `ZipInputStream` here.

A green run on CI is here: https://github.com/cortinico/sentry-android-gradle-plugin/actions/runs/651464179


marandaneto (approved):
LGTM

PR #44, opened 2021-02-18
Rewrite the plugin in Kotlin using the latest Gradle APIs and best practices

ansman:
Sorry for a PR out of the blue but I had some spare time. Hopefully this will be useful for you. Documentation are still TODO.

This is based on #43

PR #65, opened 2021-03-18
Export the Cli provider outside the plugin

cortinico:
Taking some of the work from #44, I've moved all the CLI logic to a separate file and rewritten it in Kotlin + I've added some tests.

Differently than #44 I haven't created a separate task for it as I think it would introduce more complexity.

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
@@ -1,0 +1,108 @@
+package io.sentry.android.gradle
+
+import org.gradle.api.Project
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.util.Properties
+
+internal object SentryCliProvider {
+    /**
+     * Return the correct sentry-cli executable path to use for the given project.  This
+     * will look for a sentry-cli executable in a local node_modules in case it was put
+     * there by sentry-react-native or others before falling back to the global installation.
+     */
+    @JvmStatic
+    fun getSentryCliPath(project: Project): String {
+        // If a path is provided explicitly use that first.
+        project.logger.info("[sentry] Searching cli from sentry.properties file...")
+
+        searchCliInPropertiesFile(project)?.let {
+            project.logger.info("[sentry] cli Found: $it")
+            return@getSentryCliPath it
+        } ?: project.logger.warn("[sentry] sentry-cli not found in sentry.properties file")
+
+        // next up try a packaged version of sentry-cli
+        val cliSuffix = getCliSuffix()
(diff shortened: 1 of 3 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -0,0 +1,127 @@
...
+        project.logger.info("[sentry] Searching cli from sentry.properties file...")
+
+        searchCliInPropertiesFile(project)?.let {
+            project.logger.info("[sentry] cli Found: $it")
+            return@getSentryCliPath it
+        } ?: project.logger.warn("[sentry] sentry-cli not found in sentry.properties file")
+
+        // in case there is a version from npm right around the corner use that one.
+        // This is the case for react-native-sentry for instance.
+        project.logger.info("[sentry] Searching cli inside ../node_modules/...")
```
marandaneto, line 27: I'm not entirely sure and need to confirm, but I guess all the searching for react-native folders (node_modules) are not used anymore, the RN SDK has its own Gradle tasks now, we don't need to keep compatibility anymore.
Also, this plugin already ships sentry-cli so it's not needed to look for one unless there's a special need I'm not aware of.

https://github.com/getsentry/sentry-react-native/blob/master/sentry.gradle

@bruno-garcia @jennmueng can you confirm that?

marandaneto: I've confirmed it, we could remove this part of searching for sentry-cli over node_modules, which means killing this use case, removing the method `searchCliInNodeModules`, etc...

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    internal fun getCliSuffix(): String {
+        val osName = System.getProperty("os.name").toLowerCase()
+        val osArch = System.getProperty("os.arch")
+        return when {
+            "mac" in osName -> "Darwin-x86_64"
+            "linux" in osName -> if (osArch == "amd64") "Linux-x86_64" else "Linux-$osArch"
+            "win" in osName -> "Windows-i686.exe"
+            else -> ""
```
marandaneto, line 124: should we return `null` instead? what's the advantage of returning empty if we check `isNotBlank`

marandaneto:
> I haven't created a separate task for it as I think it would introduce more complexity.

makes sense to me too

marandaneto (approved):
just a comment to clean it up a bit of ded code https://github.com/getsentry/sentry-android-gradle-plugin/pull/65/files#r597458417

other than that, LGTM

PR #66, opened 2021-03-20
Fix broken test on cli from resources 

cortinico:
Turns out there are some broken tests as the `resources/bin/` folder was excluded by the .gitignore. I'm renaming it to `dummy-bin` so tests should be green after this.

### DIFF
--- a/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
+++ b/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
@@ -117,41 +117,51 @@ class SentryCliProviderTest {
 
     @Test
     fun `searchCliInResources finds the file correctly`() {
-        val resourcePath = "/testFixtures/bin/dummy-sentry-cli"
+        val resourcePath = "./dummy-bin/dummy-sentry-cli"
+        val resourceFile = javaClass.getResource(".")
+            ?.let { File(it.file, resourcePath) }
+            ?.apply {
+                parentFile.mkdirs()
+                createNewFile()
+            }
 
         val foundPath = searchCliInResources(resourcePath)
         assertNotNull(foundPath)
-        assertTrue(foundPath.endsWith(resourcePath))
+        assertTrue(foundPath.endsWith("/dummy-bin/dummy-sentry-cli"))
+
+        resourceFile?.delete()
     }
 
     @Test
     fun `searchCliInResources returns null if file does not exist`() {
-        val resourcePath = "/testFixtures/bin/i-dont-exist"
+        val resourcePath = "./dummy-bin/i-dont-exist"
 
         assertNull(searchCliInResources(resourcePath))
     }
 
     @Test
     fun `loadCliFromResourcesToTemp finds the file correctly`() {
-        val resourcePath = "/testFixtures/bin/dummy-sentry-cli"
+        val resourcePath = "./dummy-bin/dummy-sentry-cli"
+        val resourceFile = javaClass.getResource(".")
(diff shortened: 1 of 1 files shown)


marandaneto:
wondering why I'm not getting assigned automatically for review.

https://github.com/getsentry/sentry-android-gradle-plugin/blob/master/.github/CODEOWNERS

maybe `feat/gradle-plugin-v2` isn't up to date with `master`?

plugin-build/src/test/resources/testFixtures/dummy-bin/dummy-sentry-cli
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+echo "This is just a dummy script"
```
marandaneto, line 2: should we create and delete on exit (per test or test class) instead of adding this file to the repo?

marandaneto (approved):
left a suggestion but LGTM

plugin-build/src/test/resources/testFixtures/dummy-bin/dummy-sentry-cli
cortinico: Yup fixed

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
```diff
@@ -117,41 +117,51 @@ class SentryCliProviderTest {
...
 
         val loadedPath = loadCliFromResourcesToTemp(resourcePath)
         assertNotNull(loadedPath)
 
         val binContent = File(loadedPath).readText()
-        assertEquals(
-            """
-            #!/bin/bash
-            echo "This is just a dummy script"
+        assertEquals("echo \"This is just a dummy script\"", binContent)
```
marandaneto, line 157: I feel that this test already covers if the file exists and its content, hence the test `searchCliInResources finds the file correctly` is not useful anymore, I'd suggest removing it

marandaneto:
@cortinico a tip for other PRs, we already use squash & merge, so no need to force-push :)
I actually like to see the commits so I can see what you've changed after review, instead of trying to find out the changed pieces, wdyt?

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
cortinico: I'd actually keep both as they're testing two different APIs: `searchCliInResources` vs `loadCliFromResourcesToTemp`. Agree that there is a lot of overlapping, but the two function should be tested in isolation.

If we decide to remove one of the two function, we can also remove the related test.

cortinico:
> @cortinico a tip for other PRs, we already use squash & merge, so no need to force-push :)
> I actually like to see the commits so I can see what you've changed after review, instead of trying to find out the changed pieces, wdyt?

Yup sorry 😅 I get confused with too many OSS projects and different git flows.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
marandaneto: indeed, totally makes sense, I missed that part (testing search in resources or temp)

marandaneto:
> > @cortinico a tip for other PRs, we already use squash & merge, so no need to force-push :)
> > I actually like to see the commits so I can see what you've changed after review, instead of trying to find out the changed pieces, wdyt?
> 
> Yup sorry 😅 I get confused with too many OSS projects and different git flows.

haha no worries, totally relate that

PR #68, opened 2021-03-25
Export Task Providers to a separate class

cortinico:
Moved all the task providers to a separate file and added test coverage for them.
Plus I've updated all the parameters from `ApplicationVariant` to `String` so become trivial to test.

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
@@ -1,0 +1,88 @@
+package io.sentry.android.gradle
+
+import com.android.build.gradle.api.ApplicationVariant
+import org.gradle.api.Project
+import org.gradle.api.Task
+
+internal object SentryTasksProvider {
+
+    /**
+     * Returns the transformer task for the given project and variant.
+     * It could be either ProGuard or R8
+     *
+     * @return the task or null otherwise
+     */
+    @JvmStatic
+    fun getTransformerTask(project: Project, variantName: String): Task? =
+        project.findTask(
+            // Android Studio 3.3 includes the R8 shrinker.
+            "transformClassesAndResourcesWithR8For${variantName.capitalized}",
+            "transformClassesAndResourcesWithProguardFor${variantName.capitalized}",
+            "minify${variantName.capitalized}WithR8",
+            "minify${variantName.capitalized}WithProguard"
+        )
+
+    /**
+     * Returns the dex task for the given project and variant.
+     *
+     * @return the task or null otherwise
+     */
+    @JvmStatic
+    fun getDexTask(project: Project, variantName: String): Task? =
+        project.findTask(
+            // Android Studio 3.3 includes the R8 shrinker.
+            "transformClassesWithDexFor${variantName.capitalized}",
+            "transformClassesWithDexBuilderFor${variantName.capitalized}",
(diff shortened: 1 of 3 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
```diff
@@ -0,0 +1,88 @@
...
+     * It could be either ProGuard or R8
+     *
+     * @return the task or null otherwise
+     */
+    @JvmStatic
+    fun getTransformerTask(project: Project, variantName: String): Task? =
+        project.findTask(
+            // Android Studio 3.3 includes the R8 shrinker.
+            "transformClassesAndResourcesWithR8For${variantName.capitalized}",
+            "transformClassesAndResourcesWithProguardFor${variantName.capitalized}",
```
marandaneto, line 20: I wonder if this is still necessary if we support AGP 4.0 and up only

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @return the task if found or null otherwise
+     */
+    @JvmStatic
+    fun getAssembleTask(project: Project, variant: ApplicationVariant): Task {
+        return try {
+            variant.assembleProvider.get()
+        } catch (ignored: Exception) {
+            project.logger.error("[sentry] .assembleProvider failed with: ${ignored.message}")
+            variant.assemble
```
marandaneto, line 80: I wonder if this is still necessary if we support AGP 4.0 and up only

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryTaskProviderTest.kt
```diff
@@ -0,0 +1,164 @@
...
+    fun `getTransformerTask returns null for missing task`() {
+        val project = ProjectBuilder.builder().build()
+
+        assertNull(getTransformerTask(project, "debug"))
+    }
+
+    @Test
+    fun `getTransformerTask returns transform for R8`() {
+        val project = ProjectBuilder.builder().build()
+        val task = project.tasks.register("transformClassesAndResourcesWithR8ForDebug")
```
marandaneto, line 32: these 2 lines are repeated across all the tests, maybe let's add a method to not DRY?

`registerTask(name: String): Tuple<Project, Task>` or something like that

what do you think?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
cortinico: True, thought the CI matrix runs on AGP 3.4 and above so is probably worth to keep.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
cortinico: True, thought the CI matrix runs on AGP 3.4 and above so is probably worth to keep.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryTaskProviderTest.kt
cortinico: Yup added 👍

marandaneto (approved):
LGTM

PR #72, opened 2021-04-13
Export MappingFileProvider

cortinico:
This PR converts the `getMappingFile` to Kotlin + adds tests for it.

Moreover, I had to introduce the `SystemPropertyRule`. The reason is that changing properties with `System.setProperty` was causing property values to be retained, hence adding flakiness to tests.

Here a green CI run: https://github.com/cortinico/sentry-android-gradle-plugin/runs/2331188077

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryMappingFileProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryMappingFileProvider.kt
@@ -1,0 +1,30 @@
+package io.sentry.android.gradle
+
+import com.android.build.gradle.api.ApplicationVariant
+import org.gradle.api.Project
+import java.io.File
+
+internal object SentryMappingFileProvider {
+
+    /**
+     * Returns the obfuscation mapping file (might be null if the obfuscation is disabled).
+     *
+     * @return the mapping file or null if not found
+     */
+    @JvmStatic
+    fun getMappingFile(project: Project, variant: ApplicationVariant): File? =
+        try {
+            val mappingFiles = variant.mappingFileProvider.get().files
+            if (mappingFiles.isEmpty()) {
+                project.logger.warn("[sentry] .mappingFileProvider.files is empty for ${variant.name}")
+                null
+            } else {
+                project.logger.info("[sentry] Mapping File ${mappingFiles.first()} for ${variant.name}")
+                mappingFiles.first()
+            }
+        } catch (ignored: Throwable) {
+            project.logger.error("[sentry] .mappingFileProvider failed with: ${ignored.message}")
+            ignored.printStackTrace()
+            variant.mappingFile
+        }
+}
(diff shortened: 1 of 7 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryMappingFileProvider.kt
```diff
@@ -0,0 +1,30 @@
...
+                project.logger.warn("[sentry] .mappingFileProvider.files is empty for ${variant.name}")
+                null
+            } else {
+                project.logger.info("[sentry] Mapping File ${mappingFiles.first()} for ${variant.name}")
+                mappingFiles.first()
+            }
+        } catch (ignored: Throwable) {
+            project.logger.error("[sentry] .mappingFileProvider failed with: ${ignored.message}")
+            ignored.printStackTrace()
+            variant.mappingFile
```
marandaneto, line 28: in theory we don't need to support that anymore

cortinico: Yup but then the AGP matrix will fail for 3.x. I'd say let's remove the 3.x support after the release of plugin v2

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryMappingFileProviderTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        // This forces the project to be evaluated
+        project.getTasksByName("assembleDebug", false)
+
+        val debugVariant = android.applicationVariants.first { it.name == "debug" }
+        assertTrue { getMappingFile(project, debugVariant)!!.endsWith("build/outputs/mapping/debug/mapping.txt") }
+
+        val releaseVariant = android.applicationVariants.first { it.name == "release" }
+        assertTrue { getMappingFile(project, releaseVariant)!!.endsWith("build/outputs/mapping/release/mapping.txt") }
```
marandaneto, line 32: let's replace every `/` or `\` to `File.separator` instead, it's safer for file separators

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryMappingFileProvider.kt
marandaneto: yeah just have seen the CI.
so either we keep support for AGP 3.x or we drop `before` the release of Plugin v2, as we are doing a major bump, it's fine to do such breaking change, but after releasing would require another major bump

marandaneto (approved):
left a few suggestions, but looks good, a nice approach with the `WithSystemProperty` annotation.
LGTM

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryMappingFileProviderTest.kt
```diff
@@ -0,0 +1,52 @@
...
+                    buildType.setMinifyEnabled(true)
+                }
+            }
+        }
+
+        // This forces the project to be evaluated
+        project.getTasksByName("assembleDebug", false)
+
+        val debugVariant = android.applicationVariants.first { it.name == "debug" }
+        assertTrue { getMappingFile(project, debugVariant)!!.endsWith("build/outputs/mapping/debug/mapping.txt") }
```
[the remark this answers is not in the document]
cortinico: Yes it does as the build types are configured with `.all`:

```
buildTypes {
                it.all { buildType ->
                    buildType.setMinifyEnabled(true)
                }
            }
```

PR #73, opened 2021-04-15
Export PropertiesFileProvider

cortinico:
Converting another section of the plugin to Kotlin 🎉 
Some of the paths used for search of the properties file are debatable (specifically those on the root projects). I can't seem to figure out why someone would place the property there. @marandaneto do you have some context?

The rest of the PR is mostly tests.

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPropertiesFileProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPropertiesFileProvider.kt
@@ -1,0 +1,53 @@
+package io.sentry.android.gradle
+
+import com.android.build.gradle.api.ApplicationVariant
+import org.gradle.api.Project
+import java.io.File
+
+internal object SentryPropertiesFileProvider {
+
+    private const val FILENAME = "sentry.properties"
+
+    /**
+     * Find sentry.properties and returns the path to the file.
+     * @param project the given project
+     * @param variant the given variant
+     * @return A [String] for the path if sentry.properties is found or null otherwise
+     */
+    @JvmStatic
+    fun getPropertiesFilePath(project: Project, variant: ApplicationVariant): String? {
+        val flavorName = variant.flavorName
+        val buildTypeName = variant.buildType.name
+
+        val projDir = project.projectDir
+        val rootDir = project.rootDir
+
+        // Local Project dirs
+        val possibleFiles = mutableListOf(
+            "$projDir/src/$buildTypeName/$FILENAME"
+        )
+        if (flavorName.isNotBlank()) {
+            possibleFiles.add("$projDir/src/$buildTypeName/$flavorName/$FILENAME")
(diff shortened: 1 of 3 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPropertiesFileProvider.kt
```diff
@@ -0,0 +1,53 @@
...
+    private const val FILENAME = "sentry.properties"
+
+    /**
+     * Find sentry.properties and returns the path to the file.
+     * @param project the given project
+     * @param variant the given variant
+     * @return A [String] for the path if sentry.properties is found or null otherwise
+     */
+    @JvmStatic
+    fun getPropertiesFilePath(project: Project, variant: ApplicationVariant): String? {
```
marandaneto, line 18: let's use the File.separator API instead of `/`

marandaneto:
@cortinico no idea, its past me so I'm not sure, maybe because the differences across Gradle/AGP versions or even hybrid SDKs that also need the plugin eg a react native app

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPropertiesFileProviderTest.kt
```diff
@@ -0,0 +1,239 @@
...
+
+    private val sep = File.separator
+
+    @Test
+    fun `getPropertiesFilePath finds file inside debug folder`() {
+        val project = ProjectBuilder.builder().build()
+        project.plugins.apply("com.android.application")
+        val android = project.extensions.getByType(AppExtension::class.java).apply {
+            compileSdkVersion(30)
+        }
```
marandaneto, line 21: looks like this block repeats across tests, maybe we create a method that returns a `project`?
it might require a param with a default value but allows customizing the `AppExtension` as some tests require only `compileSdkVersion(30)` and others with flavors and so on

marandaneto (approved):
left 2 comments to be addressed, but pre-approving as the PR looks good, thanks @cortinico

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPropertiesFileProvider.kt
cortinico: Yup updated in tests as well. Is not super elegant though (and also not really needed when dealing with Gradle API). Though for the future we should probably clean this up and use `Path` instead.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPropertiesFileProviderTest.kt
cortinico: Yup added a:
```
    private fun createTestAndroidProject(
        parent: Project? = null,
        block: AppExtension.() -> Unit = {}
    ): Pair<Project, AppExtension> {
```

That will do the job 👌

PR #76, opened 2021-04-18
Export UploadNativeSymbols to its own task

cortinico:
The `uploadNativeSymbols` task had his task-action inlined in the plugin apply function. I'm exporting it to a proper Task class and rewriting it in Kotlin here.

### DIFF
--- a/plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/SentryUploadNativeSymbolsTaskTest.kt
+++ b/plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/SentryUploadNativeSymbolsTaskTest.kt
@@ -1,7 +1,6 @@
 package io.sentry.android.gradle.tasks
 
 import org.gradle.api.Project
-import org.gradle.api.tasks.TaskProvider
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.Test
 import java.io.File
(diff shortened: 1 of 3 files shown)


plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/SentryUploadNativeSymbolsTaskTest.kt
```diff
@@ -0,0 +1,114 @@
...
+class SentryUploadNativeSymbolsTaskTest {
+
+    @Test
+    fun `cli-executable is set correctly`() {
+        val project = createProject()
+        val task: TaskProvider<SentryUploadNativeSymbolsTask> =
+            project.tasks.register("testUploadNativeSymbols", SentryUploadNativeSymbolsTask::class.java) {
+                it.cliExecutable.set("sentry-cli")
+                it.includeNativeSources.set(false)
+                it.variantName.set("debug")
```
marandaneto, line 22: maybe we create an aux. method like `createProject` but for the task? this block seems kinda duplicated across tests, if possible I'd extract and use default param values so we reuse it when possible.

marandaneto (approved):
left a suggestion but feel free to judge if it makes sense.

thanks, LGTM.

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/SentryUploadNativeSymbolsTaskTest.kt
cortinico: Yup makes sense 👍

PR #78, opened 2021-04-24
Bump multiple dependencies

### LINKED ISSUE #70: Sentry Gradle Plugin not compatible with Gradle 7.0
_Gradle:_
7.0

---
I have the following issue:

Gradle throws an error because `SentryProguardConfigTask` is not properly annotated: 

```
:app:addSentryProguardSettingsForRelease FAILED	
FAILURE: Build failed with an exception.	
* What went wrong:	
A problem was found with the configuration of task ':app:addSentryProguardSettingsForRelease' (type 'SentryProguardConfigTask').	
  - Type 'SentryProguardConfigTask' property 'applicationVariant' is missing an input or output annotation.	
    Reason: A property without annotation isn't considered during up-to-date checking.	
    Possible solutions:	
      1. Add an input or output annotation.	
      2. Mark it as @Internal.	
    Please refer to https://docs.gradle.org/7.0-rc-2/userguide/validation_problems.html#missing_annotation for more details about this problem.	
```
There might be other tasks that also require annotation, I didn't check. 

**Steps to reproduce:**
just run a release build (`gw :app:assembleRelease`) using Gradle 7.0

**Actual result:**
failure

**Expected result:**
success

cortinico:
Doing a cleanup and bump of dependencies after the Kotlin conversion:
- Kotlin to 1.4.32
- AGP to 4.1.3
- Gradle to 7.0
- Junit 4.13.2

I've also converted the inner build.gradle to use Kotlin DSL.
The bump to Gradle 7.0 forced me to remove `com.bmuschko.nexus` as it was unsupported. I switched to use the native `maven-publish` plugin to customize the pom.

The CI matrix was also updated to use AGP 4.2 RC and I've added a entry for Gradle 7.0

I also realized that KtLint was not applied correctly to the inner build. This is now fixed and I re-run KtLint on the whole plugin project.

Fixes #70


cortinico:
Here a green CI run https://github.com/cortinico/sentry-android-gradle-plugin/actions/runs/780098401

plugin-build/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    kotlin("jvm") version "1.4.32"
```
marandaneto, line 4: maybe we move all the names and versions to `BuildPluginsVersion`?

marandaneto (approved):
awesome, that's pretty good.
left a suggestion for DRY names and versions, but other than that, LGTM

plugin-build/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+    filter {
+        exclude("**/generated/**")
+        include("**/kotlin/**")
+    }
+}
+
+/* ktlint-disable max-line-length */
+publishing {
+    publications {
+        this.withType(MavenPublication::class.java) {
```
marandaneto, line 59: do we need:

```
java {
    withJavadocJar()
    withSourcesJar()
}
```

or something like that? otherwise sources and javadocs wont be generated I guess so publishing to MC would fail

plugin-build/build.gradle.kts
cortinico: Yup. I had to share the `buildSrc` between the two builds are they're isolated by default.

plugin-build/build.gradle.kts
cortinico: Added. That's actually not a requirement for Maven Central.
What is instead a requirement is the GPG signing that is not set up here.

marandaneto: are you sure? I remember that having javadoc and sources are mandatory files to publish on maven central, they could be empty, but they need to exist.

cortinico: > are you sure?

You were right, they're actually a requirement :) TIL

PR #79, opened 2021-04-26
Setup publishing to Maven Central

cortinico:
This sets up the publishing to Sonatype using the `gradle-nexus.publish-plugin` plugin.

### DIFF
--- a/plugin-build/build.gradle.kts
+++ b/plugin-build/build.gradle.kts
@@ -4,6 +4,8 @@ plugins {
     kotlin("jvm") version BuildPluginsVersion.KOTLIN
     id("java-gradle-plugin")
     id("maven-publish")
+    id("signing")
+    id("io.github.gradle-nexus.publish-plugin") version BuildPluginsVersion.NEXUS
     id("org.jlleitschuh.gradle.ktlint") version BuildPluginsVersion.KTLINT
 }
 
@@ -58,6 +60,20 @@ ktlint {
     }
 }
 
+nexusPublishing {
+    repositories {
+        // Maven Central (Sonatype) login info go to:
+        // ~/.gradle/gradle.properties
+        //
+        // mavenCentralRepositoryUsername=user name
+        // mavenCentralRepositoryPassword=password
+        sonatype {
+            username.set(findProperty("mavenCentralRepositoryUsername") as? String)
+            password.set(findProperty("mavenCentralRepositoryPassword") as? String)
+        }
+    }
+}
+
 /* ktlint-disable max-line-length */
 publishing {
     publications {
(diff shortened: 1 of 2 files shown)


plugin-build/build.gradle.kts
```diff
@@ -4,6 +4,8 @@ plugins {
     kotlin("jvm") version BuildPluginsVersion.KOTLIN
     id("java-gradle-plugin")
     id("maven-publish")
+    id("signing")
+    id("io.github.gradle-nexus.publish-plugin") version BuildPluginsVersion.NEXUS
```
marandaneto, line 8: any reason or advantages over this plugin and https://github.com/vanniktech/gradle-maven-publish-plugin as suggested?

cortinico: I'm unsure about the differences. I used to use https://github.com/marcphilipp/nexus-publish-plugin that evolved into `io.github.gradle-nexus.publish-plugin`

cortinico: We can swap it if needed, no strong opinion on this.

bruno-garcia: @cortinico are you aware of any plugin or CLI that can just take the files already compiled and upload them instead?

SDK release at Sentry is split in two phases: 

> Prepare

This will just build, run tests and package the artifact

> Publish

This step, potentially done by someone else (with different rights) would just get the artifact builds on the first step, and push them to the package registry.

Currently our Java and Android flow relies on plugins that need to build/package/publish in one go, and that makes our lifes a bit harder.

Any pointers here would help us out a lot.

cortinico: > @cortinico are you aware of any plugin or CLI that can just take the files already compiled and upload them instead?

This should do the job: https://github.com/martinbonnin/vespene

cortinico: @marandaneto I've update the publishing plugin here 77b2611 
However I was not able to let the signing work. I believe is due to https://github.com/vanniktech/gradle-maven-publish-plugin/issues/228 but I haven't investigated further. Don't you mind giving it a try, otherwise we can just revert 77b2611

marandaneto: @cortinico using `com.vanniktech.maven.publish` you dont need:

```
java {
    withJavadocJar()
    withSourcesJar()
}
```

you don't need to apply `signing`, it goes here https://github.com/getsentry/sentry-java/blob/main/build.gradle.kts#L92-L112
yo udon't need the `publications` block, it goes here -> https://github.com/getsentry/sentry-java/blob/main/gradle.properties#L31-L44

you need:

and dokka with default conf. https://github.com/getsentry/sentry-java/blob/main/build.gradle.kts#L26

cortinico: Updated 👍 So I've tried to sign locally and it works on my machine. I had to specifically set the `secretKeyRingFile` as the plugin was not picking it up from my gpg-agent :/

marandaneto: nice, still missing the conditional signing as stated before, see https://github.com/getsentry/sentry-java/blob/main/build.gradle.kts#L92-L112

marandaneto: @cortinico indeed, you need to specify `secretKeyRingFile`
its described in the shared link above though https://github.com/getsentry/sentry-java/blob/main/build.gradle.kts#L108

cortinico: > nice, still missing the conditional signing as stated before,

Updated. I've added only a check on the CI env variable as the plugin is already filtering the `-SNAPSHOT` versions.

PR #80, opened 2021-04-29
Conditionally apply maven.publish

cortinico:
We conditionally apply the maven.publish plugin only if we're on Gradle 6.6.0+
as such plugin is not compatible with lower versions of Gradle and will break
the CI matrix.

Here a green run: https://github.com/cortinico/sentry-android-gradle-plugin/actions/runs/795218822


marandaneto (approved):

marandaneto:
niice thanks @cortinico

PR #81, opened 2021-04-29
Fix test: use / instead of File.separator for class.getResources

marandaneto:
_#skip-changelog_

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
@@ -27,9 +27,7 @@ internal object SentryCliProvider {
         project.logger.info("[sentry] cliSuffix is $cliSuffix")
 
         if (!cliSuffix.isNullOrBlank()) {
-            val sep = File.separator
-
-            val resourcePath = "{$sep}bin{$sep}sentry-cli-$cliSuffix"
+            val resourcePath = "/bin/sentry-cli-$cliSuffix"
 
             // if we are not in a jar, we can use the file directly
             project.logger.info("[sentry] Searching for $resourcePath in resources folder...")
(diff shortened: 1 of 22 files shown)


cortinico (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -27,7 +27,9 @@ internal object SentryCliProvider {
         project.logger.info("[sentry] cliSuffix is $cliSuffix")
 
         if (!cliSuffix.isNullOrBlank()) {
-            val resourcePath = "/bin/sentry-cli-$cliSuffix"
+            val sep = File.separator
+
+            val resourcePath = "{$sep}bin{$sep}sentry-cli-$cliSuffix"
```
marandaneto, line 32: @cortinico wondering if it should be `File.separator` for the java resources or if it's OS independent.
most tests are failing, but the test for this use case does use `File.separator`, ideas? ^^

cortinico: I believe we should migrate the whole thing to use `kotlin.io.path.Path`. It's experimental in Kotlin 1.4 but allows us to get rid of these kind of problems. As an alternative we could use `Path` from nio directly.

marandaneto: yeah, makes sense, but I guess it's failing because the Java resource path is OS independent, maybe?

marandaneto: https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getResource(java.lang.String)

`getResource` requires `/`, so I am just removing `File.separator` from the test instead of fixing it here

marandaneto:
closing this, the current impl. is already correct, my bad

cortinico:
I'll follow up on those + all the other occurrencies to use `Path` from java.nio in another PR

Issue #84, opened 2021-04-29
Plugin eagerly creates tasks for all variants


cerisier:
_Android Gradle Plugin:_
**4.1.3**

_Gradle:_
**6.7.1**

_Optimizer:_
- [ ]  ProGuard
- [x]  R8

_Have you created the sentry.properties file?_
- [x]  Yes -> where?
- [ ]  No

_SDKs:_
- [x]  sentry-android-gradle-plugin -> 2.0.0-alpha.2

---
I have the following issue:

Plugin eagerly creates tasks for all variants.
```
> Configure project :app
[sentry] .mappingFileProvider is missing for com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@6626fee8 - Error: Cannot query the value of this property because it has no value available.
[sentry] .mappingFileProvider is missing for com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@57da97c8 - Error: Cannot query the value of this property because it has no value available.
[sentry] .mappingFileProvider is missing for com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@67849248 - Error: Cannot query the value of this property because it has no value available.
[sentry] .mappingFileProvider is missing for com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@2c55d775 - Error: Cannot query the value of this property because it has no value available.
[sentry] .mappingFileProvider is missing for com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@25e00289 - Error: Cannot query the value of this property because it has no value available.
[sentry] .mappingFileProvider is missing for com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@14cd05c6 - Error: Cannot query the value of this property because it has no value available.
```

Probable reason for this is caused by usage of the Eager API `project.tasks.create` vs the configuration avoidance API `project.tasks.register`.

Screenshot below shows output of `./gradlew help --scan` in Performance > Configuration detail page showing tasks created during configuration.

![image](https://user-images.githubusercontent.com/1126594/116582779-fbde7680-a915-11eb-8e3e-8bfd98b31fc9.png)

Issue #82, opened 2021-04-29
Plugin fails for variants that generate no mappings


cerisier:
_Android Gradle Plugin:_
**4.1.3**

_Gradle:_
**6.7.1**

_Optimizer:_
- [ ]  ProGuard
- [x]  R8

_Have you created the sentry.properties file?_
- [x]  Yes -> where?
- [ ]  No

_SDKs:_
- [x]  sentry-android-gradle-plugin -> 2.0.0-alpha.2

---
I have the following issue:

Plugin fails with missing mappings for variants that do not produce mappings (`minifyEnabled false`)
```
problem was found with the configuration of task ':app:uploadSentryProguardMappingsRegularLoveDebugRegular-love-debug' (type 'SentryUploadProguardMappingsTask').
--
  | > No value has been specified for property 'mappingsFile'.
```

cerisier:
Just seen you created a similar issue #83. Letting you close if needed.

marandaneto:
I'll let yours as you opened first :)

what we can do is:

not creating `uploadSentryProguardMappingsTask` if `variant.buildType.isMinifyEnabled` is disabled for example

PR #85, opened 2021-04-29
Fix: Avoid unnecessary config

marandaneto:
Should fix https://github.com/getsentry/sentry-android-gradle-plugin/issues/84 and https://github.com/getsentry/sentry-android-gradle-plugin/issues/82

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -25,9 +25,9 @@ class SentryPlugin : Plugin<Project> {
     @OptIn(ExperimentalStdlibApi::class)
     override fun apply(project: Project) {
         val extension = project.extensions.create(
-                "sentry",
-                SentryPluginExtension::class.java,
-                project
+            "sentry",
+            SentryPluginExtension::class.java,
+            project
         )
         project.afterEvaluate {
             check(project.plugins.hasPlugin(AppPlugin::class.java)) {
@@ -37,14 +37,14 @@ class SentryPlugin : Plugin<Project> {
             val androidExtension = project.extensions.getByType(AppExtension::class.java)
 
             fun withLogging(varName: String, initializer: () -> Task?) =
-                    initializer().also {
-                        project.logger.info("[sentry] $varName is ${it?.path}")
-                    }
+                initializer().also {
+                    project.logger.info("[sentry] $varName is ${it?.path}")
+                }
 
             val cliExecutable = getSentryCliPath(project)
 
             val extraProperties = project.extensions.getByName("ext")
-                    as ExtraPropertiesExtension
+                as ExtraPropertiesExtension
 
             val sentryOrgParameter = runCatching {
                 extraProperties.get(SENTRY_ORG_PARAMETER).toString()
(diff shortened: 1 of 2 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -36,84 +36,116 @@ class SentryPlugin : Plugin<Project> {
...
+                            )
+                            it.mappingsUuid.set(generateUuidTask.get().outputUuid)
+                            mappingFile?.let { mapFile ->
+                                it.mappingsFile.set(mapFile)
+                            }
+                            it.autoUpload.set(extension.autoUpload.get())
+                            it.sentryOrganization.set(sentryOrgParameter)
+                            it.sentryProject.set(sentryProjectParameter)
+                        }
+                        variant.register(uploadSentryProguardMappingsTask.get())
```
cerisier, line 130: Doing this will eagerly create the `uploadSentryProguardMappingsTask` task.
I am wondering if we need `variant.register` here since we already configure tasks dependency below.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -36,84 +36,116 @@ class SentryPlugin : Plugin<Project> {
...
+                        androidExtension.sourceSets.getByName(variant.name).assets.srcDir(
+                            generateUuidTask.get().outputDirectory
```
cerisier, line 132: Doing this will eagerly create the `generateUuidTask` task.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -36,84 +36,117 @@ class SentryPlugin : Plugin<Project> {
...
             androidExtension.applicationVariants.all { variant ->
-                variant.outputs.all { variantOutput ->
-                    val taskSuffix = variant.name.capitalize(Locale.ROOT) +
-                        variantOutput.name.capitalize(Locale.ROOT)
 
-                    fun withLogging(varName: String, initializer: () -> Task?) =
-                        initializer().also {
-                            project.logger.info("[sentry] $varName is ${it?.path}")
-                        }
+                val varNameCapitalize = variant.name.capitalize(Locale.ROOT)
```
cortinico, line 58: Do you really need this extra variable?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -36,84 +36,117 @@ class SentryPlugin : Plugin<Project> {
...
-                    val sentryProjectParameter = runCatching {
-                        extraProperties.get(SENTRY_PROJECT_PARAMETER).toString()
-                    }.getOrNull()
-
-                    val mappingFile = getMappingFile(project, variant)
-                    val cliExecutable = getSentryCliPath(project)
-                    val sep = File.separator
+                    mappingFile = getMappingFile(project, variant)
+                } else {
+                    project.logger.info("[sentry] isMinifyEnabled is disabled.")
```
cortinico, line 90:
```suggestion
                    project.logger.info("[sentry] isMinifyEnabled is false for variant ${variant.name}.")
```

cortinico (commented):
The change looks good to me. Perhaps adding a couple of tests would be 👌 

There are a couple of improvements mentioned by @cerisier that I believe we should include:

> Doing this will eagerly create the generateUuidTask task.

Agree. @marandaneto can you please export this:
https://github.com/getsentry/sentry-android-gradle-plugin/blob/149bbfefbc5f37774f2373d2bc30802603668562/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt#L86-L88
To a variable and re-use it when populating the srcset?

> I am wondering if we need variant.register here since we already configure tasks dependency below.

I had the same feeling when rewriting the plugin. I'm unsure why there is both a `.register()` and explicit dependency declaration here.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cerisier: This one will go away as it was removed in prior PR 👌

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cerisier: This one is going to be a bit more problematic since this is not accessible from a `TaskProvider`. Any idea @cortinico ?

marandaneto: indeed, requires calling `generateUuidTask.get()`, will need to figure this out and I'm not an AGP/Gralde specialist :( will dig into it tomorrow

PR #95, opened 2021-04-30
Add AGP 7.x to the CI Matrix

### LINKED ISSUE #89: Add AGP 7.x support
Let's add it to the CI matrix and see what happens

https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/workflows/test-publish-dry-run.yaml

requires setting up a conditional Java 8 or 11 though, AGP 7 only runs with Java 11

cortinico:
## :scroll: Description
Adding an entry for AGP `7.x` to both the CI and the test matrix.
I've also cleaned up the matrix a bit, removing `experimental` (I was planning to use it for AGP 7.x but turns out is green already) and the `name` parameter. I've added the `java` parameter to properly configure the Java version.

Fixes #89 

Tests are green locally (as long as you run them on Java 11+).

## :pencil: Checklist
- [x] No breaking changes

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -29,6 +29,12 @@ jobs:
           restore-keys: |
             cache-gradle-
 
+      - name: Setup Java Version
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '11'
+
       - name: Run Gradle tasks
         run: ./gradlew preMerge --continue
 
--- a/.github/workflows/test-publish-dry-run.yaml
+++ b/.github/workflows/test-publish-dry-run.yaml
@@ -16,32 +16,28 @@ jobs:
       matrix:
         agp: ["4.1.0"]
         gradle: ["6.8.3"]
-        experimental: [false]
-        name: ["stable"]
+        java: ["8"]
         include:
           - agp: "4.0.0"
             gradle: "6.1.1"
-            experimental: false
-            name: AGP-4.0.0
+            java: "8"
           - agp: "4.1.3"
             gradle: "6.5"
-            experimental: false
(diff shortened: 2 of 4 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -29,6 +29,12 @@ jobs:
           restore-keys: |
             cache-gradle-
 
+      - name: Setup Java Version
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '11'
```
marandaneto, line 36: shouldn't be behind a `${{ matrix.java }}` too? when it runs `./gradlew assembleRelease` using Java 11 it might fail

cortinico: > using Java 11 it might fail

Why this?

marandaneto: just guessing here, Android always required Java 8 or am I missing something?

marandaneto: and if it's not the case, why do we need a Java matrix in the `test-publish-dry-run.yaml` file then? Java 11 would work for every use case

cortinico: `test-publish-dry-run.yaml` is more testing from the users prospective. We want to try on an environment that is similar to the potential user machine, therefore having Java 8 on pre-AGP7 is reasonable.

The pre-merge workflow is instead building the project + running the Gradle runner tests. We necessarily need Java 11 here as on of those test is doing a run on AGP7 and will fail if run on Java8.

marandaneto: makes sense, and CI is green so all good

marandaneto (approved):
LGTM

PR #110, opened 2021-05-21
Add support for hashing the CLI

### LINKED ISSUE #103: Check sentry-cli version or hash
`download-sentry-cli.sh` deletes the file and download them again

https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/buildSrc/build.gradle.kts#L8-L16 task only executes if the files don't exist, ideally, check if the files are up to date or not based on hash or version and download them again if necessary.

right now when we bump a version in the `download-sentry-cli` file, we need to call `download-sentry-cli.sh` manually if the files already exist

cortinico:
## :scroll: Description
Fixes #103 

I've extended the condition to decide when to download the sentry CLI with a check upon the local MD5 hash. I believe that's simple enough to protect against stale version of the CLI.

Tested locally, works fine so far.

## :pencil: Checklist
- [x] No breaking changes

## :crystal_ball: Next steps

I've added a comment on the `shouldDownloadSentryCli` function to remind to update the `expected-checksums.md5` file once the CLI gets bumped. Not a major deal but the CLI will be re-downloaded every time otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 * Bump: sentry-cli to 1.64.1
-* Add support for hashing sentry-cli to trigger a re-download upon bumps
+* Add support for hashing sentry-cli to trigger a re-download upon bumps (#110)
 
 # 2.0.0-alpha.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Bump: sentry-cli to 1.64.1
+* Add support for hashing sentry-cli to trigger a re-download upon bumps
 
 # 2.0.0-alpha.3
 
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -15,15 +15,15 @@ if(shouldDownloadSentryCli()) {
 }
 
 /**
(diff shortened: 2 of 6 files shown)


cortinico:
That's a bit unfortunate. `macos` runners don't have `md5sum` but only `md5`. `windows` and `linux` runners have only `md5sum` but not `md5` :|

cortinico:
I ended up using `shasum` as it's the only command available on all the runners (and aliases to `sha1sum`). Should the SHA algorithm change, that's not going to be a problem as this is a local solution and works as long as the algo is consistent on the same machine/worker.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Bump: sentry-cli to 1.64.1
+* Add support for hashing sentry-cli to trigger a re-download upon bumps
```
marandaneto, line 4:
```suggestion
* Add support for hashing sentry-cli to trigger a re-download upon bumps (#110)
```

marandaneto (approved):

marandaneto:
> I ended up using `shasum` as it's the only command available on all the runners (and aliases to `sha1sum`). Should the SHA algorithm change, that's not going to be a problem as this is a local solution and works as long as the algo is consistent on the same machine/worker.

yeah sounds good to me.

PR #112, opened 2021-05-23
Support Configuration Avoidance

### LINKED ISSUE #84: Plugin eagerly creates tasks for all variants

cortinico:
## :scroll: Description
I've cherry-picked the commits on #85 + I've applied 2943df5 to make sure we can leverage task configuration avoidance.

I had to remove the `Property<UUID>` as that was forcing the `generateUuidTask` task to be created at configuration time. Now the connection point between the two task is the output asset folder.

Closes #85
Fixes #84 

JUnit tests are attached.

## :pencil: Checklist
- [x] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -36,83 +36,117 @@ class SentryPlugin : Plugin<Project> {
 
             val androidExtension = project.extensions.getByType(AppExtension::class.java)
 
+            fun withLogging(varName: String, initializer: () -> Task?) =
+                initializer().also {
+                    project.logger.info("[sentry] $varName is ${it?.path}")
+                }
+
+            val cliExecutable = getSentryCliPath(project)
+
+            val extraProperties = project.extensions.getByName("ext")
+                as ExtraPropertiesExtension
+
+            val sentryOrgParameter = runCatching {
+                extraProperties.get(SENTRY_ORG_PARAMETER).toString()
+            }.getOrNull()
+            val sentryProjectParameter = runCatching {
+                extraProperties.get(SENTRY_PROJECT_PARAMETER).toString()
+            }.getOrNull()
+
             androidExtension.applicationVariants.all { variant ->
-                variant.outputs.all { variantOutput ->
-                    val taskSuffix = variant.name.capitalize(Locale.ROOT) +
-                        variantOutput.name.capitalize(Locale.ROOT)
 
-                    fun withLogging(varName: String, initializer: () -> Task?) =
-                        initializer().also {
-                            project.logger.info("[sentry] $varName is ${it?.path}")
-                        }
+                val varNameCapitalize = variant.name.capitalize(Locale.ROOT)
 
-                    val dexTask = withLogging("dexTask") { getDexTask(project, variant.name) }
-                    val preBundleTask = withLogging("preBundleTask") {
+                val bundleTask = withLogging("bundleTask") {
+                    getBundleTask(project, variant.name)
+                }
+
+                val assembleTask = withLogging("assembleTask") { getAssembleTask(variant) }
+
+                val sentryProperties = getPropertiesFilePath(project, variant)
+
+                val isMinifyEnabled = variant.buildType.isMinifyEnabled
+
+                var dexTask: Task? = null
+                var preBundleTask: Task? = null
+                var transformerTask: Task? = null
+                var packageTask: Task? = null
+                var mappingFile: File? = null
+                val sep = File.separator
+
+                if (isMinifyEnabled) {
+                    dexTask = withLogging("dexTask") { getDexTask(project, variant.name) }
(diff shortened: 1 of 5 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -36,83 +36,119 @@ class SentryPlugin : Plugin<Project> {
...
+                        "[sentry] isMinifyEnabled is false for variant ${variant.name}."
+                    )
+                }
 
-                    val mappingFile = getMappingFile(project, variant)
-                    val cliExecutable = getSentryCliPath(project)
-                    val sep = File.separator
+                variant.outputs.all { variantOutput ->
+                    val taskSuffix = variant.name.capitalize(Locale.ROOT) +
+                        variantOutput.name.capitalize(Locale.ROOT)
```
marandaneto, line 99: could use string templates here "${var}"

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadProguardMappingsTask.kt
```diff
@@ -87,4 +96,19 @@ abstract class SentryUploadProguardMappingsTask : Exec() {
...
+
+        fun readUuidFromFile(file: File): String {
+            check(file.exists()) {
+                "UUID properties file is missing"
+            }
+            val content = file.readText().trim()
+            check(content.startsWith(PROPERTY_PREFIX)) {
+                "io.sentry.ProguardUuids property is missing"
+            }
+            return content.removePrefix(PROPERTY_PREFIX)
```
marandaneto, line 111: another way of doing this would be via `java.util.Properties.load(InputStream inStream)`
not sure if Kotlin has a nicer extension

marandaneto: do you think its worth changing it?

cortinico: Yeah I thought about using that as well, but I think that we probably don't need to use `java.util.Properties` as this is a file with a single property and we control it end-to-end (user is not supposed to edit it).

We could switch to `java.util.Properties` if we happen to use this file to pass more values.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cortinico: Yup fixed

cortinico: Actually not, using string templates will lead the formatter to produce this:

```
val taskSuffix =
    "${variant.name.capitalize(Locale.ROOT)}${variantOutput.name.capitalize(
        Locale.ROOT
    )}"
```

That is not great. I'd rather stick to the `+` here

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadProguardMappingsTask.kt
marandaneto: wfm, thanks

PR #119, opened 2021-05-27
Enhancement: Use plugin manager and other improvements

### LINKED ISSUE #104: Use Kotlin 1.3 instead of 1.4
Right now we're using Kotlin 1.4, we need to be sure that we're not writing incompatible code with 1.3 or to pin Kotlin 1.3 compatibility using `languageVersion`, the reason is https://github.com/gradle/gradle/issues/16345

cc @cortinico

marandaneto:
## :scroll: Description
Does not require `afterEvaluate` block by using `pluginManager.withPlugin` combined with `applicationVariants.configureEach`  instead of `applicationVariants.all`
inspired by https://github.com/remcomokveld/sentry-android-gradle-plugin/commit/67fcdd8ef3e7bec64537f631876ab3bfd614e489

Fix #104

tests are green and also executed locally

## :pencil: Checklist

- [X] No breaking changes


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Enhancement: Use pluginManager instead of project.afterEvaluate
```
marandaneto, line 3:
```suggestion
* Enhancement: Use pluginManager instead of project.afterEvaluate (#119)
```

plugin-build/build.gradle.kts
```diff
@@ -29,6 +29,7 @@ tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions {
         jvmTarget = "1.8"
         freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
+        languageVersion = "1.3"
```
cortinico, line 32: I'm unsure we need this.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -94,84 +88,87 @@ class SentryPlugin : Plugin<Project> {
...
-
-                        // To include proguard uuid file into aab, run before bundle task.
-                        preBundleTask?.dependsOn(uploadSentryProguardMappingsTask)
+                var taskSuffix = variant.name
+                // capitalizeUS
+                taskSuffix = if (taskSuffix.isEmpty()) {
+                    ""
+                } else {
+                    taskSuffix.substring(0, 1).toUpperCase(Locale.US) + taskSuffix.substring(1)
+                }
```
cortinico, line 97: Why are you doing this? Plus this should potentially be exported to a function and unit tested.

.github/workflows/test-publish.yaml
```diff
@@ -31,4 +31,4 @@ jobs:
             cache-gradle-
 
       - name: Run the uploadSentryProguardMappings task
-        run: ./gradlew assembleRelease uploadSentryProguardMappingsReleaseRelease
+        run: ./gradlew assembleRelease uploadSentryProguardMappingsRelease
```
cortinico, line 34: Commenting here but this is related to the whole PR:

This is basically a breaking change for the API exposed to customers.
This means that any build logic (`build.gradle` or custom plugins) that rely on this plugin, will break.

I don't think this is a major concern, but it worth to point it out clearly in the release notes.

cortinico:
Great stuff with the `configureEach` 💪

.github/workflows/test-publish.yaml
marandaneto: the task is called automatically, we never expect people calling this directly, so I don't think its a breaking change

plugin-build/build.gradle.kts
marandaneto: its not required, its just better compatibility and fix https://github.com/getsentry/sentry-android-gradle-plugin/issues/104
we're only using the `capitalize` from Kotlin 1.4, not a biggie

.github/workflows/test-publish.yaml
cortinico: > we never expect people calling this directly

😅 

I'm not entirely sure here, but if someone has a `./gradlew uploadSentryProguardMappingsReleaseRelease` in their CI build because they want to explicitly be sure that the upload happens -> Their CI step will break.

marandaneto: it'd be an undocumented API, it's fine in this case, our docs say, set up the plugin and that's it, tasks are called automatically.
obviously, we can mention in the changelog

plugin-build/build.gradle.kts
cortinico: > its just better compatibility and fix #104

It doesn't actually Fix #104. To fix that you would have to downgrade the Kotlin version here https://github.com/getsentry/sentry-android-gradle-plugin/blob/6d92620153554a4c581b38b56ef4c805cd921ef0/buildSrc/src/main/java/Dependencies.kt#L4

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cortinico: Gotcha 👍 
You can also reword it with:

```suggestion
                var taskSuffix = variant.name
                // capitalizeUS
                taskSuffix = if (taskSuffix.isEmpty()) {
                    ""
                } else {
                    taskSuffix.first().toUpperCase(Locale.US) + taskSuffix.drop(1)
                }
```

plugin-build/build.gradle.kts
marandaneto: well it does if you've written Kotlin 1.3 compatible code, it could be run using Kotlin >= 1.4
To be sure you've written 1.3 compatible code, `languageVersion` is the way to go afaik

cortinico: Got it 👍 Let's go for it. We're not making extended usage of Kotlin 1.4 APIs so that's not a major deal.

cortinico (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
marandaneto: extracted and added test to it

marandaneto: `Char` does not have a .toUpperCase(locale) so cant use it if I want to `toUpperCase(Locale.US)`

marandaneto: `drop` calls `substring` so its not needed, just one more method call

PR #126, opened 2021-05-28
Feat: use Sentry SDK

marandaneto:
## :scroll: Description
Feat: use Sentry SDK in the Plugin
We should offer a way to turn it off though, maybe a boolean flag in the plugin extensions

The idea is to have more context when we get issues using the Sentry gradle plugin which heavily depends on some data like AGP, Gradle version, environment etc.

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
looks like that the stack traces are always Gradle stuff, need to figure out how to capture our own stack traces, maybe Gradle is always wrapping up but not adding as a cause

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -1,6 +1,7 @@
 package io.sentry.android.gradle
 
 import com.android.build.gradle.AppExtension
+import io.sentry.Sentry
 import io.sentry.android.gradle.SentryCliProvider.getSentryCliPath
 import io.sentry.android.gradle.SentryMappingFileProvider.getMappingFile
 import io.sentry.android.gradle.SentryPropertiesFileProvider.getPropertiesFilePath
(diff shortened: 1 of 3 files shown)


marandaneto:
Screenshot 2021-05-28 at 14 55 46

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -19,9 +20,17 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.plugins.ExtraPropertiesExtension
+import org.gradle.util.GradleVersion
 
 class SentryPlugin : Plugin<Project> {
     override fun apply(project: Project) {
+        Sentry.init {
+            it.dsn = "https://1053864c67cc410aa1ffc9701bd6f93d@o447951.ingest.sentry.io/5428559"
+            // read release dynamically somehow
```
cortinico, line 29: For the future: this could be done with
https://github.com/gmazzo/gradle-buildconfig-plugin

marandaneto: indeed, good idea, we use it on https://github.com/getsentry/sentry-java as well

philipphofmann:
@marandaneto said we can close this.

PR #122, opened 2021-05-28
Enha: Lazy mapping file

marandaneto:
## :scroll: Description
Enha: Lazy mapping file

## :bulb: Motivation and Context
Fix #118

## :pencil: Checklist

- [X] I reviewed the submitted code
- [X] I added tests to verify the changes
- [ ] I updated the docs if needed
- [X] No breaking changes

PR #128, opened 2021-05-30
Use .mappingFileProvider

### LINKED ISSUE #118: use `variant.mappingFileProvider` lazily instead of `variant.mappingFileProvider.get()`
inspired by https://github.com/remcomokveld/sentry-android-gradle-plugin/commit/67fcdd8ef3e7bec64537f631876ab3bfd614e489

cortinico:
## :scroll: Description
This removes the `SentryMappingFileProvider` and switches to use directly `variant.mappingFileProvider`.

Fixes #118
Closes #122

Tests are attached

## :pencil: Checklist

- [ ] No breaking changes
👆 That's technically a breaking change, but I don't think users were instantiating the tasks directly so that should not be a major deal.


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadProguardMappingsTask.kt
```diff
@@ -49,6 +51,9 @@ abstract class SentryUploadProguardMappingsTask : Exec() {
     abstract val autoUpload: Property<Boolean>
 
     override fun exec() {
+        if (!mappingsFiles.isPresent || mappingsFiles.get().isEmpty) {
+            error("[sentry] Mapping files is missing!")
```
marandaneto, line 55:
```suggestion
            error("[sentry] Mapping files are missing!")
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -107,9 +107,7 @@ class SentryPlugin : Plugin<Project> {
                             sentryProperties?.let { file -> project.file(file) }
                         )
                         it.uuidDirectory.set(uuidOutputDirectory)
-                        mappingFile?.let { mapFile ->
-                            it.mappingsFile.set(mapFile)
-                        }
+                        it.mappingsFiles = mappingFileProvider
```
marandaneto, line 110: would be possible to just assign `variant.mappingFileProvider` directly?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadProguardMappingsTask.kt
```diff
@@ -30,8 +32,8 @@ abstract class SentryUploadProguardMappingsTask : Exec() {
     val uuidFile: Provider<RegularFile>
         get() = uuidDirectory.file("sentry-debug-meta.properties")
 
-    @get:InputFile
-    abstract val mappingsFile: RegularFileProperty
+    @get:InputFiles
```
marandaneto, line 35: I think that did the trick to solve what I have tried before and it failed :D, Files instead of File

marandaneto (approved):
nice, LGTM

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cortinico: Indeed 👍

PR #129, opened 2021-05-30
Add a workflow step for App Bundles

### LINKED ISSUE #123: Add tests for App Bundle
Right now tests are for APK only but not for `aab` which is the future of Android Apps.

so ideally, we also have a CI step for `./gradlew bundleRelease` and test the code that is used eg `getPreBundleTask` and `getPackageTask`
CI should also unzip the `aab` in test-publish-dry-run.yaml and see if the file exist

cortinico:
## :scroll: Description
Fixes #123 

I've tested the change locally and it's green. Waiting for a first CI result.

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/.github/workflows/test-publish-dry-run.yaml
+++ b/.github/workflows/test-publish-dry-run.yaml
@@ -68,7 +68,7 @@ jobs:
       - name: Set Gradle Version
         run: ./gradlew wrapper --gradle-version $VERSION_GRADLE --distribution-type=all
 
-      - name: Run the uploadSentryProguardMappings task
+      - name: Build the release APK and run uploadSentryProguardMappings task
         run: ./gradlew assembleRelease uploadSentryProguardMappingsRelease
 
       - name: Check sentry-debug-meta.properties inside APKs
@@ -78,4 +78,16 @@ jobs:
           rm -r output
           unzip ./examples/android-gradle-kts/build/outputs/apk/release/android-gradle-kts-release-unsigned.apk -d output
           cat output/assets/sentry-debug-meta.properties
+          rm -r output
+
+      - name: Build the release App Bundle and run uploadSentryProguardMappings task
+        run: ./gradlew bundleRelease uploadSentryProguardMappingsRelease
+
+      - name: Check sentry-debug-meta.properties inside App Bundle
+        run: |
+          unzip ./examples/android-gradle/build/outputs/bundle/release/android-gradle-release.aab -d output
+          cat output/base/assets/sentry-debug-meta.properties
+          rm -r output
+          unzip ./examples/android-gradle-kts/build/outputs/bundle/release/android-gradle-kts-release.aab -d output
+          cat output/base/assets/sentry-debug-meta.properties
           rm -r output


.github/workflows/test-publish-dry-run.yaml
```diff
@@ -79,3 +79,15 @@ jobs:
           unzip ./examples/android-gradle-kts/build/outputs/apk/release/android-gradle-kts-release-unsigned.apk -d output
           cat output/assets/sentry-debug-meta.properties
           rm -r output
+
+      - name: Build the release App Bundle and run uploadSentryProguardMappings task
+        run: ./gradlew bundleRelease uploadSentryProguardMappingsRelease
```
marandaneto, line 84: a note not related to this PR (because the APK test also do it):

our test calls the assemble task and `uploadSentryProguardMappingsRelease`, but the thing is, in the real world people won't call `uploadSentryProguardMappingsRelease` directly so our tests should not rely on it.

if CI passes because `uploadSentryProguardMappingsRelease` was called manually but it does not work if you don't call `uploadSentryProguardMappingsRelease` manually, we have a false positive green CI.

what's about removing these task calls? tests should be green by just calling `assembleRelease` or `bundleRelease`

marandaneto (approved):
niice, thanks @cortinico
left a comment but its not a blocker, LGTM

.github/workflows/test-publish-dry-run.yaml
cortinico: > if CI passes because `uploadSentryProguardMappingsRelease` was called manually but it does not work if you don't call `uploadSentryProguardMappingsRelease` manually, we have a false positive green CI.

Agree.

Ideally I would like to:

```
./gradlew bundleRelease
```

And then verify that the Gradle call invoked `uploadSentryProguardMappingsRelease`.
I haven't found an elegant way to verify that, other than:

```
./gradlew bundleRelease --info | tee output.txt && grep uploadSentryProguardMappingsRelease output.txt
```

That is far from optimal 🤔 

The risk of calling just `bundleRelease` is that you don't verify at all that `uploadSentryProguardMappingsRelease` is invoked. The Gradle execution can succeed and the checks inside the APK will also be green (due to the generate UUID task).

marandaneto: mm I agree with the statement, but if the mapping file is inside of the APK or AAB, that means the `uploadSentryProguardMappingsRelease` was invoked and did its job, so I guess I'd be fine with it.

`uploadSentryProguardMappingsRelease` does the task scheduling, plus calling the other tasks and glue them  together, so I'd assume that only calling the `bundle/assebleRelease` plus unzipping the apk/aab and checking the file inside of the assets folder is good enough, wdyt?

cortinico: > that means the uploadSentryProguardMappingsRelease was invoked and did its job

Actually it means that `generateSentryProguardUuidRelease` was run, but not necessarily `uploadSentryProguardMappingsRelease`

marandaneto: yeah indeed, `SentryUploadProguardMappingsTask` is only responsible to call sentry-cli with the arguments, let's keep as it is for now then, good point

PR #136, opened 2021-06-02
ref(craft): Modernize Craft config

BYK:
#skip-changelog

#skip-changelog


bruno-garcia:
Odd that Danger failed due to permission issues. Usually happens with outside contributors

bruno-garcia (approved):
tks

PR #144, opened 2021-07-07
ci(release): Run CI on release branches

iker-barriocanal:
In order to get the artifacts that may be generated when doing a release, release branches must run CI.

_#skip-changelog_

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - release/**
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - release/**
   pull_request:
     branches:
       - '*'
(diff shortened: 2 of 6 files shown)


.github/workflows/danger.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - release/**
```
cortinico, line 6: Don't you need to update all the other workflows as well?

iker-barriocanal: No, Craft will create a `release/{version}` branch and take the artifacts from the CI of that branch. See the first step in https://github.com/getsentry/craft#integrating-your-project-with-craft.

cortinico: The first step mention that:

> Set up a workflow that builds your assets and runs your tests. Allow building release branches (their names follow release/{VERSION} by default, configurable through releaseBranchPrefix).

Here you're editing the `danger.yml` workflow that is just checking that the release notes are correct. I believe you should include the `release/**` branch rule on also the other workflows that are actually running tests (e.g. `pre-merge.yml` and others). Or am I missing something? :)

iker-barriocanal: Yes, you're right. Good catch, updated.

cortinico (approved):

PR #145, opened 2021-07-14
build(dist): Generate distribution artifacts

iker-barriocanal:
[Craft](https://github.com/getsentry/craft) will manage the release of this repo. In order to achieve that, the distribution artifacts must be available so that Craft can use them to make a release. At the moment, those artifacts aren't being generated. This PR aims to update the build so that these artifacts are generated.

_#skip-changelog_


plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,34 @@ if (gradle.gradleVersion >= "6.6.0") {
...
+    configure<DistributionContainer> {
+        this.getByName("main").contents {
+            from("build${sep}libs")
+            from("build${sep}publications${sep}maven")
+            from("build${sep}publications${sep}sentryPluginPluginMarkerMaven")
+        }
+    }
+
+    tasks.named("distZip").configure {
+        this.dependsOn("publishToMavenLocal")
```
marandaneto, line 88: need to validate if this is the right name, `./gradlew tasks` would show the right names

marandaneto: just did, its right

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,34 @@ if (gradle.gradleVersion >= "6.6.0") {
     val publish = extensions.getByType(MavenPublishPluginExtension::class.java)
     publish.releaseSigningEnabled = BuildUtils.shouldSignArtifacts()
 }
+
+subprojects {
```
marandaneto, line 74: as this is a module and not a high level build script, you need to remove the `subprojects` and keep all the rest

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,34 @@ if (gradle.gradleVersion >= "6.6.0") {
...
+
+    configure<DistributionContainer> {
+        this.getByName("main").contents {
+            from("build${sep}libs")
+            from("build${sep}publications${sep}maven")
+            from("build${sep}publications${sep}sentryPluginPluginMarkerMaven")
+        }
+    }
+
+    tasks.named("distZip").configure {
```
marandaneto, line 87: this is failing right now due to:

> * What went wrong:
> Execution failed for task ':distZip'.
> > Entry sentry-android-gradle-plugin-2.1.1-SNAPSHOT/pom-default.xml is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.0.2/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,34 @@ if (gradle.gradleVersion >= "6.6.0") {
     val publish = extensions.getByType(MavenPublishPluginExtension::class.java)
     publish.releaseSigningEnabled = BuildUtils.shouldSignArtifacts()
 }
+
```
marandaneto, line 73: right now the gradle plugin is gpg signing the files, if we don't want it, we need to
https://github.com/getsentry/sentry-java/blob/main/build.gradle.kts#L95-L99

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,34 @@ if (gradle.gradleVersion >= "6.6.0") {
...
+subprojects {
+    apply<DistributionPlugin>()
+
+    val sep = File.separator
+
+    configure<DistributionContainer> {
+        this.getByName("main").contents {
+            from("build${sep}libs")
+            from("build${sep}publications${sep}maven")
+            from("build${sep}publications${sep}sentryPluginPluginMarkerMaven")
```
marandaneto, line 83:
```suggestion
            from("build${sep}publications${sep}sentryPluginPluginMarkerMaven")
            duplicatesStrategy = DuplicatesStrategy.INCLUDE
```

marandaneto: that would fix https://github.com/getsentry/sentry-android-gradle-plugin/pull/145/files#r669520898 I guess

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,40 @@ if (gradle.gradleVersion >= "6.6.0") {
...
+        if (!file.exists()) {
+            throw IllegalStateException(
+                "Distribution file: $distributionFilePath does not exist"
+            )
+        }
+        if (file.length() == 0L) {
+            throw IllegalStateException("Distribution file: $distributionFilePath is empty")
+        }
+    }
+}
```
cortinico, line 101: Can you provide some context on why this is needed (like who is supposed to consume this)?

iker-barriocanal: Updated the PR description.

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,41 @@ if (gradle.gradleVersion >= "6.6.0") {
     val publish = extensions.getByType(MavenPublishPluginExtension::class.java)
     publish.releaseSigningEnabled = BuildUtils.shouldSignArtifacts()
 }
+
+apply<DistributionPlugin>()
```
cortinico, line 74: Please add: `id("distribution")` on top and remove this one 👍

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,41 @@ if (gradle.gradleVersion >= "6.6.0") {
...
+val sep = File.separator
+
+configure<DistributionContainer> {
+    this.getByName("main").contents {
+        from("build${sep}libs")
+        from("build${sep}publications${sep}maven")
+        from("build${sep}publications${sep}sentryPluginPluginMarkerMaven")
+        duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    }
+}
```
cortinico, line 85:
```suggestion
distributions {
    main {
      duplicatesStrategy = DuplicatesStrategy.INCLUDE
      contents {
        from("build${sep}libs")
        from("build${sep}publications${sep}maven")
        from("build${sep}publications${sep}sentryPluginPluginMarkerMaven")
      }
    }
}
```
This should also work.

plugin-build/build.gradle.kts
```diff
@@ -70,3 +70,41 @@ if (gradle.gradleVersion >= "6.6.0") {
...
+
+afterEvaluate {
+    configure<MavenPublishPluginExtension> {
+        // signing is done when uploading files to MC
+        // via gpg:sign-and-deploy-file (release.kts)
+        releaseSigningEnabled = false
+    }
+}
```
cortinico, line 110: I believe this is unnecessary. Specifically the task should not depend on `publishToMavenLocal` (alternatively on `assemble`).
You should also not check the size of the output. If something goes wrong Gradle will fail :)

iker-barriocanal: Depending on `publishToMavenLocal` is required. If not, the first build won't generate `distributions/` and you'll need to run it again.

plugin-build/build.gradle.kts
iker-barriocanal: It seems it's not compatible with Gradle 8.0 ([see CI](https://github.com/getsentry/sentry-android-gradle-plugin/runs/3067385850#step:5:302)), so moving back to the previous state.

cortinico: The CI is failing as it's not properly formatted (as I typed that manually 😅). You should be able to solve this by running: `./gradlew -p plugin-build ktlintKotlinScriptFormat` that will reformat the build file.

plugin-build/build.gradle.kts
marandaneto: the build should fail if it's not generated as our release process depends on it.

plugin-build/build.gradle.kts
marandaneto: @iker-barriocanal why is it resolved if its not applied?

plugin-build/build.gradle.kts
marandaneto: @iker-barriocanal the suggestion makes sense, its kotlin friendlier, lets do it and apply the formatting

plugin-build/build.gradle.kts
cortinico: I understand. Not elegant but yeah ¯\_(ツ)_/¯

Also a couple of things that are still worth fixing:
1. This whole code should go inside the if-then-else where the maven.publish plugin is applied (check for Gradle 6.x). This because `distribution` now depends on `maven.publish` and it should be applied only on Gradle 6.x
2. I still believe the whole `this.doLast` block can be removed. If you want to be sure that the task is actually executing you can replace it with a:
```kotlin
if(inputs.sourceFiles.isEmpty) throw StopExecutionException("...")
``` 

That will prevent the `distZip` fail to return with a `NO-SOURCE` result if the `publishToMavenLocal` produces no output.

iker-barriocanal: The task is being run, the issue is the timing. If the `doLast` block is removed in the latest changes (`fef85c5`), it checks whether the file exists before creating it. I don't know how the task execution is scheduled, nor how they are optimized (if they are); but that task seems to be required to execute last (to not run before the `zip` is created). Let me know if you dive more into that.

cortinico: Yup I see what's the problem here. I've pushed a branch with the fix: https://github.com/cortinico/sentry-android-gradle-plugin/tree/nc/build-artifacts

Specifically on https://github.com/cortinico/sentry-android-gradle-plugin/commit/8d11c3b4aca92f33b9fea7ead9b2dbce1069bd9d

You can use `onlyIf` that gets evaluated at execution time to do the job.
I've verified it locally with:

```
$ rm -rf plugin-build/build && ./gradlew -p plugin-build distZip && ls plugin-build/build/distributions

To honour the JVM settings for this build a single-use Daemon process will be forked. See https://docs.gradle.org/7.0.2/userguide/gradle_daemon.html#sec:disabling_the_daemon.
Daemon will be stopped at the end of the build

> Task :dokkaHtml
Initializing plugins
Dokka is performing: documentation for sentry-android-gradle-plugin
Validity check
Creating documentation models
Transforming documentation model before merging
Merging documentation models
Transforming documentation model after merging
Creating pages
Transforming pages
Rendering

> Task :compileKotlin
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    [...]
w: Consider providing an explicit dependency on kotlin-reflect 1.4 to prevent strange errors
w: Some runtime JAR files in the classpath have an incompatible version. Consider removing them from the classpath

> Task :publishPluginMavenPublicationToMavenLocal
Multiple publications with coordinates 'io.sentry:sentry-android-gradle-plugin:2.1.1-SNAPSHOT' are published to repository 'mavenLocal'. The publications will overwrite each other!

BUILD SUCCESSFUL in 28s
17 actionable tasks: 17 executed
sentry-android-gradle-plugin-2.1.1-SNAPSHOT.zip
```

iker-barriocanal: Tried to play around with it for a while to see if I could improve it a bit, but didn't manage, so using your approach 👍.

cortinico (approved):

marandaneto (approved):

PR #147, opened 2021-07-16
ci(dist): Generate distribution artifacts in GHA

iker-barriocanal:
Generates distribution artifacts on GHA.

_#skip-changelog_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,11 +34,10 @@ jobs:
           path: ${{ env.CACHED_BUILD_PATHS }}
           key: ${{ github.sha }}
 
-      - name: Build the plugin
+      - name: Build distribution artifacts
         run: |
           cd plugin-build
-          make clean
-          make compile
+          make dist
 
   job_upload_artifacts:
     name: Upload artifacts
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt'
-          java-version: '11'
+          java-version: '8'
 
       - name: Check cache
         uses: actions/cache@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,63 @@
+name: Build and upload artifacts
+
+on:
+  push:
+    branches:
+      - main
+      - release/**
+  pull_request:
+    branches:
+      - '*'
+
+env:
+  CACHED_BUILD_PATHS: |
+    ${{ github.workspace }}/plugin-build/build/distributions/*.zip
+
+jobs:
+  job_build:
+    name: Build the plugin
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Checkout current commit (${{ github.sha }})
+        uses: actions/checkout@v2
+
+      - name: Set up Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '11'
+
+      - name: Check cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.CACHED_BUILD_PATHS }}
+          key: ${{ github.sha }}
+
+      - name: Build the plugin
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+
+      - name: Checkout current commit (${{ github.sha }})
+        uses: actions/checkout@v2
+
+      - name: Set up Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '11'
```
marandaneto, line 29: should be java 8

.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
...
+    branches:
+      - main
+      - release/**
+  pull_request:
+    branches:
+      - '*'
+
+env:
+  CACHED_BUILD_PATHS: |
+    ${{ github.workspace }}/plugin-build/build/distributions/*.zip
```
marandaneto, line 14: we should not cache this

.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.CACHED_BUILD_PATHS }}
+          key: ${{ github.sha }}
+
+      - name: Build the plugin
+        run: |
+          cd plugin-build
+          make clean
+          make compile
```
marandaneto, line 41: I'd do `make all`
and we need to add to make all the `gradlew distZip` step too

.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+
+      - name: Checkout current commit (${{ github.sha }})
+        uses: actions/checkout@v2
+
+      - name: Check cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.CACHED_BUILD_PATHS }}
+          key: ${{ github.sha }}
```
marandaneto, line 56: I guess we dont need that either, but rather only the gradle wrapper caching

cortinico: Yup exactly. Potentially the same as:
https://github.com/getsentry/sentry-android-gradle-plugin/blob/1914df6677b4c046d711e69b28b6a1014d8612b6/.github/workflows/pre-merge.yaml#L26-L34

.github/workflows/build.yml
iker-barriocanal: Why? This cache is to keep the distribution artifact built in `job_build` available in `job_upload_artifacts`.

.github/workflows/build.yml
iker-barriocanal: `make all` also does `dryRelease`, do we want to run it on CI?

.github/workflows/build.yml
iker-barriocanal: By setting two jobs, the way to get the artifacts generated in the first one available in the second one is through cache. 

Why is the gradle wrapper caching required, if it's only built once (to generate the artifacts)? @cortinico the `pre-merge` workflow is run multiple times (not just once), so I don't think it's the same use case here.

cortinico: > By setting two jobs, the way to get the artifacts generated in the first one available in the second one is through cache.

If you need to share artifacts between jobs you should use https://github.com/actions/upload-artifact and https://github.com/actions/download-artifact.

Cache on CI is best-effort, it might get completely wiped between jobs and you should not rely on it for business critical files.

You can safely reuse the caching one from `pre-merge` as it will speedup the Gradle invocation, nothing more.

iker-barriocanal: Updated the build job to have a single job, to both build and upload the artifact. Good to know more about the cache on CI 😄

cortinico (approved):

.github/workflows/build.yml
marandaneto: not sure what to answer since I don't know what `job_build` and `job_upload_artifacts` does, I believe only `actions/upload-artifact@v2` is enough to upload the files and that's it

PR #149, opened 2021-07-22
feat(craft): Add Craft config for `maven` target

iker-barriocanal:
Adds the config to support releases using [Craft](https://github.com/getsentry/craft).

Draft PR until the release is tested end to end and https://github.com/getsentry/craft/pull/271 gets merged.

_#skip-changelog_


.craft.yml
```diff
@@ -1,8 +1,12 @@
 minVersion: 0.23.1
```
marandaneto, line 1:
```suggestion
minVersion: 0.24.4
```

.craft.yml
```diff
@@ -1,8 +1,12 @@
...
 changelogPolicy: auto
-artifactProvider:
-  name: none
 targets:
+  - name: maven
+    gradleCliPath: ./gradlew
```
marandaneto, line 5:
```suggestion
    gradleCliPath: ./plugin-build/gradlew
```

since it has to be executed within the plugin-build folder

marandaneto (approved):

PR #153, opened 2021-07-28
create alias for closeAndReleaseRepository from project root

marandaneto:
## :scroll: Description
create alias for closeAndReleaseRepository from project root

_#skip-changelog_

since we can't do `gradlew plugin-build:closeAndReleaseRepository`, because its a composite build, craft fails due to `closeAndReleaseRepository` not found in the project root, also doing `./plugin-build/gradlew closeAndReleaseRepository` also does not work since it executes the gradlew in the current folder anyway, not sure if the alias is the best way but it should fix the issue right now.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -42,3 +42,9 @@ tasks.register("preMerge") {
     dependsOn(":examples:android-gradle-kts:check")
     dependsOn(gradle.includedBuild("plugin-build").task(":check"))
 }
+
+tasks.register("closeAndReleaseRepository") {
+    description = "Runs closeAndReleaseRepository for plugin-build"
+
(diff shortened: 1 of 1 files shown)


marandaneto:
@cortinico I will merge as it is but please let me know if there's a better way to do it, Craft should be able to call `./gradlew closeAndReleaseRepository` and execute the `closeAndReleaseRepository` task from `plugin-build`, so it releases and promote the artifacts

cortinico:
> @cortinico I will merge as it is but please let me know if there's a better way to do it

The easiest way would have been to let craft call:

```
./gradlew -p plugin-build closeAndReleaseRepository
```

But the approach you used is also fine 👍

PR #156, opened 2021-07-30
Lazy configuration

remcomokveld:
## :scroll: Description
Changes the Plugin to use TaskProviders instead of tasks directly to ensure that tasks from AGP are not configured eagerly.

[Lazy Configuration](https://docs.gradle.org/current/userguide/lazy_configuration.html) "can avoid resource intensive work during the configuration phase, which can have a large impact on build performance". 

The newly added test verifies that no additional tasks are being configured when invoking the `help` task.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Enhancement: Avoid Eager Task Configuration (#156)
+
 ## 2.1.1-beta.2
 
 - No documented changes.
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -21,6 +21,7 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.plugins.ExtraPropertiesExtension
+import org.gradle.api.tasks.TaskProvider
 
 class SentryPlugin : Plugin<Project> {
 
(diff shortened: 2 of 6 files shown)


marandaneto:
@remcomokveld that's great, thanks for that.

indeed, the usage of `project.findTask(x)` makes those tasks be triggered eagerly, but not all tasks have a `TaskProvider<Task>` so we'll have to find out how to properly setup task dependencies without it, we'll have a look into it, in case you have any ideas, that would be awesome, thanks.

cortinico:
> but not all tasks have a `TaskProvider<Task>` so we'll have to find out how to properly setup task dependencies without it

Yup that's basically the problem :/ Hopefully AGP will migrate more and more to have `TaskProvider` for the tasks we depend on.

cortinico (approved):

remcomokveld:
It is not AGP that define task providers, it is a Gradle feature, and for every task that has been registered you can obtain a `TaskProvider` using `tasks.named()`.

Just pushed changes that does that, and now the test passes.

cortinico:
> and for every task that has been registered you can obtain a `TaskProvider` using `tasks.named()`.

TIL :)

Thanks for addressing this 👍 Great stuff!

cortinico (approved):

cortinico (approved):

cortinico:
@marandaneto Are you fine with merging this?

marandaneto:
> @marandaneto Are you fine with merging this?

I'm OOO till next Monday,but by what I've seen on my mobile, its LGTM, please merge it :)

PR #158, opened 2021-08-10
Do not hardcode the build/ folder

### LINKED ISSUE #157: Native symbols upload does not work when using custom buildDir
_Android Gradle Plugin:_
**4.2.1**

_Gradle:_
**6.9.0**

_Optimizer:_
- [ ]  ProGuard
- [x]  R8

- [ ]  Yes -> where?
- [ ]  No

_SDKs:_
- [ ]  sentry-android -> which version?
- [x]  sentry-react-native -> 2.6.1

---
I have the following issue:

Native symbol upload doesn't work when using a different build folder. I think [this](https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadNativeSymbolsTask.kt#L80) should use project.buildDir instead and should fix the issue, but not sure how to test this change locally.

**Steps to reproduce:**

In root build.gradle add

```groovy
allprojects {
    buildDir = System.getProperty("user.home")}/.gradle/builds/${rootProject.name}/${project.name}"
}
```

This will build the project in ~/.gradle/builds instead of the build directory in each project folder.

**Actual result:**

```
> Task :app:uploadSentryNativeSymbolsForRelease
> Found 0 debug information files
> No debug information files found
```

**Expected result:**
Native symbols that exist in the merged_native_libs folder get uploaded.

cortinico:
## :scroll: Description
This PR changes the paths hardcoded to `build/` into `project.buildDir`.

Fixes #157	

Will rely on JUnit tests to verify that everything works as usual.

## :pencil: Checklist
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Enhancement: Avoid Eager Task Configuration (#156)
+- Fix: Do not hardcode the build/ folder (#158)
 
 ## 2.1.1-beta.2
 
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
nice one, thanks!

PR #159, opened 2021-08-10
Add a NDK example to easily test the native symbol upload

cortinico:
## :scroll: Description
Adding a really simple example that relies on the NDK. This can come handy in scenarios where the upload native symbols needs to be tested.

https://github.com/getsentry/sentry-android-gradle-plugin/projects/1#card-57536559

I've added a CI step to the matrix to verify that it uploads correctly.

## :pencil: Checklist
- [x] No breaking changes

#skip-changelog


cortinico:
I believe there is a bug in the plugin, given that the uploadNativeSymbol task is not honoring the `autoUpload` flag. Ideally there should be a `--no-upload` flag added during the dry run for native symbols as well right?

examples/android-ndk/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+    ndkVersion "21.4.7075529"
+    externalNativeBuild {
+        ndkBuild {
+            path "jni/Android.mk"
+        }
+    }
+}
+
+sentry {
+    autoUpload = System.getenv("AUTO_UPLOAD")?.toBoolean() ?: false
```
marandaneto, line 40: should be false since its only meant for the `proguard` bits

examples/android-ndk/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+    uploadNativeSymbols = true
```
marandaneto, line 41: lets also enable the upload of sources

cortinico:
> we can either introduce a new boolean that controls that or find another workaround.

I've updated to run this only on the `Test Publishing UUID` workflow.

examples/android-ndk/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+        minSdkVersion 21
+        targetSdkVersion 30
+        versionCode = 1
+        versionName = "1.0"
+    }
+    buildTypes {
+        release {
+            minifyEnabled true
+            proguardFiles getDefaultProguardFile(
+                'proguard-android-optimize.txt')
```
marandaneto, line 27:
```suggestion
            minifyEnabled false
```

since its a NDK test, we can speed up the build a bit

marandaneto (approved):
nice, LGTM, thanks

.github/workflows/test-publish.yaml
```diff
@@ -32,4 +32,7 @@ jobs:
             cache-gradle-
 
       - name: Run the uploadSentryProguardMappings task
-        run: ./gradlew assembleRelease uploadSentryProguardMappingsRelease
+        run: ./gradlew assembleRelease uploadSentryProguardMappingsRelease | tee gradle.log
+
+      - name: Verify that Native Symbols were uploaded
+        run: grep "Uploaded 4 missing debug information files" gradle.log
```
marandaneto, line 38: does it fail if the message isnt found?

cortinico: Yes as `grep` will return non zero

examples/android-ndk/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+        targetSdkVersion 30
+        versionCode = 1
+        versionName = "1.0"
+    }
+    buildTypes {
+        release {
+            minifyEnabled false
+        }
+    }
+    ndkVersion "21.4.7075529"
```
marandaneto, line 28: any reason to specify the version? otherwise, AGP chooses the default version based on the AGP versioning https://developer.android.com/studio/projects/install-ndk#default-ndk-per-agp

cortinico: Yup my bad. I was playing with a couple of versions I had installed locally and i forgot to clean it up.

PR #160, opened 2021-08-18
fix: Cli downloading every Gradle sync

romtsn:
#skip-changelog

## :scroll: Description

When doing a Gradle sync in IntelliJ, the CLIs got downloaded every time for me. My guess is that because it configures **all** projects (buildSrc and root in our case), and since we use relative paths there, when configuring the root project `:` (see the screenshot), it couldn't locate the CLIs properly. So I've just changed it to rely on the `rootDir`

image

See above

## :pencil: Checklist

- [x] No breaking changes


cortinico (approved):
I've noticed the same behavior on my IDE as well. Thanks for fixing it 👍

marandaneto (approved):

PR #161, opened 2021-08-19
build(craft): Disable android on `maven` target

iker-barriocanal:
Craft requires, from https://github.com/getsentry/craft/pull/271, to set the Android config in the `maven` target. This PR updates that config.

_#skip-changelog_


marandaneto:
@iker-barriocanal should we merge this and bump `minVersion` to 0.25.0?

iker-barriocanal:
Yes. You can also bump to `v0.25.2` and get auto changelogs.

marandaneto (approved):

PR #170, opened 2021-08-27
Do not call tasks on CI, assemble tasks should be enough to trigger task execution

marandaneto:
## :scroll: Description
_#skip-changelog_

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -52,5 +52,5 @@ jobs:
       - name: Build the Debug variant
         run: ./gradlew assembleDebug
 
-      - name: Run the uploadSentryProguardMappings task
-        run: ./gradlew assembleRelease uploadSentryProguardMappingsRelease
+      - name: Build the Release variant
+        run: ./gradlew assembleRelease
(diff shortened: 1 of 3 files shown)


marandaneto:
@cortinico I guess we talked about it in the past but I don't remember the output, do you think this change makes sense?

cortinico:
> do you think this change makes sense?

I have mixed feelings about this change.

The problem with calling just `assembleRelease/Debug` is that, if by any chance there is a bug in the plugin and the `upload...` task doesn't get called, the CI will not discover it.

You will just end up assembling your APK that contains your symbols and you will never verify that the `upload...` task works correctly. Having both tasks in the invocation ensures that both tasks are executed (regardless of the task dependency).

At the same time I understand that users should not call `upload...` directly, and just assemble so it's reasonable to make the CI be as close to the real-world scenario.

> CI is failing due to "Install NDK (Side by side) 21.1.6352462 (revision: 21.1.6352462)" failed.

Seems like the CI is green no?

marandaneto:
> The problem with calling just `assembleRelease/Debug` is that, if by any chance there is a bug in the plugin and the `upload...` task doesn't get called, the CI will not discover it.

indeed, but that's why we have tests after assembling them, either checking the file inside of the APK or if it prints out the sentry-cli logging from uploading symbols, if these tests fail, we know that either the tasks didn't get trigger at all or they failed, its a good start of investigation IMO.

> Seems like the CI is green no?

yes, I rerun a few times and it worked, I guess it was just flaky in the GH images

cortinico (approved):

PR #171, opened 2021-08-30
Use gradle-build-action

### LINKED ISSUE #168: Replace manual gradlew calls to gradle-build-action
https://github.com/gradle/gradle-build-action

I believe they do caching better than us https://github.com/gradle/gradle-build-action#caching
and we could remove the usage of `actions/cache` on our end

cortinico:
## :scroll: Description
Replace manual `./gradlew` calls with `gradle/gradle-build-action` invocations

Fixes #168 

Let's let the CI tell us the truth.

## :pencil: Checklist
- [x] No breaking changes

## :crystal_ball: Next steps
There are currently two "blockers" to this:

1. `test-publish.yaml` is using `| tee gradle.log` that is not supported by this action. This is used to verify that native symbols are effectively uploaded. 
2. `build.yaml` is using a makefile to make Gradle invocations.

Therefore we end up having a mixture of Gradle invocations & caching mechanisms on Github Actions. I'm unsure if we want to go this way or not.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,6 @@ jobs:
     name: Build and upload distribution artifact
     runs-on: ubuntu-latest
     steps:
-
       - name: Checkout current commit (${{ github.sha }})
         uses: actions/checkout@v2
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,15 +15,9 @@ jobs:
 
       - name: Checkout current commit (${{ github.sha }})
         uses: actions/checkout@v2
-      - name: Cache Gradle Files
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.gradle/caches/
-            ~/.gradle/wrapper/
-          key: cache-gradle-${{ matrix.os }}
-          restore-keys: |
-            cache-gradle-
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2.1
 
       - name: Set up Java
         uses: actions/setup-java@v2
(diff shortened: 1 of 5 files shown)


marandaneto:
oh indeed, the makefile could be gone in favor of using the official action but the `| tee gradle.log` can't, let's hold this a bit then.
maybe it's worth opening an issue in the GH action that allows us to `tee` the output either via an stream or file.

cortinico:
Opened an issue here https://github.com/gradle/gradle-build-action/issues/72

bigdaz:
Please try again with `gradle-build-action@v2.1`. You should be able to keep your existing workflow, simply adding an initial step to setup-gradle with caching and build-scan integration.

marandaneto:
> Opened an issue here [gradle/gradle-build-action#72](https://github.com/gradle/gradle-build-action/issues/72)

this is now unblocked :)

romtsn:
@cortinico PTAL, I've updated the action usage everywhere with the new version

cortinico:
> @cortinico PTAL, I've updated the action usage everywhere with the new version

Thanks for doing that. I had this in my backlog to do but I'm a bit low on bandwidth at the moment 🙏  Waiting for a green CI to stamp ✅

romtsn:
Assuming y'all are satisfied with the green CI I just merged :) can follow up if smth is needed

cortinico:
All good for me

romtsn:
Aaaaand it's red indeed :D looks like jcenter is down https://status.gradle.com/

gimp87:
plugin-build/expected-checksums.sha

PR #172, opened 2021-08-30
Run publish also on merges

### LINKED ISSUE #169: test-publish.yaml is always being skipped
release: https://github.com/getsentry/sentry-android-gradle-plugin/runs/3440741944?check_suite_focus=true
pr: https://github.com/getsentry/sentry-android-gradle-plugin/runs/3440735398?check_suite_focus=true

could it be because of the check `github.event.pull_request.head.repo.full_name == github.repository`?

cortinico:
## :scroll: Description
This relaxes the `test-publish` condition to run if the owner is `getsentry` (either from a local PR or a merge).

Fixes #169 

Will check if the PR is green just after merging this

## :pencil: Checklist
- [x] No breaking changes

#skip-changelog

### DIFF
--- a/.github/workflows/test-publish.yaml
+++ b/.github/workflows/test-publish.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   publish:
-    if: ${{ github.repository == 'getsentry/sentry-android-gradle-plugin'}}
+    if: ${{ github.repository == 'getsentry/sentry-android-gradle-plugin' || github.event.pull_request.head.repo.full_name == 'getsentry/sentry-android-gradle-plugin'}}
     runs-on: ubuntu-latest
 
     env:
--- a/.github/workflows/test-publish.yaml
+++ b/.github/workflows/test-publish.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   publish:
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    if: ${{ github.repository == 'getsentry/sentry-android-gradle-plugin'}}
     runs-on: ubuntu-latest
 
     env:
--- a/.github/workflows/test-publish.yaml
+++ b/.github/workflows/test-publish.yaml
@@ -11,7 +11,6 @@ on:
 
 jobs:
   publish:
-    if: ${{ github.repository == 'getsentry/sentry-android-gradle-plugin' || github.event.pull_request.head.repo.full_name == 'getsentry/sentry-android-gradle-plugin'}}
     runs-on: ubuntu-latest
 
     env:
@@ -33,6 +32,8 @@ jobs:
 
       - name: Build the Release variant
         run: ./gradlew assembleRelease | tee gradle.log
+        if: env.SENTRY_AUTH_TOKEN != null
 
       - name: Verify that Native Symbols were uploaded
         run: grep "Uploaded 4 missing debug information files" gradle.log
+        if: env.SENTRY_AUTH_TOKEN != null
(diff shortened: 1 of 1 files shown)


.github/workflows/test-publish.yaml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   publish:
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    if: ${{ github.repository == 'getsentry/sentry-android-gradle-plugin'}}
```
marandaneto, line 14: right now it causes:

> caused by: sentry reported an error: Invalid token header. No credentials provided. (http status: 401)

we want to skip it in such cases, right? and not fail.

cortinico: Yes and in such cases `github.repository` should be `'cortinico/sentry-android-gradle-plugin'`.
Seems like it's not the case actually

cortinico:
It seems like there is no easy way to skip the job if the secrets are not available: https://github.community/t/how-can-i-test-if-secrets-are-available-in-an-action/17911

I've opted for skipping the jobs steps instead

marandaneto:
the downside with the current implementation is that the repo is gonna be cloned, cache restored, but the execution skipped which is a waste of resources, I will have a quick look as well before merging

cortinico:
> but the execution skipped which is a waste of resources, I will have a quick look as well before merging

We can add a `if()` condition to all the steps. I wasn't able to find a better solution. The alternative would be to split this workflow in two. 

Or to merge this workflow inside the `test-publish-dry-run`, so that it runs the real publishing only in certain cases.

marandaneto:
> > but the execution skipped which is a waste of resources, I will have a quick look as well before merging
> 
> We can add a `if()` condition to all the steps. I wasn't able to find a better solution. The alternative would be to split this workflow in two.
> 
> Or to merge this workflow inside the `test-publish-dry-run`, so that it runs the real publishing only in certain cases.

inside the dry-run not sure, since it's a `dry` run, it'd be fine to be within another workflow though, I guess.

I just noticed that https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/workflows/pre-merge.yaml#L20-L21 also has the token but there's no condition, so if its required, its gonna fail when opening from fork PRs I guess?

but a condition to all the steps solve the solution for now as well, wdyt?

cortinico:
> inside the dry-run not sure, since it's a `dry` run, it'd be fine to be within another workflow though, I guess.

I was more thinking about having just a "publish" workflow that will run dry-run on forks, and dry-run + real publishing on merge/local branches.

> I just noticed that [https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/workflows/pre-merge.yaml#L20-L21](https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/workflows/pre-merge.yaml?rgh-link-date=2021-08-31T10%3A16%3A50Z#L20-L21) also has the token but there's no condition, so if its required, its gonna fail when opening from fork PRs I guess?

It won't fail as the `preMerge` task is not setting `AUTO_UPLOAD: true` so it will effectively not uplaod anything and don't need the token (I guess it could be removed from the `pre-merge` workflow).

> but a condition to all the steps solve the solution for now as well, wdyt?

Yeah it works for now 👍 I'm adding it.

marandaneto:
nice, thanks @cortinico 

just created https://github.com/getsentry/sentry-android-gradle-plugin/pull/177 so we remove the unnecessary token from the pre-merge repo

PR #173, opened 2021-08-30
Verifies that the CLI is bundled correctly

### LINKED ISSUE #163: Add e2e test for published version
Not sure if it's possible since maven central has some sort of delay after publishing a version.
the idea is that we e2e test a published version and detect issues in advance.

https://github.com/getsentry/sentry-android-gradle-plugin/releases/tag/2.1.1 didn't embed sentry-cli due to https://github.com/getsentry/sentry-android-gradle-plugin/pull/162 but we needed to be notified about the bug instead of identifying it thru automatized tests.

if not possible to do an e2e test, we can just add a new unit test that verifies that sentry-cli is embedded within the compiled jar

cortinico:
## :scroll: Description
I've moved the CLI downloader from `buildSrc` to `plugin-build`.
Now I've plugged it as a dependency for `processResources` so it should be "harder" to ship a version without the CLI bundled. 

Fixes #163

I've run the bash script locally.

## :pencil: Checklist
- [x] No breaking changes

## :crystal_ball: Next steps
I'm unsure if we want to add further testing for this.

#skip-changelog


cortinico:
Now this is ready to go 👍 In the CI log you'll see if the binary are actually included or not

marandaneto (approved):

marandaneto:
@cortinico that's great, thanks a lot for doing this

marandaneto:
> I'm unsure if we want to add further testing for this.

I'd say that the CI test covers the bug we have had, which unzips the file and see if sentry CLI is in it

Issue #109, opened 2021-05-21
Support new Android Gradle APIs


marandaneto:
We could improve how to find the mapping file using the latest APIs
https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryMappingFileProvider.kt

https://developer.android.com/reference/tools/gradle-api/4.1/com/android/build/api/artifact/ArtifactType.OBFUSCATION_MAPPING_FILE

https://medium.com/androiddevelopers/new-apis-in-the-android-gradle-plugin-f5325742e614

this won't work for people using AGP 4.0 though, only >= 4.1

marandaneto:
https://github.com/ReactiveCircus/app-versioning/blob/b30fb5435b25ad7fb8acdf61179c7010f4074fdb/src/main/kotlin/io/github/reactivecircus/appversioning/AppVersioningPlugin.kt uses the `onVariants` API, we can learn from there

marandaneto:
variant API is now stable: https://developer.android.com/studio/releases/gradle-plugin#7-0-0
https://github.com/android/gradle-recipes/tree/agp-7.0/Kotlin

cortinico:
> variant API is now stable: [developer.android.com/studio/releases/gradle-plugin#7-0-0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0)
> [https://github.com/android/gradle-recipes/tree/agp-7.0/Kotlin](https://github.com/android/gradle-recipes/tree/agp-7.0/Kotlin?rgh-link-date=2021-07-29T08%3A52%3A49Z)

Awesome 🎉 
I believe we want to try to land this https://github.com/getsentry/sentry-android-gradle-plugin/pull/106 first right?

marandaneto:
@cortinico yeah landing https://github.com/getsentry/sentry-android-gradle-plugin/pull/106 would be nice, but if migrating to AGP 7.x is painless, I'd not mind ignoring it since AGP 4.1 to 4.2 does not have any benefits by the final user.

AGP 7.x though introduces breaking changes, but we can always bump a major and recommend people using the v2.x when using AGP < 7.x

what do you think? I'm fine either way, right now I believe our plugin 2.x works pretty well with AGP 7.x right? as we run tests on the beta version as well, so not an urgent issue to work on either.

marandaneto:
[...]

PR #174, opened 2021-08-31
feat: transform api agp

romtsn:
#skip-changelog

## :scroll: Description

* Bumps AGP version to 7.0+ to use new [transform APIs](https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/variant/Component.html#transformclasseswith)
* Adds basic structure and classes for bytecode instrumentation

https://github.com/getsentry/sentry-java/issues/1571

Manually for now

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -1,5 +1,5 @@
 object BuildPluginsVersion {
-    val AGP = System.getenv("VERSION_AGP") ?: "4.2.2"
+    val AGP = System.getenv("VERSION_AGP") ?: "7.0.1"
     const val DOKKA = "1.4.32"
     const val KOTLIN = "1.4.32"
     const val KTLINT = "10.0.0"
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -1,6 +1,10 @@
 package io.sentry.android.gradle
 
+import com.android.build.api.instrumentation.FramesComputationMode
+import com.android.build.api.instrumentation.InstrumentationScope
+import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.gradle.AppExtension
+import com.android.build.gradle.internal.utils.setDisallowChanges
 import io.sentry.android.gradle.SentryCliProvider.getSentryCliPath
 import io.sentry.android.gradle.SentryPropertiesFileProvider.getPropertiesFilePath
 import io.sentry.android.gradle.SentryTasksProvider.getAssembleTaskProvider
@@ -11,6 +15,7 @@ import io.sentry.android.gradle.SentryTasksProvider.getPackageBundleTask
 import io.sentry.android.gradle.SentryTasksProvider.getPackageProvider
 import io.sentry.android.gradle.SentryTasksProvider.getPreBundleTask
 import io.sentry.android.gradle.SentryTasksProvider.getTransformerTask
+import io.sentry.android.gradle.instrumentation.SpanAddingClassVisitorFactory
 import io.sentry.android.gradle.tasks.SentryGenerateProguardUuidTask
 import io.sentry.android.gradle.tasks.SentryUploadNativeSymbolsTask
 import io.sentry.android.gradle.tasks.SentryUploadProguardMappingsTask
@@ -23,6 +28,7 @@ import org.gradle.api.Task
 import org.gradle.api.plugins.ExtraPropertiesExtension
 import org.gradle.api.tasks.TaskProvider
 
+@Suppress("UnstableApiUsage")
 class SentryPlugin : Plugin<Project> {
 
     private val sep = File.separator
@@ -35,6 +41,8 @@ class SentryPlugin : Plugin<Project> {
         )
         project.pluginManager.withPlugin("com.android.application") {
             val androidExtension = project.extensions.getByType(AppExtension::class.java)
+            val androidComponentsExtension =
+                project.extensions.getByType(AndroidComponentsExtension::class.java)
             val cliExecutable = getSentryCliPath(project)
 
             val extraProperties = project.extensions.getByName("ext")
@@ -47,12 +55,23 @@ class SentryPlugin : Plugin<Project> {
                 extraProperties.get(SENTRY_PROJECT_PARAMETER).toString()
             }.getOrNull()
 
+            androidComponentsExtension.onVariants { variant ->
+                variant.transformClassesWith(
+                    SpanAddingClassVisitorFactory::class.java,
+                    InstrumentationScope.ALL
+                ) { params ->
+                    if (extension.forceInstrumentDependencies.get()) {
+                        params.invalidate.setDisallowChanges(System.currentTimeMillis())
+                    }
+                }
+                variant.setAsmFramesComputationMode(FramesComputationMode.COPY_FRAMES)
+            }
+
             androidExtension.applicationVariants.matching {
                 it.name !in extension.ignoredVariants.get() &&
                     it.flavorName !in extension.ignoredFlavors.get() &&
                     it.buildType.name !in extension.ignoredBuildTypes.get()
             }.configureEach { variant ->
-
                 val bundleTask = withLogging(project.logger, "bundleTask") {
                     getBundleTask(project, variant.name)
                 }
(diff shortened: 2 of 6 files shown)


romtsn:
@cortinico you can only skim through it for the gradle part, no need to go deep into the bytecode stuff :)

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPluginExtension.kt
```diff
@@ -47,4 +47,13 @@ abstract class SentryPluginExtension @Inject constructor(project: Project) {
...
     val ignoredFlavors: ListProperty<String> = objects.listProperty(String::class.java)
         .convention(emptyList())
+
+    /**
+     * Forces dependencies instrumentation, even if they were already instrumented.
+     * Useful when there are issues with code instrumentation, e.g. the dependencies are
+     * partially instrumented.
+     * Defaults to false.
+     */
+    val forceInstrumentDependencies: Property<Boolean> = objects.property(Boolean::class.java)
```
romtsn, line 57: I am not sure if we should expose this to the end users, as it only makes sense for devs of the plugin (from my understanding). But I'd probably tackle this later, when we figure out what configuration should we provide to the users

cortinico:
> * Bumps AGP version to 7.0+ to use new [transform APIs](https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/variant/Component.html#transformclasseswith)

Could we export the AGP bump to a separate PR?

romtsn:
@cortinico sure, do you want me to target `main` in that PR? Not sure if there are other consequences we need to consider though

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -1,5 +1,5 @@
 object BuildPluginsVersion {
-    val AGP = System.getenv("VERSION_AGP") ?: "4.2.2"
+    val AGP = System.getenv("VERSION_AGP") ?: "7.0.1"
```
marandaneto, line 2: this change would be ideally done with #109 since there are new APIs, also requires bumping a major version of the plugin

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
```diff
@@ -0,0 +1,21 @@
+package io.sentry.android.gradle.instrumentation
+
+import io.sentry.android.gradle.instrumentation.database.AndroidXSQLiteClassVisitor
+import org.objectweb.asm.ClassVisitor
+
+enum class Instrumentable(val fqName: String) {
+    ANDROIDX_SQLITE_DATABASE("androidx.sqlite.db.framework.FrameworkSQLiteDatabase");
```
marandaneto, line 7: does the instrumentation happen before proguard/r8? wondering if the class would be renamed if so

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
```diff
@@ -0,0 +1,21 @@
+package io.sentry.android.gradle.instrumentation
+
+import io.sentry.android.gradle.instrumentation.database.AndroidXSQLiteClassVisitor
+import org.objectweb.asm.ClassVisitor
+
+enum class Instrumentable(val fqName: String) {
```
marandaneto, line 6: curious, why is it an enum?

buildSrc/src/main/java/Dependencies.kt
romtsn: Yeah, but not sure I want to switch to those other changes rn. I just did it for the purpose of trying out the new api, which turned out to be working :) Maybe I'll finish the instrumentation first, and then if it's not tackled by then, I will implement #109 properly? This branch is anyway targeting an integration branch and no `main`

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
romtsn: That's a good point, I also have it in my to-do to test it with r8, thought to postpone it for later, but will check it tomorrow.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
romtsn: Hm, cause we need to do a `when` over it and it has a finite number of classes that we want to instrument.. I'm open for suggestions though, it could also be just a `Map<String, ClassVisitor>` I think, is that what you meant?

cortinico:
> @cortinico sure, do you want me to target `main` in that PR? Not sure if there are other consequences we need to consider though

I believe we want to do as @marandaneto suggested: potentially bump minor or major version. Plus the tests and the CI matrix needs to be updated to reflect this (like the default job should be the one on AGP 7.x).

If "it works" without any code changes, it's great :) We can switcht to using newer APIs in a subsequent PR.

buildSrc/src/main/java/Dependencies.kt
marandaneto: yep, that sounds great, but we have to work on #109 before releasing it, otherwise, we'd be mixing deprecated APIs and non-stable APIs in the same version, not the best combination.
obviously this could be done in another PR at a later point

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -47,12 +55,23 @@ class SentryPlugin : Plugin<Project> {
...
+                variant.transformClassesWith(
+                    SpanAddingClassVisitorFactory::class.java,
+                    InstrumentationScope.ALL
+                ) { params ->
+                    if (extension.forceInstrumentDependencies.get()) {
+                        params.invalidate.setDisallowChanges(System.currentTimeMillis())
+                    }
+                }
+                variant.setAsmFramesComputationMode(FramesComputationMode.COPY_FRAMES)
+            }
```
marandaneto, line 68: this part should also respect the `ignoredVariants`, `ignoredFlavors` and `ignoredBuildTypes` but we can do it later together with the new APIs of AGP 7

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
marandaneto: yeah it will make more sense when we have more classes, wfm

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
romtsn: I will drop a TODO here and then have a list of checkboxes in the final PR to `main` to resolve this

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
marandaneto: btw whats `fq` stands for in `fqName`? maybe a more descriptive name?

marandaneto (approved):

buildSrc/src/main/java/Dependencies.kt
cortinico: So are we doing the bump to AGP 7 in this PR?

marandaneto: yes and no, this PR merges to the feat/bytecode-instrumentation branch and @romtsn will work on #109 in parallel, since the bytecode man. depends on #109

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
romtsn: tested with r8 - it worked. The transformation step runs as part of the `minifyWithR8` gradle task, and it internally makes sure to run custom transformations before the obfuscation

> btw whats `fq` stands for in `fqName`? maybe a more descriptive name?

fq stand for fully-qualified name and it's kind of a common standard for this -> https://www.google.com/search?q=fqname. Kotlin, Google, others are using it. But if you will, I can change that, though it will look a bit long to me

buildSrc/src/main/java/Dependencies.kt
romtsn: Exactly, will open a new PR to `main` with just the bump and later follow up with another PRs to switch to new APIs if not done by that time yet

PR #180, opened 2021-09-02
feat: Instrument Room calls

romtsn:
## :scroll: Description

This is just a draft PR/integration branch for all the changes related to bytecode instrumentation. All other PRs will be targeting this one.

## :bulb: Motivation and Context

https://github.com/getsentry/sentry-java/issues/1571

## :pencil: Checklist

- [ ] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [ ] No breaking changes

## :crystal_ball: Next steps
Some TODOs left:
- [ ] Consider `ignoredVariants/ignoreBuildTypes/ignoredFlavours` when applying the instrumentation
- [ ] Properly switch to AGP 7.0+ (with using new API #109)
- [ ] Automated tests (Unit/Integration)
- [ ] Figure out what configuration do we want to provide for the SentryPlugin
- [ ] Test this out with different projects, using the same library under instrumentation, since Gradle shares them transformed between projects

PR #181, opened 2021-09-02
Bump AGP to 7.0.1

### LINKED ISSUE #179: Set plugin and examples to Java Target and Source to 1.8
So we can compile using Java 11 but not causing compatibility issues
all the https://github.com/getsentry/sentry-android-gradle-plugin/tree/main/.github/workflows could also be changed to set up Java 11 and the matrix could be also only Java 11

romtsn:
## :scroll: Description

Part of #109 and #180 
Closes #179

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt'
-          java-version: '8'
+          java-version: '11'
 
       - name: Build distribution artifacts
         run: |
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,6 +28,25 @@ subprojects {
             include("**/kotlin/**")
         }
     }
+
+    tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+        sourceCompatibility = JavaVersion.VERSION_1_8.toString()
+        targetCompatibility = JavaVersion.VERSION_1_8.toString()
+
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8.toString()
+            freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
+            languageVersion = "1.3"
+        }
+    }
+
+    plugins.withId("com.android.application") {
+        val ext = extensions.getByName("android") as com.android.build.gradle.AppExtension
+        ext.compileOptions {
+            sourceCompatibility = JavaVersion.VERSION_1_8
+            targetCompatibility = JavaVersion.VERSION_1_8
+        }
+    }
 }
 
 tasks.register("clean", Delete::class.java) {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ subprojects {
         }
     }
 
+    // to be compatible with AGP <= 4.1.x https://developer.android.com/studio/releases/gradle-plugin#java-8-default
     tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
         sourceCompatibility = JavaVersion.VERSION_1_8.toString()
         targetCompatibility = JavaVersion.VERSION_1_8.toString()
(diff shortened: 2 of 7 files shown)


romtsn:
This is a set of minimal changes for AGP 7.0 support, lemme know if I missed something. The rest can be followed up later via separate PRs

.github/workflows/build.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt'
-          java-version: '8'
+          java-version: '11'
```
marandaneto, line 34: this change depends on https://github.com/getsentry/sentry-android-gradle-plugin/issues/179

romtsn: https://github.com/getsentry/sentry-android-gradle-plugin/pull/181/commits/86a4fd1c1e75b48536c6313403461fc8767ccec2

marandaneto:
I'd check `No breaking changes` since we are only bumping a compile-time dependency and it does not break the feature when executing by older versions (For now)

marandaneto (approved):
LGTM

build.gradle.kts
```diff
@@ -28,6 +28,25 @@ subprojects {
...
+    tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+        sourceCompatibility = JavaVersion.VERSION_1_8.toString()
+        targetCompatibility = JavaVersion.VERSION_1_8.toString()
+
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8.toString()
+            freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
+            languageVersion = "1.3"
+        }
+    }
```
cortinico, line 41: I'm unsure about this to be honest.
Specifically the `languageVersion` and the opt-in flag should stay inside the included build (and not in the outer build). I'm afraid they won't be passed over.

As a side note, I would refrain from adding those customization inside the `subprojects{}` block.

romtsn: Yeah I agree about the flag and the languageVersion - will move it back, but why would you refrain from adding it in the subprojects block at all?

cortinico: >  will move it back, but why would you refrain from adding it in the subprojects block at all?

Mostly because:
- This project is setup as a composite build, and `subprojects{}` is not running on included builds. That could be confusing for who's reading the build file.
- I would try to keep the compiler customization as closer as possible to the source code where they're needed. From this diff I can't say anymore why `-Xopt-in=kotlin.RequiresOptIn` is needed. While if we keep it closer to the module where it's effectively needed, it will make it easier to understand/cleanup afterwords.
- Generally `subprojects{}` creates a tight couplign between the projects. I think it's ok to use as long as you setup linter/formatter/test-runners or other tools that you effectively want to apply for every module. But as soon as you need to specify some build specific settings, you're losing on isolation as soon as you add configuration there

cortinico: As a side note: I believe we can leave the `sourceCompatibility` and `targetCompatibility` in the `subprojects{}` block as they should be applied to all the project uncoditionally.

marandaneto: > This project is setup as a composite build, and subprojects{} is not running on included builds

because of that, I agree with @cortinico

build.gradle.kts
```diff
@@ -28,6 +28,25 @@ subprojects {
...
+
+    plugins.withId("com.android.application") {
+        val ext = extensions.getByName("android") as com.android.build.gradle.AppExtension
+        ext.compileOptions {
+            sourceCompatibility = JavaVersion.VERSION_1_8
+            targetCompatibility = JavaVersion.VERSION_1_8
+        }
+    }
```
marandaneto, line 49: let's add a comment that this is only needed if AGP <= 4.1.x since AGP 4.2 is the default value https://developer.android.com/studio/releases/gradle-plugin#java-8-default

build.gradle.kts
romtsn: Yeah, it was clear for the compiler options where is their better place, but I didn't know that subprojects wasn't applied for composite builds. Thanks! Moved it back

PR #182, opened 2021-09-06
feat: Instrument query method overload for API 16 and above

romtsn:
## :scroll: Description

Instrument `query` method [overload](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:sqlite/sqlite-framework/src/main/java/androidx/sqlite/db/framework/FrameworkSQLiteDatabase.java;l=193-206?q=frameworksqlitedatabase&ss=androidx) that can be called on API16 and above
image

Add a basic structure for writing unit tests for verifying the bytecode instrumentation - will expand on it later, it will be a ParameterizedTest in the end, most likely.

Manually + a unit test for bytecode verification

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/build.gradle.kts
```diff
@@ -18,6 +19,8 @@ repositories {
     google()
 }
 
+
+val testImplementationAar by configurations.getting
```
marandaneto, line 23:
```suggestion
val testImplementationAar by configurations.getting

```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/Instrumentable.kt
```diff
@@ -1,21 +1,28 @@
 package io.sentry.android.gradle.instrumentation
 
-import io.sentry.android.gradle.instrumentation.database.AndroidXSQLiteClassVisitor
-import org.objectweb.asm.ClassVisitor
+interface Instrumentable<Visitor> {
```
romtsn, line 3: I figured we can't really have it in the old way with enums, cause we'd need to create enums for each type (class, method, field, annotation, etc. visitors), so I switched to an interface here that all those can implement

plugin-build/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
     id("java-gradle-plugin")
     id("com.vanniktech.maven.publish") version BuildPluginsVersion.MAVEN_PUBLISH apply false
     id("org.jlleitschuh.gradle.ktlint") version BuildPluginsVersion.KTLINT
+    id("com.stepango.aar2jar") version BuildPluginsVersion.AAR_2_JAR
```
romtsn, line 13: Since gradle plugin is a pure java project, I had to use this plugin to be able to include .aar dependencies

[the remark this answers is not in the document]
romtsn: Just a left a [comment](https://github.com/getsentry/sentry-android-gradle-plugin/pull/182#discussion_r702908253) below, will probably drop a comment in the build.gradle for better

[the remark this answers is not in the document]
marandaneto: answered https://github.com/getsentry/sentry-android-gradle-plugin/pull/182#discussion_r702908253

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/database/sqlite/AndroidXSQLiteDatabaseVisitorTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.util.CheckClassAdapter
+import java.io.FileInputStream
+import java.io.PrintWriter
+import java.io.StringWriter
+
+class AndroidXSQLiteDatabaseVisitorTest {
+
+    @Test
+    fun `instrumented FrameworkSQLiteDatabase class passes Java verifier`() {
```
marandaneto, line 16: if all the other test methods are kinda similar to this one, we could have a TestUtil class that does the job, eg, taking a `String className` and returns a pair of classVisitor and classWriter byte array

marandaneto: not necessary for this PR unless we add more tests

romtsn: Yep I will convert it into a parameterized test I think, when I instrument other classes 👍

cortinico:
I'm unsure about the 20Mb `android.jar` you're adding to the repo here. Is there no way to fetch that from any repository? Is it something that robolectric could help replace?

romtsn:
@cortinico this is technically not an android project and the bytecode instrumentation has nothing to do with android per se. I'd need to apply AGP here to make it work, since robolectric works in conjunction with the `android {}` configuration block in build.gradle - I am not sure we want this. I just treat the android-sdk as any other dependency here (e.g. AGP itself is a dependency).

I wasn't able to find where can I easily pull the latest android sdk from, in theory we could set up a bootstrap task, that pulls this [repo](https://cs.android.com/android/platform/superproject/+/master:prebuilts/sdk/) in and copies over the necessary .jar to the libs folder, but not sure if it's much better, and if I want to spend time on this now.

marandaneto:
@romtsn the android dependency is `testRuntimeOnly` so I believe the final release does not include the `android.jar` file right? cus usually, the libs filter is copied over anyway, let's guarantee that it's not the case.
I guess I already have seen some sort of maven mirror from https://cs.android.com/android/platform/superproject/+/master:prebuilts/sdk/ or something similar, where you could pull the android.jar remotely but I don't remember where either.

cortinico:
> I wasn't able to find where can I easily pull the latest android sdk from, in theory we could set up a bootstrap task, that pulls this [repo](https://cs.android.com/android/platform/superproject/+/master:prebuilts/sdk/) in and copies over the necessary .jar to the libs folder, but not sure if it's much better, and if I want to spend time on this now.

Yup I see your point. I believe the easiest/fastest would be add a

```
implementation(files("${System.getenv("ANDROID_HOME")}/platforms/android-30/android.jar"))
```

or something along those lines.

romtsn:
Okie, I've added the task to bootstrap the sdk path in https://github.com/getsentry/sentry-android-gradle-plugin/pull/182/commits/346a6875e120a47087d678395f507b64bd6f4d97, please check it out. Hopefully this works on CI as well

romtsn:
I guess I should support both `ANDROID_HOME` and `ANDROID_SDK_ROOT`, the same way as AS does it. https://developer.android.com/studio/command-line/variables#envar

cortinico (approved):

PR #183, opened 2021-09-08
feat: Instrumentation logging

romtsn:
## :scroll: Description

* Added logging capability for instrumentation output
* Added a new `debug` property for sentry plugin extension
* Adapted tests to print the log output into console

For the end users it will look like dis - a `build/tmp/sentry` folder with a log file per instrumented class:
image

## :pencil: Checklist

- [ ] No breaking changes


marandaneto:
@romtsn looks like this PR has changes from another PR, can we target a different branch then to make it easier for review?

romtsn:
@marandaneto yeah I fucked up the rebase, please take a look now

marandaneto:
if we ever implement the Sentry SDK within the plugin, we could send such files as attachments when user has given the consent https://github.com/getsentry/sentry-android-gradle-plugin/pull/126

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -55,6 +55,10 @@ class SentryPlugin : Plugin<Project> {
                 extraProperties.get(SENTRY_PROJECT_PARAMETER).toString()
             }.getOrNull()
 
+            // temp folder for sentry-related stuff
+            val tempDir = File("${project.buildDir}${sep}tmp${sep}sentry")
```
cortinico, line 59:
```suggestion
            val tmpDir = File("${project.buildDir}${sep}tmp${sep}sentry")
```

For consistency

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -55,6 +55,10 @@ class SentryPlugin : Plugin<Project> {
                 extraProperties.get(SENTRY_PROJECT_PARAMETER).toString()
             }.getOrNull()
 
+            // temp folder for sentry-related stuff
```
cortinico, line 58: Should this be created with `Files.createTempDir()` instead?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPluginExtension.kt
```diff
@@ -56,4 +56,11 @@ abstract class SentryPluginExtension @Inject constructor(project: Project) {
      */
     val forceInstrumentDependencies: Property<Boolean> = objects.property(Boolean::class.java)
         .convention(false)
+
+    /**
+     * Enabled debug output of the plugin. Useful when there are issues with code instrumentation,
+     * shows the modified bytecode.
+     * Defaults to false.
+     */
+    val debug: Property<Boolean> = objects.property(Boolean::class.java).convention(false)
```
cortinico, line 65: As a side note, this is a `debug` option but only for code instrumentation. So `debugInstrumentation` (or similar) would be a better name.

Just `debug` gives the false feeling that you're enabling `debug` debugging for the whole plugin (which would be a nice feature but it's not implemented).

romtsn: what else do you think is missing if it was a `debug` for the whole plugin?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
romtsn: hm, we want this specifically to be placed into `build/tmp/sentry` folder of the target project, so the users can access it if necessary - not sure if `createTempDir` is suitable for that

marandaneto: if the idea is that they collect those files in case there's a problem and we have to investigate it, `build/tmp/sentry` makes it easier

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPluginExtension.kt
marandaneto: since we already use `org.gradle.api.Logger` everywhere already and already has logging capabilities, I'd also go with `debugInstrumentation` and don't spend time implementing a file logging for the whole plugin

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPluginExtension.kt
```diff
@@ -62,5 +62,5 @@ abstract class SentryPluginExtension @Inject constructor(project: Project) {
      * shows the modified bytecode.
```
cortinico, line 62: The doc should also be updated accordingly

PR #184, opened 2021-09-08
feat: ExecSQL method instrumentation

romtsn:
## :scroll: Description

* Adds `execSQL` method instrumentation, which can show us such queries as `CREATE TABLE` and so on

image

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/database/sqlite/visitor/AbstractSQLiteDatabaseMethodVisitor.kt
```diff
@@ -0,0 +1,137 @@
+package io.sentry.android.gradle.instrumentation.database.sqlite.visitor
+
+import org.objectweb.asm.Label
+import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Opcodes
+
+abstract class AbstractSQLiteDatabaseMethodVisitor(
```
romtsn, line 7: I tried to consolidate the common things under this abstract class, but am still not sure if it's the best way. I guess I will figure it out later when I instrument other classes/libs/modules and refactor if necessary

marandaneto (approved):

PR #190, opened 2021-09-27
feat: Instrument Room generated classes (transactions)

romtsn:
## :scroll: Description

* Adds custom Gradle task to ASMify specified files (used to ASMify synthetic classes in Daos)
* Adds room instrumentable to instrument room dao queries with transactions (which include insert/update/delete by default + those annotated with `@Transaction`)
* Room instrumentable is currently disabled for the plugin, because we don't yet support all cases and it will not let us build

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/build.gradle.kts
```diff
@@ -134,3 +135,5 @@ tasks.withType<Test> {
         showStandardStreams = true
     }
 }
+
+tasks.register<io.sentry.android.gradle.internal.ASMifyTask>("asmify")
```
cortinico, line 139: Why do you do this?

romtsn: wdym? why am I adding a new task, or why am I registering it here?

cortinico: Sorry. I was wondering why you're registering a new task of the `ASMifyTask` type inside the `plugin-build` build file

romtsn: I thought it's the best place for this task, cause the `plugin-build` has ASM dependencies in the classpath, which we can utilise when asmifying some classes. AsmifyTask itself is a little nifty helper, that allows us to get the asmified code of any `.class` file.

marandaneto: do we need this so the plugin could do its job or is it related to the samples or tests?

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -21,14 +22,10 @@ object Libs {
     val ASM = "org.ow2.asm:asm-util:${LibsVersion.ASM}"
     val SQLITE = "androidx.sqlite:sqlite:${LibsVersion.SQLITE}"
     val SQLITE_FRAMEWORK = "androidx.sqlite:sqlite-framework:${LibsVersion.SQLITE}"
+    val SENTRY_ANDROID = "io.sentry:sentry-android:${LibsVersion.SENTRY}"
```
marandaneto, line 25: why moving from Samples to Libs? the SDK is still used only by the samples and/or tests

plugin-build/buildSrc/src/main/kotlin/io/sentry/android/gradle/internal/ASMifyTask.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.options.Option
+import org.gradle.api.tasks.options.OptionValues
+import java.io.File
+import java.io.FileOutputStream
+import java.util.*
+
+/**
+ * A task which outputs a asmified version of a given .class file. This is useful, when
+ * the ASM IntelliJ Plugin fails to decompile nested synthetic classes (e.g. lambdas).
```
marandaneto, line 13: if I got the description correctly, this should not be part of the plugin-build then, right, since its not required to do the bytecode man work.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/RoomCallMethodVisitor.kt
```diff
@@ -0,0 +1,130 @@
...
+    api = api,
+    methodVisitor = methodVisitor,
+    descriptor = descriptor
+) {
+
+    private val label5 = Label()
+    private val label6 = Label()
+    private val label7 = Label()
+    private val label8 = Label()
+    private val label9 = Label()
```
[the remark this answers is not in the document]
romtsn: mm, I think you mean that I extracted them into the abstract class rather, but we still need to use labels. This visitor is just different from those others, since we need to inject a catch-block into the existing try-finally, hence more labels and jumps

buildSrc/src/main/java/Dependencies.kt
romtsn: I thought it would look weird to use Samples.Room.Sentry dependency in the plugin-build (which is an included build, and in theory should be agnostic of other composite build's projects)

plugin-build/buildSrc/src/main/kotlin/io/sentry/android/gradle/internal/ASMifyTask.kt
romtsn: I mean, it's not strictly required, but it's a helper task for bytecode instrumentation, I still used it when developing. I'm open to move it somewhere else, I just don't know what's the better place, since the task is related to bytecode instrumentation, plus, it also uses asm dependencies from the plugin-build classpath

buildSrc/src/main/java/Dependencies.kt
marandaneto: makes sense if we need in the plugin-build, maybe we have to decide about this first -> https://github.com/getsentry/sentry-android-gradle-plugin/pull/190#discussion_r721611650

plugin-build/buildSrc/src/main/kotlin/io/sentry/android/gradle/internal/ASMifyTask.kt
marandaneto: got it, the main point is, this is used for developing the plugin and not executing the plugin.
the class is in the `buildSrc` folder tho, so this class isn't packaged by the final jar, is that correct? if so, I am fine with it

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/RoomCallMethodVisitor.kt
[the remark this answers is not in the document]
marandaneto: yeah could be, thanks

plugin-build/buildSrc/src/main/kotlin/io/sentry/android/gradle/internal/ASMifyTask.kt
romtsn: Yeah, I think that's the point of `buildSrc` folder, to have assisting tasks and plugins in there, that are added to the classpath of the project, but are not being published (https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources)

marandaneto: yep, I just missed this when writing the comment, but 2nd pass made more sense, thanks, resolving the comment.

plugin-build/build.gradle.kts
romtsn: figured in this thread https://github.com/getsentry/sentry-android-gradle-plugin/pull/190#discussion_r721616166

PR #194, opened 2021-10-04
feat: More @Deletion test cases

romtsn:
## :scroll: Description

* Adds more @Delete query test cases that room can generate
* Unignores `instrumentation/*.class` files, as I figured they were not under the version control
* Actually fixes a bug in the visitor after adding new test cases

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/RoomCallMethodVisitor.kt
```diff
@@ -62,19 +62,23 @@ class RoomCallMethodVisitor(
         when {
             (opcode == Opcodes.INVOKEVIRTUAL && name == SET_TRANSACTION_SUCCESSFUL) -> {
                 // the original method wants to return, but we intervene here to set status
+                instrumenting.set(true)
```
marandaneto, line 65: would make sense to put this before and after `when` instead of each case?

marandaneto (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/RoomCallMethodVisitor.kt
romtsn: Will be gone in the follow-up PR, sorry for the mess

PR #189, opened 2021-09-27
feat: Generify instrumentables

romtsn:
## :scroll: Description

This changes approach of `Instrumentable` interface, so we can easier instrument room-generated classes. The main change is that Class and Method instrumentables are split and receive `ClassContext` and `MethodContext` respectively, to decide whether they need to instrument something or not.  ClassContext is coming from AGP.

## :pencil: Checklist

- [ ] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [ ] No breaking changes

PR #185, opened 2021-09-09
feat: SQLiteStatement instrumentation (insert)

romtsn:
## :scroll: Description

* Adds instrumentation for `FrameworkSQLiteStatement` class
* Adds new `EditActivity` screen for adding/editing tracks in the sample
* Makes visitor test a paramterized one

image

image

## :bulb: Motivation and Context

To instrument insert/update/delete queries as well

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [ ] I updated the docs if needed
- [ ] No breaking changes

PR #176, opened 2021-08-31
feat: Query method instrumentation

romtsn:
#skip-changelog

## :scroll: Description

Adds `androidx.sqlite.db.FrameworkSqliteDatabase.query()` method visitor that wraps the query logic with [sentry spans](https://docs.sentry.io/platforms/android/performance/instrumentation/custom-instrumentation/#add-more-spans-to-the-transaction). Note: we rely that `sentry-android` SDK is already included into the target app.

Before instrumentation:
image

After instrumentation:
image

Sentry performance dashboard:
image

## :bulb: Motivation and Context

https://github.com/getsentry/sentry-java/issues/1571

## :green_heart: How did you test it?
Manually

## :pencil: Checklist

- [x] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [ ] No breaking changes

PR #175, opened 2021-08-31
feat: Room sample

romtsn:
#skip-changelog

## :scroll: Description

Adds Room sample to test and demonstrate Room instrumentation via bytecode manipulation

## :bulb: Motivation and Context

https://github.com/getsentry/sentry-java/issues/1571

## :green_heart: How did you test it?
Manually for now

## :pencil: Checklist

- [x] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [x] No breaking changes

PR #196, opened 2021-10-06
feat/tracingInstrumentation opt

marandaneto:
- feat: transform api agp (#174)
- feat: Room sample (#175)
- feat: Query method instrumentation (#176)
- feat: Instrument query method overload for API 16 and above (#182)
- feat: Instrumentation logging (#183)
- feat: ExecSQL method instrumentation (#184)
- feat: SQLiteStatement instrumentation (insert) (#185)
- ktlint
- ktlint
- Implement automatic var counting (#188)
- feat: Generify instrumentables (#189)
- feat: Instrument Room generated classes (transactions) (#190)
- feat: More @Deletion test cases (#194)
- Fix ktlint
- ktlint and drop agp 4.1.x
- drop agp 4.x version
- fix broken tests and ktlint
- add agp 7.1 alpha to CI matrix
- refactor bootstrap sdk
- add feat/bytecode-instrumentation to CI branches
- refactor and extract isVariantAllowed
- Add tracingInstrumentation option

## :pencil: Checklist

- [ ] No breaking changes


romtsn (approved):
LGTM, just a few more options I thought of:
* Have a list of ignored classes/packages/modules from instrumentation?
* Not sure if the `ignoreVariants`&co should be separated for instrumentation and the rest of plugin's job? E.g. I want to instrument debug builds, but don't want the mappings getting uploaded
* We could also have a list of all available instrumenters, that can be tweaked as necessary (like androidx.sqlite, androidx.room, clicks in the future and so on), but that's probably an overkill

PR #203, opened 2021-10-13
Gradle enhacements

marandaneto:
_#skip-changelog_


cortinico:
Quick question: does this mean we're dropping support for AGP 4.x? If so should be added in the changelog at least for this branch?

PR #204, opened 2021-10-13
Add support for M1 Macs

### LINKED ISSUE #192: Support new sentry-cli architectures
https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/download-sentry-cli.sh#L5

https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt#L99-L108

eg M1 macs require sentry-cli-Darwin-universal instead of Darwin-x86_64
check some other architectures too

cortinico:
## :scroll: Description
This PR adds support for the `Darwin-universal` CLI binary and for M1 Macs.

Fixes #192

Added tests. I don't have a M1 device so I can't really test it locally.

## :pencil: Checklist
- [x] No breaking changes

## :crystal_ball: Next steps

I looked into adding also the other arch but I don't think it's worth at the moment. We will enlarge the plugin size so much that we will reach ~100Mb and I believe is not great forcing users to download all the arch especially if they end up using only one.

### DIFF
--- a/plugin-build/download-sentry-cli.sh
+++ b/plugin-build/download-sentry-cli.sh
@@ -2,7 +2,7 @@
 cd $(dirname "$0")
 REPO=getsentry/sentry-cli
 VERSION=1.69.1
-PLATFORMS="Darwin-universal Darwin-x86_64 Linux-i686 Linux-x86_64 Windows-i686"
+PLATFORMS="Darwin-universal Linux-i686 Linux-x86_64 Windows-i686"
 
 rm -f src/main/resources/bin/sentry-cli-*
 for plat in $PLATFORMS; do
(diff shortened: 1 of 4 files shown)


marandaneto:
> I looked into adding also the other arch but I don't think it's worth at the moment. We will enlarge the plugin size so much that we will reach ~100Mb and I believe is not great forcing users to download all the arch especially if they end up using only one.

@cortinico we can add on-demand or refactor in the future that it downloads the correct binary when executing the plugin (downside is you must be online for the 1st execution), so it becomes a very thin jar, but let's just add the M1 support for now

PR #209, opened 2021-10-15
Add autoUploadNativeSymbols to SentryUploadNativeSymbolsTask

### LINKED ISSUE #208: Integrate sentry-cli dry run mode for `upload-dif` (upload debug symbols)
since sentry-cli upload-dif now supports a dry run mode via https://github.com/getsentry/sentry-cli/issues/986 we can remove the `SENTRY_AUTH_TOKEN` env or `if: env.SENTRY_AUTH_TOKEN != null` workaround from the GH actions, but rather integrating the dry run mode into the `SentryUploadNativeSymbolsTask` class, also requires upgrading sentry-cli

cortinico:
## :scroll: Description
Adding handling of `autoUploadNativeSymbols` in the native symbols upload (similarly to is done for the mapping upload).

Fixes #208

Tests are attached.

## :pencil: Checklist
- [x] No breaking changes

## :crystal_ball: Next steps
Will have to wait for a CLI release to bump it.


cortinico:
Just a small heads up: I had to reintroduce the `if: env.SENTRY_AUTH_TOKEN != null`. The reason behind it is that the `test-publish` is executed also when PRs are coming from external forks (like this one) and should be skipped in that case.

marandaneto:
> Just a small heads up: I had to reintroduce the `if: env.SENTRY_AUTH_TOKEN != null`. The reason behind it is that the `test-publish` is executed also when PRs are coming from external forks (like this one) and should be skipped in that case.

that's still fine, another work around would be to enable or not the flag `autoUpload` and `autoUploadDebugSymbols` whether PR comes from a fork or not, but I guess we tried that before and didn't work out so well

PR #225, opened 2021-11-06
Make the ignoreXXX properties in SentryPluginExtension SetProperty

ansman:
## :scroll: Description
The changed properties are better suited as sets values should be unique.

This is technically a breaking change but a major release is coming up and I doubt it will cause any issues since it's mostly source compatible.

Set properties are more semantically correct since the properties are used to filter objects.

I ran the tests, no changes.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-* Fix: Correctly add the proguard UUID output directory to the source set (#226) 
+* Fix: Correctly add the proguard UUID output directory to the source set (#226)
+* Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 
 ## 3.0.0-beta.1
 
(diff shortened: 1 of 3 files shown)


cortinico:
> This is technically a breaking change but a major release is coming up and I doubt it will cause any issues since it's mostly source compatible.

Agree. I'd leave this to @marandaneto and @romtsn as they're driving the 3.x release. Potentially this can be added as well over there. I'm unsure about the timeline

cortinico (approved):

marandaneto:
the changes do make sense but since 3.0.0 already reached beta (avoiding breaking changes that are not urgent), I'd rather do it on v4

marandaneto:
blocked till https://github.com/getsentry/sentry-android-gradle-plugin/pull/227 gets merged as well.
@ansman would you mind adding a changelog entry? we'd need to follow up with migration docs as well + updating the user-facing docs.
https://docs.sentry.io/platforms/android/proguard/#variant-filtering and https://docs.sentry.io/platforms/android/migration/

ansman:
The docs does not need to be updated. You can still set the property with a list since it has an overload for `void set(@Nullable Iterable<? extends T> elements);`

Migration would only be needed if anyone is relying on the property returning a `List`, I can add that

marandaneto:
sounds good, then have a look at `KtLint`, changelog entry and pull from `main`, CI should be happy after that.
thanks for doing that btw @ansman

marandaneto (approved):

Issue #221, opened 2021-10-22
Add test fixtures to verify classes post-instrumentation


romtsn:
We could also create test fixtures to verify classes after they've been instrumented, e.g. use Textifier from ASM to see the bytecode representation of the target class with spans, and then compare them as plain text.

romtsn:
prior art: https://cs.android.com/android/platform/superproject/main/+/main:external/dagger2/java/dagger/hilt/android/plugin/main/src/test/kotlin/TransformTest.kt

PR #232, opened 2021-11-10
Change description of Room queries to Dao classname

romtsn:
## :scroll: Description

Change the Room queries description to fully-qualified class name of the Dao

image

Ran existing tests

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Fix: Correctly add the proguard UUID output directory to the source set (#226)
 * Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 * Expose SentryPluginExtension.tracingInstrumentation (#229)
-* Ref: Change Room queries description to Dao class name (#231)
+* Ref: Change Room queries description to Dao class name (#232)
 
 ## 3.0.0-beta.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fix: Correctly add the proguard UUID output directory to the source set (#226)
 * Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 * Expose SentryPluginExtension.tracingInstrumentation (#229)
+* Ref: Change Room queries description to Dao class name (#231)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fix: Correctly add the proguard UUID output directory to the source set (#226)
 * Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 * Expose SentryPluginExtension.tracingInstrumentation (#229)
+* Ref: Change Room queries description to Dao class name (#231)
```
romtsn, line 8:
```suggestion
* Ref: Change Room queries description to Dao class name (#232)
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/AbstractRoomVisitor.kt
```diff
@@ -0,0 +1,42 @@
+package io.sentry.android.gradle.instrumentation.androidx.room.visitor
+
+import io.sentry.android.gradle.instrumentation.AbstractSpanAddingMethodVisitor
+import org.objectweb.asm.Label
+import org.objectweb.asm.MethodVisitor
+
+abstract class AbstractRoomVisitor(
```
marandaneto, line 7: this is only DRY right? same code as before + `className`

romtsn: yep

marandaneto:
is it possible to unit or integration test it? since no tests were broken its already kinda of a red flag

romtsn:
We'd need to implement https://github.com/getsentry/sentry-android-gradle-plugin/issues/221 first. I know it's not ideal now, but the fact that JVM bytecode verifier accepts the instrumented classes is already a good sign :) and I kinda hand-tested it as you can see on the screenshot. I will prioritize that issue after the file i/o is done

marandaneto (approved):
LGTM

PR #233, opened 2021-11-12
fix: Log broken bytecode in case it breaks

romtsn:
## :scroll: Description

Previously we wouldn't log the bytecode if there was an exception, now we do.
Also, display a meaningful error to the user with a link to our issue tracker in case it breaks

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 * Expose SentryPluginExtension.tracingInstrumentation (#229)
 * Ref: Change Room queries description to Dao class name (#232)
-- Fix: Log broken bytecode when build fails (#233)
+* Fix: Log broken bytecode when build fails (#233)
 
 ## 3.0.0-beta.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 * Expose SentryPluginExtension.tracingInstrumentation (#229)
 * Ref: Change Room queries description to Dao class name (#232)
+- Fix: Log broken bytecode when build fails (#233)
 
 ## 3.0.0-beta.1
 
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
@@ -4,16 +4,17 @@ import com.android.build.api.instrumentation.AsmClassVisitorFactory
 import com.android.build.api.instrumentation.ClassContext
 import com.android.build.api.instrumentation.ClassData
 import com.android.build.api.instrumentation.InstrumentationParameters
+import io.sentry.android.gradle.SentryPlugin
 import io.sentry.android.gradle.instrumentation.androidx.room.AndroidXRoomDao
 import io.sentry.android.gradle.instrumentation.androidx.sqlite.database.AndroidXSQLiteDatabase
 import io.sentry.android.gradle.instrumentation.androidx.sqlite.statement.AndroidXSQLiteStatement
+import io.sentry.android.gradle.util.warn
 import java.io.File
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 import org.objectweb.asm.ClassVisitor
-import org.slf4j.LoggerFactory
 
 @Suppress("UnstableApiUsage")
 abstract class SpanAddingClassVisitorFactory :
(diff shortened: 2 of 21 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
```diff
@@ -61,12 +58,12 @@ abstract class SpanAddingClassVisitorFactory :
                 parameters = parameters.get()
             )
             ?: nextClassVisitor.also {
-                logger.warn(
+                SentryPlugin.logger.warn {
```
marandaneto, line 61: isn't possible to get via ctor right? not a big fun of statics

marandaneto:
can we add tests to the conditional parts if possible?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Feature: Make the ignoreXXX properties in SentryPluginExtension sets (#225)
 * Expose SentryPluginExtension.tracingInstrumentation (#229)
 * Ref: Change Room queries description to Dao class name (#232)
+- Fix: Log broken bytecode when build fails (#233)
```
cortinico, line 9:
```suggestion
* Fix: Log broken bytecode when build fails (#233)
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
romtsn: yeah it's not :( I can make it a top-level property, but not sure if it's any better. Also seen this current pattern in other plugins like Firebase-perf, so it's probably fine

romtsn:
> can we add tests to the conditional parts if possible?

I will try, I think it's possible 👍

marandaneto (approved):
LGTM

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/CatchingMethodVisitor.kt
```diff
@@ -0,0 +1,36 @@
+package io.sentry.android.gradle.instrumentation.util
+
+import io.sentry.android.gradle.SentryPlugin
+import io.sentry.android.gradle.instrumentation.MethodContext
+import io.sentry.android.gradle.util.error
+import org.objectweb.asm.MethodVisitor
+import org.slf4j.Logger
+
+interface ExceptionHandler {
+    fun handle(exception: Exception)
```
marandaneto, line 10:
```suggestion
    fun handle(exception: Throwable)
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/CatchingMethodVisitor.kt
```diff
@@ -0,0 +1,36 @@
...
+    private val className: String,
+    private val methodContext: MethodContext,
+    private val exceptionHandler: ExceptionHandler? = null,
+    private val logger: Logger = SentryPlugin.logger
+) : MethodVisitor(apiVersion, prevVisitor) {
+
+    override fun visitMaxs(maxStack: Int, maxLocals: Int) {
+        try {
+            super.visitMaxs(maxStack, maxLocals)
+        } catch (e: Exception) {
```
marandaneto, line 25:
```suggestion
        } catch (e: Throwable) {
```

also check in the other catch blocks, always `Throwable` instead of `Exception`

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/CommonClassVisitorTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        var debug = false
+
+        val sut
+            get() = CommonClassVisitor(
+                Opcodes.ASM9,
+                ParentClassVisitor(),
+                "SomeClass",
+                listOf(TestInstrumentable()),
+                TestSpanAddingParameters(debugOutput = debug, inMemoryDir = tmpDir)
+            )
```
marandaneto, line 29: this could be a method that takes `debug` as default param=false, so you avoid:
`fixture.debug = true` every time, also, your fixture would be stateless which is what we want.

ideally, there's a single instance of `fixture` instead of a new instance of fixture on `setUp`, this is only necessary if we have states with statics or something.

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/util/CatchingMethodVisitorTest.kt
```diff
@@ -0,0 +1,81 @@
...
+}
+
+class CapturingExceptionHandler : ExceptionHandler {
+
+    var capturedException: Exception? = null
+
+    override fun handle(exception: Exception) {
+        capturedException = exception
+    }
+}
```
marandaneto, line 81: be sure that we are able to catch and log Throwables as well

PR #240, opened 2021-12-02
Introduce the includeProguardMapping option and deprecate autoUpload

### LINKED ISSUE #238: Add `includeProguardMapping` property to completely enable or disable the feature
Today we only have `autoUpload` which means sentry-cli is run in dry-mode, so the file isn't uploaded, but the `SentryGenerateProguardUuidTask` task is still executed and the file `sentry-debug-meta.properties` is still generated.

The idea is that this feature works in conjunction with `autoUpload`, which means:

if `includeProguardMapping` is disabled, the task `SentryGenerateProguardUuidTask` bails out since the user does not want to do anything with proguard mapping file, same for SentryUploadProguardMappingsTask.

if the `includeProguardMapping` is enabled, but `autoUpload` disabled, it works as it is today, it generates the `sentry-debug-meta.properties` with the generated id, and the user can later upload manually via sentry-cli

the name can be different, but I guess `includeProguardMapping` makes more sense since the plugin does some other stuff too.
we could technically deprecate `autoUpload` and rename it to `autoUploadProguardMapping`, but fallback to `autoUpload` because of compatibility.

cortinico:
## :scroll: Description
This PR essentially includes two changes:
1. Add the `includeProguardMapping` to exclude all the Proguard related code (defaulted to true).
2. Deprecate `autoUpload` in favor of `autoUploadProguardMapping`.

If this lands the documentation should be updated as well.

* Fixes #238 

Unit tests are attached.

## :pencil: Checklist
- [x] No breaking changes

To be precise there is a breaking change in the Task API but IIRC that's not a major deal as we don't expect users to use the task directly.

### DIFF
--- a/examples/android-ndk/build.gradle
+++ b/examples/android-ndk/build.gradle
@@ -35,7 +35,8 @@ android {
 if (System.getenv("AUTO_UPLOAD")) {
     apply plugin: 'io.sentry.android.gradle'
     sentry {
-        autoUpload = false
+        includeProguardMapping = true
+        autoUploadProguardMapping = false
         uploadNativeSymbols = true
         includeNativeSources = true
         autoUploadNativeSymbols = CI.INSTANCE.canAutoUpload()
--- a/examples/android-ndk/build.gradle
+++ b/examples/android-ndk/build.gradle
@@ -35,7 +35,6 @@ android {
 if (System.getenv("AUTO_UPLOAD")) {
     apply plugin: 'io.sentry.android.gradle'
     sentry {
-        includeProguardMapping = true
         autoUploadProguardMapping = false
         uploadNativeSymbols = true
         includeNativeSources = true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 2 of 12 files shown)


examples/android-ndk/build.gradle
```diff
@@ -35,7 +35,8 @@ android {
 if (System.getenv("AUTO_UPLOAD")) {
     apply plugin: 'io.sentry.android.gradle'
     sentry {
-        autoUpload = false
+        includeProguardMapping = true
```
marandaneto, line 38:
```suggestion
```

since its enabled by default

marandaneto:
> To be precise there is a breaking change in the Task API but IIRC that's not a major deal as we don't expect users to use the task directly.

@cortinico agreed

marandaneto (approved):
thanks for doing this @cortinico 
LGTM from my side, let's see if @romtsn has opinions.

romtsn (approved):
LGTM

marandaneto:
we'll need to update https://docs.sentry.io/platforms/android/proguard/ and https://docs.sentry.io/platforms/android/migration/#migrating-from-iosentrysentry-android-gradle-plugin-2x-to-iosentrysentry-android-gradle-plugin-300 once this goes GA :)

PR #239, opened 2021-11-25
feat: FileInputStream/FileOutputStream instrumentation

romtsn:
_#skip-changelog_

## :scroll: Description

* Introduce new screen in the sample where we utilise File I/O apis
* Introduce new remapping

image

## :green_heart: How did you test it?
Manually for now - automated tests will follow in future PRs

## :pencil: Checklist

- [x] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [x] No breaking changes

## :crystal_ball: Next steps
The PR is targeting integration branch for now, the next steps are:
* Runtime instrumentation lib with the actual instrumented classes
* More classes instrumentation (FileReader/FileWriter, context.openFileInput/Output)
* Tests

PR #241, opened 2021-12-16
feat: Add FileReader/FileWriter instrumentation

romtsn:
_#skip-changelog_

## :scroll: Description

* Added FileReader/FileWriter instrumentation
* Modified the sample to pick a File I/O strategy from the UI to be able to test it

Supporting PR on the java SDK - https://github.com/getsentry/sentry-java/pull/1843

## :pencil: Checklist

- [x] No breaking changes


examples/android-room/build.gradle.kts
```diff
@@ -73,6 +73,6 @@ sentry {
 
     tracingInstrumentation {
         forceInstrumentDependencies.set(true)
-        debug.set(true)
+//        debug.set(true)
```
marandaneto, line 76: was it intentional?

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/remap/RemappingInstrumentable.kt
```diff
@@ -0,0 +1,35 @@
...
+        apiVersion: Int,
+        originalVisitor: ClassVisitor,
+        parameters: SpanAddingClassVisitorFactory.SpanAddingParameters
+    ): ClassVisitor =
+        ClassRemapper(originalVisitor, SimpleRemapper(mapping))
+
+    override fun isInstrumentable(data: ClassContext): Boolean {
+        return when {
+            data.currentClassData.className.startsWith("io.sentry") &&
+                !data.currentClassData.className.startsWith("io.sentry.android.roomsample") -> false
```
marandaneto, line 31: also consider `io.sentry.samples.` and `io.sentry.mobile.`, similar to `SentryStackTraceFactory` on Java SDK

marandaneto: `RemappingInstrumentable` should also be fixed

marandaneto: Similar to https://github.com/getsentry/sentry-android-gradle-plugin/pull/239#discussion_r757341503 which was resolved but not worked on.
This is not only for samples actually since the Sentry Mobile app has a package called `io.sentry.mobile...` and it'd not work at all.

examples/android-room/build.gradle.kts
romtsn: yep, will remove

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/remap/RemappingInstrumentable.kt
romtsn: yea, but we don't use SAGP in those, so why bother?

romtsn: 👍 sounds good then

PR #242, opened 2021-12-22
ci: Add Default Platform Label to PRs and Issues

philipphofmann:
#skip-changelog

### DIFF
--- a/.github/workflows/add-platform-label.yml
+++ b/.github/workflows/add-platform-label.yml
@@ -1,0 +1,17 @@
+name: Add Platform Label
+on:
+  issues:
+    types: [opened]
+  pull_request:
+    types: [opened]
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: andymckay/labeler@e6c4322d0397f3240f0e7e30a33b5c5df2d39e90
+        with:
+          add-labels: "Platform: Android"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


romtsn (approved):

.github/workflows/add-platform-label.yml
```diff
@@ -0,0 +1,17 @@
...
+  pull_request:
+    types: [opened]
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: andymckay/labeler@e6c4322d0397f3240f0e7e30a33b5c5df2d39e90
```
cortinico, line 14: Sorry for being late here. What's the reason for using this version of `labeller` rather than the official one? https://github.com/actions/labeler

romtsn: cc @philipphofmann

philipphofmann: We want to be able to filter issues across repositories in https://github.com/orgs/getsentry/projects/19/views/7. As GitHub projects is still in beta, we can't filter by repositories and labels. Therefore we use this workflow.

cortinico: > We want to be able to filter issues across repositories in [github.com/orgs/getsentry/projects/19/views/7](https://github.com/orgs/getsentry/projects/19/views/7). As GitHub projects is still in beta, we can't filter by repositories and labels. Therefore we use this workflow.

I think we should be able to migrate to:
- actions/labeller for labelling all the PRs
- updating the default issue template to apply the `Platform: Android` on all the issues.

I would suggest to don't rely on third party actions + pinning on a commit sha as the approach is somehow brittle and adds unnecessary dependencies.

bruno-garcia: Thanks @cortinico, TIL. `actions/labeler` sounds like the way to go indeed

philipphofmann: If you choose a blank issue, you don't get the label, though, but worth the tradeoff instead of using a third-party action.

philipphofmann: Here you go https://github.com/getsentry/sentry-android-gradle-plugin/pull/248 and https://github.com/getsentry/sentry-android-gradle-plugin/pull/247

cortinico: > If you choose a blank issue

nit: We could set `blank_issues_enabled` to false to prevent them: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

philipphofmann: Thanks a lot @cortinico for the input.

PR #243, opened 2021-12-24
feat: Add sentry-android sdk version check

romtsn:
_#skip-changelog_

## :scroll: Description

Add a sentry-android SDK version check in the projects.afterevaluate step

This is necessary to avoid breaking customers' code at runtime, for instance, if they are using an older version of `sentry-android` which does not contain file I/O runtime classes. So we just disable the auto-instrumentation features based on the main SDK version

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -79,6 +79,9 @@ class SentryPlugin : Plugin<Project> {
                         params.debug.setDisallowChanges(
                             extension.tracingInstrumentation.debug.get()
                         )
+                        params.features.setDisallowChanges(
+                            extension.tracingInstrumentation.features.get()
+                        )
                         params.sdkStateFile.set(
                             project.file(
                                 File(project.buildDir, buildSdkStateFilePath(variant.name))
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -24,13 +24,13 @@ import io.sentry.android.gradle.util.SentryPluginUtils.capitalizeUS
 import io.sentry.android.gradle.util.SentryPluginUtils.withLogging
 import io.sentry.android.gradle.util.getSentryAndroidSdkState
 import io.sentry.android.gradle.util.info
+import java.io.File
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.plugins.ExtraPropertiesExtension
 import org.gradle.api.tasks.TaskProvider
 import org.slf4j.LoggerFactory
-import java.io.File
 
 @Suppress("UnstableApiUsage")
 class SentryPlugin : Plugin<Project> {
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -4,6 +4,7 @@ import com.android.build.api.instrumentation.FramesComputationMode
 import com.android.build.api.instrumentation.InstrumentationScope
 import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.gradle.AppExtension
+import com.android.build.gradle.internal.cxx.json.writeJsonFile
 import com.android.build.gradle.internal.utils.setDisallowChanges
 import io.sentry.android.gradle.SentryCliProvider.getSentryCliPath
 import io.sentry.android.gradle.SentryPropertiesFileProvider.getPropertiesFilePath
@@ -21,18 +22,18 @@ import io.sentry.android.gradle.tasks.SentryUploadNativeSymbolsTask
 import io.sentry.android.gradle.tasks.SentryUploadProguardMappingsTask
 import io.sentry.android.gradle.util.SentryPluginUtils.capitalizeUS
 import io.sentry.android.gradle.util.SentryPluginUtils.withLogging
-import java.io.File
+import io.sentry.android.gradle.util.getSentryAndroidSdkState
+import io.sentry.android.gradle.util.info
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.plugins.ExtraPropertiesExtension
 import org.gradle.api.tasks.TaskProvider
 import org.slf4j.LoggerFactory
+import java.io.File
 
 @Suppress("UnstableApiUsage")
 class SentryPlugin : Plugin<Project> {
-
-    private val sep = File.separator
 
     override fun apply(project: Project) {
         val extension = project.extensions.create(
@@ -78,6 +79,11 @@ class SentryPlugin : Plugin<Project> {
                         params.debug.setDisallowChanges(
                             extension.tracingInstrumentation.debug.get()
                         )
+                        params.sdkStateFile.set(
+                            project.file(
+                                File(project.buildDir, buildSdkStateFilePath(variant.name))
+                            )
+                        )
                         params.tmpDir.set(tmpDir)
                     }
                     variant.setAsmFramesComputationMode(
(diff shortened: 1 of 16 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/util/SentryAndroidSdkState.kt
```diff
@@ -0,0 +1,32 @@
...
+            Regex(
+                "((([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)"
+            )
+        /* ktlint-enable max-line-length */
+
+        fun from(semVer: String): SentryAndroidSdkState =
+            when {
+                semVer == "unspecified" -> MISSING
+                !semverRegex.matches(semVer) -> error("Unknown version $semVer of sentry-android")
+                semVer < PERFORMANCE.minVersion -> MISSING
```
bruno-garcia, line 26: It'll require a good level of attention to add new ones here. Maybe worth a comment?

romtsn: sure, you mean a comment mentioning that if we add a new instrumentation, we should also make sure this class is updated, right?

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/SentryAndroidSdkChecker.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+    logger.warn { "Unable to detect sentry-android dependency" }
+    return SentryAndroidSdkState.MISSING
+}
+
+private fun Set<ResolvedDependency>.findSentryAndroidSdk(): String? {
+    val queue = LinkedList(this)
+    while (queue.isNotEmpty()) {
+        val dep = queue.remove()
+        if (dep.moduleGroup == "io.sentry" && dep.moduleName == "sentry-android") {
```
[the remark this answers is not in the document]
romtsn: mm, we could do that, but it's always added transitively and we always say add `io.sentry:sentry-android:$version` in our docs, so maybe it's fine like that? I doubt someone is using sentry-android-core directly

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/SentryAndroidSdkChecker.kt
```diff
@@ -0,0 +1,56 @@
...
+    if (version != null) {
+        return try {
+            val sdkState = SentryAndroidSdkState.from(version)
+            logger.info {
+                "Detected sentry-android $sdkState for version: $version, " +
+                    "variant: $variantName, config: $configurationName"
+            }
+            sdkState
+        } catch (e: IllegalStateException) {
+            logger.warn { e.localizedMessage }
```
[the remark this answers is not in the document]
romtsn: We are getting this when there's some garbage instead of a semver as version, e.g. `io.sentry:sentry-android:test` (which is probably not gonna happen anyway, because gradle would fail the sync), so I thought warning is a good middle-ground here, because error would mean we fail the build (which we don't), and info is not displayed by default in the output

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/SentryAndroidSdkChecker.kt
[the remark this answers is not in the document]
romtsn: Changed in 422a02c, since I've figured yday in discord someone wanted to use sentry-android-core directly, so it's more precise, thanks!

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -113,9 +122,7 @@ class SentryPlugin : Plugin<Project> {
                         getPackageBundleTask(project, variant.name)
                     }
                 } else {
-                    project.logger.info(
-                        "[sentry] isMinifyEnabled is false for variant ${variant.name}."
-                    )
+                    project.logger.info { "isMinifyEnabled is false for variant ${variant.name}." }
```
cortinico, line 125: Why you removed `[sentry]` from the logging message? The tasks still have it.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -210,7 +217,22 @@ class SentryPlugin : Plugin<Project> {
                     // if its a bundle aab, assemble might not be executed, so we hook into bundle task
                     bundleTask?.configure { it.finalizedBy(uploadSentryNativeSymbolsTask) }
                 } else {
-                    project.logger.info("[sentry] uploadSentryNativeSymbols won't be executed")
+                    project.logger.info { "uploadSentryNativeSymbols won't be executed" }
```
cortinico, line 220: Same as above.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -210,7 +217,22 @@ class SentryPlugin : Plugin<Project> {
...
+                }
+
+                project.afterEvaluate {
```
cortinico, line 223: Is this necessary? If yes, please add a comment on why.

cortinico (commented):
Sorry I came a bit late to the review :) Anyway I left some comments.

I'm also curious to know why is it needed to resolve a configuration to search for `"sentry-android-core"`? Is there a scenario where the Gradle Plugin can be applied to a project that doesn't also import the Android SDK? If so, why is this the case?

It would be way less error prone to just offer the possibility to add an `implementation` dependency to the Sentry Android SDK. Maybe this can be an opt-out feature, so the user could manually configure it if they wish.

This will simplify also the `SentryAndroidSdkState` and the `isAtLeast` as you already know the version of the Android SDK that is shipped with the Gradle Plugin.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
romtsn: It's our custom [extension function](https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/main/kotlin/io/sentry/android/gradle/util/SentryLogging.kt) which automatically prepends all logs with `[sentry]`. I will update the SentryPluginTest in a separate PR then, thanks for pointing

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
romtsn: yep, the dependency configuration can only be resolved after the configuration/evaluation phase is done. Ideally this would've been a task, but I had a hard time finding out which task I should depend on - seems like AGP starts its transforms without binding to a specific task based on my tests (and we need to know the SDK version before the transforms are executed). Will leave a comment 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cortinico: Thanks for pointing out, I forgot about it 👍

romtsn:
> Sorry I came a bit late to the review :) Anyway I left some comments.
> 
> I'm also curious to know why is it needed to resolve a configuration to search for `"sentry-android-core"`? Is there a scenario where the Gradle Plugin can be applied to a project that doesn't also import the Android SDK? If so, why is this the case?
>
> It would be way less error prone to just offer the possibility to add an `implementation` dependency to the Sentry Android SDK. Maybe this can be an opt-out feature, so the user could manually configure it if they wish.
> 
> This will simplify also the `SentryAndroidSdkState` and the `isAtLeast` as you already know the version of the Android SDK that is shipped with the Gradle Plugin.

Yeah, that's unlikely that SAGP would be used without the main Android SDK, but we are not solely checking for presence of the SDK here, but also the version. Based on the version we enable/disable bytecode manipulation features, which otherwise would crash the app at runtime, if we would be calling to some non-existing classes.

We thought of having SAGP applying the main SDK as a dependency, but that still cannot guarantee that the version won't be resolved by Gradle to a higher one (if there's one), or that it can even be forced by the users to something else. This approach may be brittle, but seems the most failsafe one at least. We can't afford this on the SDK I believe as it will be crashing the end-users apps.

This can be a nice feature though, I think we will eventually end up having SAGP as a single integration point

romtsn:
Submitted #250 to address your review @cortinico

cortinico:
Thanks for the clarifications @romtsn 👍

PR #245, opened 2022-01-03
feat: Auto-Instrumentation features configuration

romtsn:
_#skip-changelog_

## :scroll: Description

* Adds new `features` configuration option under the `tracingInstrumentation` sub-extension. This allows specifying a set of features enabled for bytecode manipulation, currently only contains `FILE_IO` and `DATABASE` respectively.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
@@ -54,10 +54,10 @@ enum class InstrumentationFeature {
     DATABASE,
 
     /**
-     * When enabled the SDK will replace [java.io.FileInputStream], [java.io.FileOutputStream],
-     * [java.io.FileReader], [java.io.FileWriter] with Sentry-specific implementations of those.
-     * It starts and finishes a Span within a File I/O operation performed in a target app.
-     * This feature uses bytecode manipulation.
+     * When enabled the SDK will create spans for [java.io.FileInputStream],
+     * [java.io.FileOutputStream], [java.io.FileReader], [java.io.FileWriter].
+     * This feature uses bytecode manipulation and replaces the above
+     * mentioned classes with Sentry-specific implementations.
      */
     FILE_IO
 }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
@@ -3,12 +3,12 @@ package io.sentry.android.gradle
 import javax.inject.Inject
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
+import org.gradle.api.provider.SetProperty
 
 open class TracingInstrumentationExtension @Inject constructor(objects: ObjectFactory) {
     /**
      * Enable the tracing instrumentation.
-     * Does bytecode manipulation for 'androidx.sqlite' and 'androidx.room' libraries.
-     * It starts and finishes a Span within any CRUD operation performed by sqlite or room.
+     * Does bytecode manipulation for specified [features].
      * Defaults to true.
      */
     val enabled: Property<Boolean> = objects.property(Boolean::class.java)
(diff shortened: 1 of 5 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
```diff
@@ -31,4 +31,32 @@ open class TracingInstrumentationExtension @Inject constructor(objects: ObjectFa
...
+                InstrumentationFeature.DATABASE,
+                InstrumentationFeature.FILE_IO
+            )
+        )
+}
+
+enum class InstrumentationFeature {
+    /**
+     * Does bytecode manipulation for 'androidx.sqlite' and 'androidx.room' libraries.
+     * It starts and finishes a Span within any CRUD operation performed by sqlite or room.
```
philipphofmann, line 52: `m`: I think we should focus on what this feature does rather than implementation details. The same applies to the comment for `FILE_IO`. Maybe something like:
```suggestion
     * When enabled the SDK will create spans for any CRUD operation performed by 'androidx.sqlite'
     * and 'androidx.room'. This feature uses bytecode manipulation.
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
```diff
@@ -69,17 +78,38 @@ abstract class SpanAddingClassVisitorFactory :
...
+        parameters: SpanAddingParameters
+    ): Boolean =
+        sdkState.isAtLeast(PERFORMANCE) &&
+            parameters.features.get().contains(InstrumentationFeature.DATABASE)
+
+    private fun isFileIOInstrEnabled(
+        sdkState: SentryAndroidSdkState,
+        parameters: SpanAddingParameters
+    ): Boolean =
+        sdkState.isAtLeast(FILE_IO) &&
```
philipphofmann, line 110: `m`: Why isn't this `PERFORMANCE`, but `FILE_IO`?

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginTest.kt
```diff
@@ -186,6 +186,36 @@ class SentryPluginTest(
         assertTrue(":app:transformReleaseClassesWithAsm" in build.output)
     }
 
+    @Test
+    fun `applies only DB instrumentables when only DATABASE feature enabled`() {
```
philipphofmann, line 190: `m`: Should we also add a similar test for `when only FILE_IO enabled`?

philipphofmann (commented):
Great job. A few suggestions.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
romtsn: bcoz we are relying on some runtime classes which are available only from a certain version (FILE_IO=5.5.0 in this case), otherwise we would crash at runtime

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginTest.kt
romtsn: 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
romtsn: That looks better, thanks

plugin-build/src/main/kotlin/io/sentry/android/gradle/TracingInstrumentationExtension.kt
```diff
@@ -31,4 +31,33 @@ open class TracingInstrumentationExtension @Inject constructor(objects: ObjectFa
...
+     * When enabled the SDK will create spans for any CRUD operation performed by 'androidx.sqlite'
+     * and 'androidx.room'. This feature uses bytecode manipulation.
+     */
+    DATABASE,
+
+    /**
+     * When enabled the SDK will replace [java.io.FileInputStream], [java.io.FileOutputStream],
+     * [java.io.FileReader], [java.io.FileWriter] with Sentry-specific implementations of those.
+     * It starts and finishes a Span within a File I/O operation performed in a target app.
+     * This feature uses bytecode manipulation.
```
philipphofmann, line 60: `l`: Maybe better?
```suggestion
     * When enabled the SDK will create spans for [java.io.FileInputStream],
     * [java.io.FileOutputStream], [java.io.FileReader], [java.io.FileWriter].
     * This feature uses bytecode manipulation and replaces the above
     * mentioned classes with Sentry-specific implementations.
```

philipphofmann (approved):
LGTM

cortinico (approved):

PR #257, opened 2022-01-13
feat: Improve gradle-build-action caching

romtsn:
_#skip-changelog_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,12 @@ jobs:
 
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-includes: |
+            caches
+            notifications
+            jdks
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
 
       - name: Set up Java
         uses: actions/setup-java@v2
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -28,6 +28,11 @@ jobs:
       - name: Run Gradle tasks
         uses: gradle/gradle-build-action@v2
         with:
+          gradle-home-cache-includes: |
+            caches
+            notifications
+            jdks
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
           arguments: preMerge --continue
 
       - name: Upload Test Results
@@ -40,6 +45,11 @@ jobs:
       - name: Build the Debug variant
         uses: gradle/gradle-build-action@v2
         with:
+          gradle-home-cache-includes: |
+            caches
+            notifications
+            jdks
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
           arguments: assembleDebug
 
       - name: Build the Release variant
--- a/.github/workflows/test-publish-dry-run.yaml
+++ b/.github/workflows/test-publish-dry-run.yaml
@@ -47,6 +47,7 @@ jobs:
       - name: Build the Release variant
         uses: gradle/gradle-build-action@v2
         with:
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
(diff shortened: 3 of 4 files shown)


plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/VisitorTest.kt
```diff
@@ -82,11 +82,14 @@ class VisitorTest(
 
     @After
     fun printLogs() {
-        tmpDir.root.listFiles()
-            ?.filter { it.name.contains("instrumentation") }
-            ?.forEach {
-                print(it.readText())
-            }
+        // only print bytecode when running locally
+        if (System.getenv("CI")?.toBoolean() != true) {
```
marandaneto, line 86: If it's just a test, why is that filtered? if it fails on CI, we cannot see the output

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/fakes/TestSpanAddingParameters.kt
```diff
@@ -42,29 +28,10 @@ class TestSpanAddingParameters(
             .convention(setOf(InstrumentationFeature.FILE_IO, InstrumentationFeature.DATABASE))
 
     override val sdkStateFile: RegularFileProperty
-        get() = filePropertyFactory.newFileProperty().value(filePropertyFactory.file(_sdkStateFile))
+        get() = TODO()
```
marandaneto, line 31: What's the relation of this change with this PR related to gradle-build caching, and Why is it a TODO now?

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/VisitorTest.kt
romtsn: yeah but if it doesn't fail, it becomes too noisy, so you're gonna have a hard time looking for the real problem if something else fails. And if this fails, you anyway would checkout and run locally to fix the problem, I don't think one could fix bytecode issues by just looking at the CI output :D 

As an alternative, we could use upload-action to upload output of the bytecode visitors as a zip archive

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/fakes/TestSpanAddingParameters.kt
romtsn: True, the PR could be named better. I initially opened it to make the CI green, and this was one of the reasons it was failing (we were using some internal Gradle APIs that were trying to create a temp folder). After checking I realized that this property is not used in tests at all, so I just dropped a TODO to make it compile. It could also be a `mock()` if you prefer.

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/VisitorTest.kt
romtsn: filed https://github.com/getsentry/sentry-android-gradle-plugin/issues/259

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/fakes/TestSpanAddingParameters.kt
marandaneto: Gotcha, a `mock` would make more sense since we know there's nothing to be done, TODO gives us the impression we've to do something about it, all good then, it was just to understand the change, happy to keep as it is as well, your call.

romtsn: will change later when I touch this again 👍

PR #258, opened 2022-01-14
ci: Run tests of iosched with Sentry

philipphofmann:
Checkout https://github.com/google/iosched, apply Sentry to the project, and run the tests to validate
the sentry-android-gradle-plugin isn't introducing any major bugs to a real-world Android app.

We don't know how stable the tests of iosched are. When tests are failing it could either be that the tests
are flaky or that our plugin introduced a bug. Therefore we run the tests both with Sentry enabled
and not enabled. The plan is to let them run in CI for a while to get some data on this.

The tests set `integration-tests` as Sentry environment so we can filter the transactions. 

#skip-changelog


scripts/add-sentry-to-iosched.patch
```diff
@@ -0,0 +1,98 @@
...
+ subprojects {
++    apply(plugin= "io.sentry.android.gradle")
++
++    sentry {
++        tracingInstrumentation {
++            enabled.set(true)
++
++            features.set(setOf(InstrumentationFeature.DATABASE, InstrumentationFeature.FILE_IO))
++        }
++    }
```
marandaneto, line 39: This isn't needed I guess since it's the default values cc @romtsn

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+  # Sentry to the app with auto performance monitoring enabled. We then run the UI tests to make sure
+  # adding our SDK doesn't cause any major issues.
+  build-iosched:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: 'google/iosched'
+          ref: '6f96c1528c00d3d89bedb33e5a0a10d67629e8bd'
```
cortinico, line 23: Why this specific ref?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - uses: actions/checkout@v2
+        with:
+          path: 'sentry-android-gradle-plugin'
+
+      - name: Add Sentry
+        run: git apply ./sentry-android-gradle-plugin/scripts/add-sentry-to-iosched.patch
+
+      - name: Copy CI gradle.properties
+        run: mkdir -p ~/.gradle ; cp .github/ci-gradle.properties ~/.gradle/gradle.properties
```
cortinico, line 33:
```suggestion
        run: mkdir -p ~/.gradle && cp .github/ci-gradle.properties ~/.gradle/gradle.properties
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          java-version: 11
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/modules-*
+            ~/.gradle/caches/jars-*
+            ~/.gradle/caches/build-cache-*
+          key: iosched-gradle-${{ hashFiles('checksum.txt') }}
```
cortinico, line 46: Should you use the same Gradle setup @romtsn introduced?

cortinico (approved):
Nice stuff :)

.github/workflows/integration-tests.yml
philipphofmann: If google/iosched gets a flaky test on the master branch, I don't want this to affect our CI. We have to manually update the ref from time to time.

.github/workflows/integration-tests.yml
philipphofmann: I just copied this from https://github.com/google/iosched/blob/6f96c1528c00d3d89bedb33e5a0a10d67629e8bd/.github/workflows/iosched.yaml#L30-L36. I'm going to add a comment to avoid confusion in the future.

.github/workflows/integration-tests.yml
philipphofmann: Added a comment.

romtsn:
@philipphofmann I updated the PR with comments and also making sure that iosched is using the includedBuild for the gradle plugin. 

The CI was failing because it seems jcenter.bintray.com is down again, not sure how do we go about this? Either we need to try to update the dependencies and get rid of jcenter() repo, or we pick another OSS app, as iosched is looking stale for a while.

romtsn (approved):

philipphofmann:
Thanks a lot, @romtsn 🥳 .

Issue #60, opened 2021-03-10
Support DexGuard


marandaneto:
The Sentry Gradle Plugin currently supports ProGuard and R8 but fails to find DexGuard Gradle's tasks.

The tasks have a different prefix name, we'd need to look up for them too as we do for ProGuard and R8

See: https://github.com/getsentry/sentry-android-gradle-plugin/blob/master/src/main/groovy/io/sentry/android/gradle/SentryPlugin.groovy#L139-L142

https://www.guardsquare.com/en/products/proguard/manual/gradleplugin

marandaneto:
@cortinico do you have any experience with DexGuard?
we probably just need to find out its task names and add here https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt#L16-L28
everything else should work ootb.

cortinico:
> @cortinico do you have any experience with DexGuard?

Sadly not, but I see two ways to approach this:
* Find the task by name. Easier but a bit error prone
* Depend on the DexGuard Gradle plugin as `compileOnly` an get the `TaskProvider` instance from them

marandaneto:
> > @cortinico do you have any experience with DexGuard?
> 
> Sadly not, but I see two ways to approach this:
> 
> * Find the task by name. Easier but a bit error prone
> * Depend on the DexGuard Gradle plugin as `compileOnly` an get the `TaskProvider` instance from them

the 2nd is the ideal one, if possible, we only use the `minify` tasks by name because they don't have a `TaskProvider` or not accessible at least for Plugin developers iirc

cortinico:
Just a heads up, I started working on this one. The implementation is not yet completed but I hope to be able to share a PR in 1/2 weeks 👍

romtsn:
@cortinico thanks a bunch, do you have access to dexguard to test it out though?

cortinico:
> @cortinico thanks a bunch, do you have access to dexguard to test it out though?

Nope, I've been testing with the standalone Proguard setup suggested here https://www.guardsquare.com/manual/setup/gradleplugin

romtsn:
@cortinico we've been in the talks with guardsquare and I think I can get you a developer license for dexguard - I will contact them today.

Also, I think we should probably just follow the current setup with finding a task by name, wdyt?

Last note, you could also take a look at how bugsnag does it, I think it has some good clues:
[...]

PR #263, opened 2022-01-25
Add Support for external Proguard

cortinico:
## :scroll: Description

This PR adds support for GuardSquare's Proguard (NOT DexGuard) to the Gradle Plugin.

I'm unsure we want to add this support, as I supposed most of the users will either use R8, AGP's Proguard or DexGuard.

This PR can however be used as a blueprint for the setup of the DexGuard Support (See #60).
I would like however to discuss this PR as there are a couple of blockers for the DexGuard implementation, specifically.

Here Proguard is applied as a `compileOnly` dependency (that won't be possible for DexGuard). This is needed as we need to access a `ProGuardAndroidExtension` to check if the user configured Proguard for a variant or not.

For DexGuard instead, I believe that we can't use a `compileOnly` dependency (or can we?).

While this is not a problem for finding the task by name, as we can do a lookup for `"transformClassesAndResourcesWithProguardTransformFor${variantName.capitalized}"`, the same can't be said for the `isMinifyEnabled` property of a variant.

How is this going to be handled with DexGuard? Potentially only by assuming that the only variant where Proguard will be enabled is "release" (i.e. assuming that the mapping is inside 
`outputs/proguard/release/mapping/mapping.txt`). That will break the workflow for users that are having say a "canary" flavor or whatever other flavor which is not "release". Also, users that are looking into minifying their debug variant will be affected as well.

Looking for some ideas on what would be the best approach here:
1. Do we want to merge this?
2. How are we going to tackle the DexGuard setup?

Tested locally with the added example. Proguard is running on terminal and mapping is correclty updated:

<details>
  <summary>gw :examples:android-guardsquare-proguard:assembleRelease</summary>

```
gw :examples:android-guardsquare-proguard:assembleRelease

> Configure project :examples:android-gradle
Unable to detect AGP versions for included builds. All projects in the build should use the same AGP version. Class name for the included build object: org.gradle.composite.internal.DefaultIncludedBuild$IncludedBuildImpl_Decorated.

> Task :examples:android-guardsquare-proguard:transformClassesAndResourcesWithProguardTransformForRelease
ProGuard, version 7.1.0
Reading input...
Reading program jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/release/R.jar] (filtered)
Reading program directory [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/javac/release/classes] (filtered)
Reading program jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/merged_java_res/release/base.jar] (filtered)
Reading library jar [/opt/android_sdk/platforms/android-30/android.jar] (filtered)
Initializing...
Ignoring unused library classes...
  Original number of library classes: 4750
  Final number of library classes:    152
Marking classes and class members to be kept...
Printing kept classes, fields, and methods...
Inlining subroutines...
Shrinking...
Printing usage to [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/outputs/proguard/release/mapping/usage.txt]...
Removing unused program classes and class elements...
  Original number of program classes:            3
  Final number of program classes:               1
Obfuscating...
Printing mapping to [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/outputs/proguard/release/mapping/mapping.txt]...
  Number of obfuscated classes:                  0
  Number of obfuscated fields:                   0
  Number of obfuscated methods:                  1
Preverifying...
Writing output...
Preparing output jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/transforms/ProguardTransform/release/3.jar] (filtered)
Preparing output jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/transforms/ProguardTransform/release/0.jar] (filtered)
  Copying resources from program jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/release/R.jar] (filtered)
Preparing output directory [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/transforms/ProguardTransform/release/1] (filtered)
  Copying resources from program directory [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/javac/release/classes] (filtered)
Preparing output jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/transforms/ProguardTransform/release/2.jar] (filtered)
  Copying resources from program jar [/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/intermediates/merged_java_res/release/base.jar] (filtered)

> Task :examples:android-guardsquare-proguard:uploadSentryProguardMappingsRelease
> compressing mappings
> uploading mappings
> Uploaded a total of 1 new mapping files
Newly uploaded debug symbols:
  88fb3ae6-7d1d-4f46-bad8-bb006241f5c3

> Task :examples:android-guardsquare-proguard:lintVitalAnalyzeRelease
Execution optimizations have been disabled for task ':examples:android-guardsquare-proguard:lintVitalAnalyzeRelease' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/generated/assets/sentry/release'. Reason: Task ':examples:android-guardsquare-proguard:lintVitalAnalyzeRelease' uses this output of task ':examples:android-guardsquare-proguard:generateSentryProguardUuidRelease' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.

> Task :examples:android-guardsquare-proguard:lintVitalRelease
Execution optimizations have been disabled for task ':examples:android-guardsquare-proguard:lintVitalRelease' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/ncor/oss/sentry-android-gradle-plugin/examples/android-guardsquare-proguard/build/generated/assets/sentry/release'. Reason: Task ':examples:android-guardsquare-proguard:lintVitalRelease' uses this output of task ':examples:android-guardsquare-proguard:generateSentryProguardUuidRelease' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.3.3/userguide/validation_problems.html#implicit_dependency for more details about this problem.
No issues found.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.3.3/userguide/command_line_interface.html#sec:command_line_warnings

Execution optimizations have been disabled for 2 invalid unit(s) of work during this build to ensure correctness.
Please consult deprecation warnings for more details.

BUILD SUCCESSFUL in 4s
55 actionable tasks: 36 executed, 19 up-to-date
```

</details>

## :pencil: Checklist
- [x] No breaking changes

## :crystal_ball: Next steps
Decide on how we want to move forward with #60


romtsn:
I'm not aware if there was ever a request to support guardsquare's proguard, so will leave this to @marandaneto, but for me it's fine to merge this in, since the changes for supporting it are rather minor.

As for dexguard, I think that's the reason why bugsnag implemented the logic in [groovy](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/3f08a0804366e2169b84abe15acbeff30f53e503/src/main/groovy/com/bugsnag/android/gradle/GroovyCompat.groovy), to work-around Kotlin's strong typing. 
We could also follow that approach, or we could try to use compileOnly but this can be problematic as we can't checkout `dexguard.jar` into version control, so it has to be somehow shared via a CDN/artifactory between developers and also CI to make the plugin compile-able (also, I'm not sure if it will work when the license expires, but I think it should). The former approach sounds like less headache

cortinico:
> or we could try to use compileOnly but this can be problematic as we can't checkout `dexguard.jar` into version control

Are we sure about it? My assumption was that we can't commit the license information but the `.jar` is not a secret (as it's potentially obfuscated). Could you cross check with GuardRail?

romtsn:
Yeah, I'm sure, otherwise you'd be able to find it on the internet, but you won't find it (at least I didn't) ;) And we also checked with them, it's not possible.

marandaneto:
Adding support to GuardSquare's Proguard is definitely a good addition.

We need to be sure that if the final App does not have `proguard-gradle` in the classpath, which is most of the cases, the Plugin should still work and do not crash.

We can check the jar file within the repo and still use it as `compileOnly`, it should not crash when it's not available, same as the previous item.
Final users will add the file to the classpath anyway, right? it's part of the DexGuard setup.

cortinico:
> We can check the jar file within the repo and still use it as `compileOnly`

> And we also checked with them, it's not possible.

@marandaneto Seems like we can't then?

romtsn:
I guess he meant to perform a "runtime" (well, runtime for the plugin, which is compile-time lol) check and try to locate the dexguard.jar both in the plugin-build dir and in the target project dir. Only add `compileOnly` if we were able to locate the dexguard.jar, and same have an if-check in the plugin logic to avoid dexguard logic if we didn't find the jar in the target project dir.

I guess it's similar to what we do with `BootstrapAndroidSdk`.

marandaneto:
> I guess he meant to perform a "runtime" (well, runtime of the plugin, which is compile-time lol) check and try to locate the dexguard.jar both in the plugin-build dir and in the target project dir. Only add `compileOnly` if we were able to locate the dexguard.jar, and, in addition, have an if-check in the plugin logic to avoid dexguard logic if we didn't find the jar in the target project dir.
> 
> I guess it's similar to what we do with `BootstrapAndroidSdk`.

Exactly, you nailed it, thanks.

cortinico:
> Only add `compileOnly` if we were able to locate the dexguard.jar

I'm struggling to understand this point. How is the plugin supposed to compile if the `.jar` is not found, hence the `compileOnly` dependency is not added?

romtsn:
oh, indeed you're right, it won't compile. We could use something like git patch to apply, in case dexguard was found, but this looks too sophisticated already. I'd still vote for writing that logic in groovy to avoid static typing and make it compilable even if the dependency is not available.

Another option I just noticed, they have a private maven repo at guardsquare, we could also try to use that. Only two things I'm not sure about:
1. if I can share credentials with @cortinico @marandaneto?
2. Whether it will work when our license expires.
image

marandaneto:
The credentials part should be discussed with @bruno-garcia.
If the file is private and we can't rely on the private maven repo which depends on the credentials too, maybe writing in Groovy isn't a bad idea.
GuardSquare's Proguard can keep as it is since it uses the `compileOnly` bits.

cortinico:
> maybe writing in Groovy isn't a bad idea.

I'm starting to think about it as well. Even though we'll end up having two compilation involved for just a file which is unfortunate

gimp87:
cortinico:nc/gs-proguard

cortinico:
What are we going to do with this PR then?
I also I see several CI Jobs failing with 401 as the token is not passed. Wasn't that fixed for PRs coming from forks?

marandaneto:
> What are we going to do with this PR then? I also I see several CI Jobs failing with 401 as the token is not passed. Wasn't that fixed for PRs coming from forks?

I'll discuss that today with @bruno-garcia and @romtsn 
I'll be away for 2 weeks but @romtsn should be able to guide you thru this PR.

Thanks for your contribution @cortinico.

examples/android-guardsquare-proguard/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+   configurations {
+      release {
+         defaultConfiguration "proguard-android-optimize.txt"
+         configuration "proguard-project.txt"
+      }
+   }
+}
+
+sentry {
+    autoUploadProguardMapping = true
```
romtsn, line 44:
```suggestion
    autoUploadProguardMapping = CI.INSTANCE.canAutoUpload()
```
this should fix the CI dry-run workflow

cortinico: oops 😅 I forgot to revert that. Thanks for doing it.

cortinico:
Windows tests are failing with:

```
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 305056 bytes for Chunk::new
```

Have you folks seen this before?

romtsn:
Hm, I thought I fixed it through https://github.com/getsentry/sentry-android-gradle-plugin/blob/60c4a82d66c251bff099c9790632ec7fd93cf424/.github/workflows/pre-merge.yaml#L19, the build wasn't failing anymore for me, but looks like no :D

romtsn (approved):

marandaneto:
> Windows tests are failing with:
> 
> ```
> #
> # There is insufficient memory for the Java Runtime Environment to continue.
> # Native memory allocation (malloc) failed to allocate 305056 bytes for Chunk::new
> ```
> 
> Have you folks seen this before?

Let's see if https://github.com/getsentry/sentry-android-gradle-plugin/pull/273 helps, it should fix the issues.

romtsn:
@cortinico JFYI, I will try to nail the problem on my windows machine, hopefully can reproduce it and fix it. Otherwise I will just ignore some tests on windows, will update you tomorrow so we can finalize your PRs

romtsn:
thanks @cortinico a lot! Would you have time to update the dexguard PR or should I take care of it?

cortinico:
> thanks @cortinico a lot! Would you have time to update the dexguard PR or should I take care of it?

It should be now good to go

peterdk:
Was this ever merged? I cannot get Sentry Gradle to upload my proguard mappings (that uses external proguard plugin instead of R8). It just won't generate the tasks, so I assume this is due to not detecting that I use proguard.

romtsn:
Yes, it was! And we also have a sample project for that https://github.com/getsentry/sentry-android-gradle-plugin/tree/main/examples/android-guardsquare-proguard, but I'd suggest you post how you set it up, and you `sentry` plugin config, etc etc

peterdk:
Ah good to know, thanks for the fast response! From a cursory look at it, I did have the same setup. But for now I just use the CLI, with versioncode/name and no UUID. Seems to work. (Crosses fingers).

Problem was that the Sentry Gradle plugin took 1 minute for each Gradle task with configuring itself (in combination with the Chinese selfhosted sentry with much latency ofcourse and 8 build flavors). So it did not have much use. If the matching fails somehow, I'll try again with the Gradle based uploading.

peterdk:
Hmm, I did have the sentry plugin after the proguard one. Maybe that is important?

PR #266, opened 2022-01-31
AGP bumps and build improvements

romtsn:
_#skip-changelog_

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -11,7 +11,7 @@ object BuildPluginsVersion {
 
 object LibsVersion {
     const val JUNIT = "4.13.2"
-    const val ASM = "9.2"
+    const val ASM = "7.0"
     const val SQLITE = "2.1.0"
     const val SENTRY = "5.5.0"
 }
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -11,7 +11,7 @@ object BuildPluginsVersion {
 
 object LibsVersion {
     const val JUNIT = "4.13.2"
-    const val ASM = "7.0"
+    const val ASM = "7.0" // compatibility matrix -> https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/instrumentation/InstrumentationContext#apiversion
     const val SQLITE = "2.1.0"
     const val SENTRY = "5.5.0"
 }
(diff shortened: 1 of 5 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
romtsn, line 3: was told by the gradle team that it's better to use -bin rather than -all

cortinico: Why this? IIRC the `-bin` doesn't come with sources. Therefore, you will miss on some features (i.e. cmd+b inside the `.gradle` files and so on).

marandaneto: I recall this being an issue as well.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -11,7 +11,7 @@ object BuildPluginsVersion {
 
 object LibsVersion {
     const val JUNIT = "4.13.2"
-    const val ASM = "9.2"
+    const val ASM = "7.0" // compatibility matrix -> https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/instrumentation/InstrumentationContext#apiversion
```
marandaneto, line 14: If we're using AGP 7.1.x, it should be ASM9, no?

marandaneto: If this is a problem, we can add the ASM version to the CI matrix, and read it as `VERSION_AGP`

romtsn: so, we actually get the ASM version from the AGP, when the plugin is running on the clients projects (hence, the compileOnly change for ASM). So no matter what we use here, it will be overridden by the gradle's dependency resolution to what AGP is using.

I just defined a min. supported version here (as the min support AGP version is 7.0.0 -> ASM 7.0). So it's similar to our kotlin compatibility approach.

Basically, our sample apps would be built against what ASM version AGP is using underneath, so the current CI matrix that we have for AGP versions is sufficient.

gradle/wrapper/gradle-wrapper.properties
romtsn: ah, right, I guess they meant this probably for the apps that usually don't do much gradle logic/plugins, but for us it'd make sense to have sources indeed.

buildSrc/src/main/java/Dependencies.kt
marandaneto: gotcha, makes sense.

cortinico (approved):

marandaneto (approved):

gradle/wrapper/gradle-wrapper.properties
cortinico: Yup 👍  thanks for clarifying

PR #267, opened 2022-02-01
Add Support for Dexguard

### LINKED ISSUE #60: Support DexGuard

cortinico:
## :scroll: Description
This PR adds support for Dexguard (sits on top of #263 - so please consider only https://github.com/getsentry/sentry-android-gradle-plugin/commit/d297c1460b4a9b7ef4883c62a41c9301a38988d1)

There are a couple of things that are worth discussing:
- [x] I had to add a `project.afterEvaluate` as I could not query Dexguard tasks as I believe they're not properly using Gradle Configuration Avoidance.
- [x] The Dexguard Tasks are declaring dependencies on `package*` tasks. Therefore I had to change the `package*` tasks to depend on the `generateUuidTask`. I believe this should be fine.
- [x] `getMappingFileProvider` won't work correctly when building an AAB with DexGuard.

Before I spend more time on this, I'd like to get some input and feedback.

Fixes #60 

Due to the nature of this change, I've just tested it locally.

## :pencil: Checklist
- [x] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -8,6 +8,7 @@ import com.android.build.gradle.internal.cxx.json.writeJsonFile
 import com.android.build.gradle.internal.utils.setDisallowChanges
 import io.sentry.android.gradle.SentryCliProvider.getSentryCliPath
 import io.sentry.android.gradle.SentryPropertiesFileProvider.getPropertiesFilePath
+import io.sentry.android.gradle.SentryTasksProvider.capitalized
 import io.sentry.android.gradle.SentryTasksProvider.getAssembleTaskProvider
 import io.sentry.android.gradle.SentryTasksProvider.getBundleTask
 import io.sentry.android.gradle.SentryTasksProvider.getMappingFileProvider
@@ -23,6 +24,7 @@ import io.sentry.android.gradle.tasks.SentryUploadProguardMappingsTask
 import io.sentry.android.gradle.transforms.MetaInfStripTransform
 import io.sentry.android.gradle.transforms.MetaInfStripTransform.Companion.metaInfStripped
 import io.sentry.android.gradle.util.AgpVersions
+import io.sentry.android.gradle.util.GroovyCompat
 import io.sentry.android.gradle.util.SentryPluginUtils.capitalizeUS
 import io.sentry.android.gradle.util.SentryPluginUtils.isMinificationEnabled
 import io.sentry.android.gradle.util.SentryPluginUtils.withLogging
@@ -173,23 +175,35 @@ class SentryPlugin : Plugin<Project> {
                         generateUuidTask.flatMap { it.outputDirectory }
                     )
 
-                    // we just hack ourselves into the proguard task's doLast.
-                    transformerTaskProvider?.configure {
-                        it.finalizedBy(uploadSentryProguardMappingsTask)
+                    if (GroovyCompat.isDexguardEnabledForVariant(project, variant.name)) {
+                        // If Dexguard is enabled, we will have to wait for the project to be evaluated
+                        // to be able to let the uploadSentryProguardMappings run after them.
+                        project.afterEvaluate {
+                            project.tasks.named(
+                                "dexguardApk${variant.name.capitalized}"
+                            ).configure { it.finalizedBy(uploadSentryProguardMappingsTask) }
+                            project.tasks.named(
+                                "dexguardAab${variant.name.capitalized}"
+                            ).configure { it.finalizedBy(uploadSentryProguardMappingsTask) }
+                        }
+                    } else {
+                        // we just hack ourselves into the Proguard/R8 task's doLast.
+                        transformerTaskProvider?.configure {
+                            it.finalizedBy(uploadSentryProguardMappingsTask)
+                        }
                     }
 
                     // To include proguard uuid file into aab, run before bundle task.
                     preBundleTaskProvider?.configure { task ->
-                        task.dependsOn(uploadSentryProguardMappingsTask)
+                        task.dependsOn(generateUuidTask)
                     }
-
-                    // The package task will only be executed if the uploadSentryProguardMappingsTask has already been executed.
+                    // The package task will only be executed if the generateUuidTask has already been executed.
                     getPackageProvider(variant)?.configure { task ->
-                        task.dependsOn(uploadSentryProguardMappingsTask)
+                        task.dependsOn(generateUuidTask)
                     }
                     // App bundle has different package task
                     packageBundleTaskProvider?.configure { task ->
-                        task.dependsOn(uploadSentryProguardMappingsTask)
+                        task.dependsOn(generateUuidTask)
                     }
                 }
 
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
@@ -100,7 +100,7 @@ internal object SentryTasksProvider {
                 )
             )
         } else {
-            // For DexGuard the mapping file can either be inside the /apk or the /aab folder
+            // For DexGuard the mapping file can either be inside the /apk or the /bundle folder
             // (depends on the task that generated it).
             val mappingDir = "outputs${sep}dexguard${sep}mapping$sep"
             project.files(
@@ -110,7 +110,7 @@ internal object SentryTasksProvider {
                 ),
                 File(
                     project.buildDir,
-                    "${mappingDir}aab${sep}${variant.name}${sep}mapping.txt"
+                    "${mappingDir}bundle${sep}${variant.name}${sep}mapping.txt"
                 )
             )
         }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
@@ -88,36 +88,35 @@ internal object SentryTasksProvider {
     fun getMappingFileProvider(
         project: Project,
         variant: ApplicationVariant
-    ): Provider<FileCollection> = if (project.plugins.hasPlugin("com.guardsquare.proguard") ||
-        isDexguardAvailable(project)
-    ) {
-        val sep = File.separator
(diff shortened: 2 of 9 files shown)


marandaneto:
@cortinico thanks for doing this.

> I had to add a project.afterEvaluate as I could not query Dexguard tasks as I believe they're not properly using Gradle Configuration Avoidance.

That's fine because it only does that if `dexguard` plugin is set or the `dexguard` task is found, so configuration avoidance won't work anyway, we should only optimize to not do this if dexguard isn't set at all.

marandaneto:
> The Dexguard Tasks are declaring dependencies on package* tasks. Therefore I had to change the package* tasks to depend on the generateUuidTask. I believe this should be fine.

Should be fine since the problem was that the APK or AAB was packaged before the file with the proguard id was written into the folder, but that's exactly what `generateUuidTask` does, so if tests are happy, no problem at all.

marandaneto:
> getMappingFileProvider won't work correctly when building an AAB with DexGuard.

Would you mind elaborating on that?

cortinico:
> getMappingFileProvider won't work correctly when building an AAB with DexGuard.

I've updated the function to do the following:

```
    fun getMappingFileProvider(
        project: Project,
        variant: ApplicationVariant
    ): Provider<FileCollection> = if (project.plugins.hasPlugin("com.guardsquare.proguard") ||
        isDexguardAvailable(project)
    ) {
        val sep = File.separator
        val path = if (project.plugins.hasPlugin("com.guardsquare.proguard")) {
            "outputs${sep}proguard${sep}${variant.name}${sep}mapping${sep}mapping.txt"
        } else {
            "outputs${sep}dexguard${sep}mapping${sep}apk${sep}${variant.name}${sep}mapping.txt"
        }
        project.provider { project.files(File(project.buildDir, path)) }
    } else {
...
```

If we're on dexguard, we search for the mapping on:

```
./outputs/dexguard/mapping/apk/release/mapping.txt
```

The problem is that if the user is invoking `bundleRelease` the mapping will be placed inside

```
./outputs/dexguard/mapping/aab/release/mapping.txt
```

marandaneto:
@cortinico gotcha, that makes sense, would be possible to check if the file exists within the `apk` folder and if not, within the `aab` folder? maybe the file does not exist when `getMappingFileProvider` is called, what we could do is, the provider hast a list of `files`, and when these files are used, we traverse the list of files checking for its existence.
Just giving some ideas here, not sure how feasible they are.

cortinico:
Yup I'll look into it 👍

cortinico:
I'm not entirely sure why `iosched api 26 sentry enabled false` is green and all the other iosched jobs are not.

romtsn:
So the ones, where sentry enabled is `false` failed because they are flaky, but the ones with sentry, failed because of that:

```
> Task :plugin-build:buildSrc:build
e: /home/runner/work/sentry-android-gradle-plugin/sentry-android-gradle-plugin/sentry-android-gradle-plugin/plugin-build/build.gradle.kts:79:43: Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public val PluginDependenciesSpec.groovy: PluginDependencySpec defined in org.gradle.kotlin.dsl

> Configure project :plugin-build

FAILURE: Build failed with an exception.


Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See docs.gradle.org/7.0.2/userguide/command_line_interface.html#sec:command_line_warnings
* Where:
Build file '/home/runner/work/sentry-android-gradle-plugin/sentry-android-gradle-plugin/sentry-android-gradle-plugin/plugin-build/build.gradle.kts' line: 79

* What went wrong:
Script compilation error:

  Line 79:     classpath += files(sourceSets["main"].groovy.classesDirectory)
                                                     ^ Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
                                                         public val PluginDependenciesSpec.groovy: PluginDependencySpec defined in org.gradle.kotlin.dsl

1 error

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

We should probably fix it? I wanted to disable integration-tests job for now, but leave the build-iosched one (which is also failing because of that^)

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -170,23 +175,35 @@ class SentryPlugin : Plugin<Project> {
...
                     )
 
-                    // we just hack ourselves into the proguard task's doLast.
-                    transformerTaskProvider?.configure {
-                        it.finalizedBy(uploadSentryProguardMappingsTask)
+                    if (GroovyCompat.isDexguardEnabledForVariant(project, variant.name)) {
+                        // If Dexguard is enabled, we will have to wait for the project to be evaluated
+                        // to be able to let the uploadSentryProguardMappings run after them.
+                        project.afterEvaluate {
+                            project.tasks.named(
```
romtsn, line 182: Just thinking if we can also use `transformerTaskProvider` here instead?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryTasksProvider.kt
```diff
@@ -81,8 +85,39 @@ internal object SentryTasksProvider {
      * @return the provider if found or null otherwise
      */
     @JvmStatic
-    fun getMappingFileProvider(variant: ApplicationVariant): Provider<FileCollection> =
+    fun getMappingFileProvider(
+        project: Project,
+        variant: ApplicationVariant
+    ): Provider<FileCollection> = if (project.plugins.hasPlugin("com.guardsquare.proguard") ||
+        isDexguardAvailable(project)
+    ) {
```
romtsn, line 93:
```suggestion
    ): Provider<FileCollection> = 
        if (project.plugins.hasPlugin("com.guardsquare.proguard") || isDexguardAvailable(project)) {
```
nit: I think this is a bit more readable

romtsn:
Is there any way we can add a bit more tests for dexguard, or we're basically out of luck, since we can't share it and it will expire soon?

romtsn (approved):
LGTM, besides some minor things. Thanks a lot for making this through!

cortinico:
> since we can't share it and it will expire soon?

Essentially this. Or we would have to create some test machinery to emulate the same Gradle extension that Dexguard is using, which is anyway brittle at the end of the day.

marandaneto:
@cortinico @romtsn what If we tag both features as [experimental](https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L34-L54)? guarded by a flag on `SentryPluginExtension`? so we know that the Plugin should work as it is, once enabled (Guardscare Proguard and Dexguard are the majority by far I believe), they know its experimental, if any issues, they would open a ticket, but it won't break for current users.

marandaneto:
> @cortinico @romtsn what If we tag both features as [experimental](https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/ApiStatus.java?rgh-link-date=2022-02-03T18%3A46%3A40Z#L34-L54)? guarded by a flag on `SentryPluginExtension`? so we know that the Plugin should work as it is, once enabled (Guardscare Proguard and Dexguard are the majority by far I believe), they know its experimental, if any issues, they would open a ticket, but it won't break for current users.

@bruno-garcia ideas?

cortinico:
> @cortinico @romtsn what If we tag both features as [experimental](https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/ApiStatus.java?rgh-link-date=2022-02-03T18%3A46%3A40Z#L34-L54)?

I think that's doable. I'm not entirely sure what are the implications of exposing an API as `@Experimental` and it's usage on a `.kts` Gradle Script. I suppose there will be warnings during the build, but that shouldn't be a problem.

> (opt-in by default)?

I think that doing both `@Experimental` and `opt-in` might be overcautious. Potentially with the `opt-in` users will have to discover the API to use it, so there (in the docs?) there could be a note saying that this is experimental.

marandaneto:
> > @cortinico @romtsn what If we tag both features as [experimental](https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/ApiStatus.java?rgh-link-date=2022-02-04T00%3A41%3A51Z#L34-L54)?
> 
> I think that's doable. I'm not entirely sure what are the implications of exposing an API as `@Experimental` and it's usage on a `.kts` Gradle Script. I suppose there will be warnings during the build, but that shouldn't be a problem.
> 
> > (opt-in by default)?
> 
> I think that doing both `@Experimental` and `opt-in` might be overcautious. Potentially with the `opt-in` users will have to discover the API to use it, so there (in the docs?) there could be a note saying that this is experimental.

yeah, indeed, `@Experimental` is enough.

cortinico:
> yeah, indeed, `@Experimental` is enough.

I was actually suggesting to do don't add it as I believe it's over cautious :) We can still add it and remove it if we notice warnings as it's just an annotation at the end of the day.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
cortinico: `transformerTaskProvider` is null at this stage because `getTransformerTask` wasn't able to find the relevant task. Adding `"dexguardApk${variant.name.capitalized}"` and `"dexguardAab${variant.name.capitalized}"` to `getTransformerTask` also doesn't work as:
1. it requires the `project.afterEvaluate {}` block
2. For Dexguard, `transformerTaskProvider` should return two tasks (`dexguardApkDebug` and `dexguardAabDebug`).

cortinico:
It seems like the Gradle build is crashing on linux now 🤔 

> Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

romtsn:
hmm, I thought I fixed it by disabling daemon for plugin-build project... maybe the gradle props need some tweaks again

cortinico:
Are you taking a look at them?

romtsn:
Thanks @cortinico once again for your great work!

PR #270, opened 2022-02-01
fix: Do not rely on visitFrame calls 

### LINKED ISSUE #228: Cannot run r8 after adding Sentry Gradle Plugin
_Android Gradle Plugin:_
**7.1.0-beta02**

_Gradle:_
**7.2**

_Optimizer:_
- [ ]  ProGuard
- [X]  R8

- [X]  Yes -> where? `/sentry.properties`
- [ ]  No

---
I have the following issue:

We have been using our own gradle plugin for Sentry but are switching to the official one right now. After migrating we can no longer minify our app as it fails with this error:
```
Execution failed for task ':app:minifyReleaseWithR8'.
> Could not resolve all files for configuration ':app:releaseRuntimeClasspath'.
   > Failed to transform work-runtime-2.5.0.aar (androidx.work:work-runtime:2.5.0) to match attributes {artifactType=android-asm-instrumented-jars, asm-transformed-variant=internal, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.libraryelements=aar, org.gradle.status=release, org.gradle.usage=java-runtime}.
      > Execution failed for AsmClassesTransform: /Users/nicklas/.gradle/caches/transforms-3/54dca5c2bcc42347fc969b1bf4683fe8/transformed/work-runtime-2.5.0-runtime.jar.
         > java.lang.NullPointerException (no error message)
```
Simply uncommenting the adding of the Sentry plugin makes the minify task work again so it's definitely related to Sentry.

romtsn:
## :scroll: Description

* Do not rely on `visitFrame` calls during instrumentation, as they can be skipped depending on flag set to the AGP (the flag can be overridden by any plugin doing bytecode transformation)
* Adapted VisitorTest to reflect AGP behavior
* Added integration test with firebase-perf plugin

Closes #228

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix: Strip out unsupported java classes from META-INF/ (so AGP does not fail before our code is reached) (#264)
 * Bump sentry-cli 1.72.0 which prevent daemonize mode from crashing upload process (#262)
+* Fix: Incompatibilities with other Gradle plugins using the same API from AGP for bytecode instrumentation (#228)
 
 ## 3.0.0-beta.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Fix: Strip out unsupported java classes from META-INF/ (so AGP does not fail before our code is reached) (#264)
 * Bump sentry-cli 1.72.0 which prevent daemonize mode from crashing upload process (#262)
-* Fix: Incompatibilities with other Gradle plugins using the same API from AGP for bytecode instrumentation (#228)
+* Fix: Incompatibilities with other Gradle plugins using the same API from AGP for bytecode instrumentation (#270)
 
 ## 3.0.0-beta.3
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Bump sentry-cli 1.72.0 which prevent daemonize mode from crashing upload process (#262)
+* Fix: Incompatibilities with other Gradle plugins using the same API from AGP for bytecode instrumentation (#228)
```
marandaneto, line 6:
```suggestion
* Fix: Incompatibilities with other Gradle plugins using the same API from AGP for bytecode instrumentation (#270)
```

marandaneto (approved):
I cannot validate if the fix works, but if CI is happy, I am happy, codewise is also good to go, thanks for figuring it out.

romtsn:
@marandaneto yep, I added an integration test with firebase-perf plugin, if I revert this fix, the. test would fail (also, the VisitorTest would fail, after I changed the flag to `SKIP_FRAMES`)

.github/workflows/integration-tests.yml
```diff
@@ -59,62 +59,63 @@ jobs:
...
-          name: emulator-logs
-          path: emulator.log
-
-      - name: Upload test reports
-        if: always()
-        uses: actions/upload-artifact@v2
-        with:
-          name: test-reports
-          path: mobile/build/reports/
+# https://github.com/getsentry/sentry-android-gradle-plugin/issues/271 we'll switch to testing the duckduckgo app
```
marandaneto, line 62: Can we make a separate PR for that? so we merge into the main branch and merge it back to this PR.

PR #282, opened 2022-02-16
feat: Add auto-installation for sentry-android SDK and integrations

romtsn:
_#skip-changelog_ 

## :scroll: Description

* Adds auto-installation for sentry-android and timber, okhttp and fragment integrations
* We perform existing dependencies/version checks only for the direct (first-level) dependencies, because the mechanisms Gradle provides do not allow calculating the entire dependency graph at this stage. There's a way to resolve the entire graph (copying a configuration and resolving it), but this is very costly to do at configuration phase, so I'd really not wanna do that. I think this is best effort we do now, and it will cover most of the cases.
* Because of the above I've added an explicit `sentryVersion` configuration setting, so the users can specify the version, for instance, if they use an older sentry-android version that is coming as a transitive dependency from one of their submodules/libs.

This is what it looks like in practice:
image
Notice how there are no sentry dependencies in the SS above.
When we build the app, the final apk looks like this (with all integrations and the SDK included):
image

Q1 goal

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -63,6 +63,11 @@ object Samples {
 
     object OkHttp {
         private const val version = "4.9.3"
-        const val okhttp = "con.squareup.okhttp3:okhttp:${version}"
+        const val okhttp = "com.squareup.okhttp3:okhttp:${version}"
+    }
+
+    object Timber {
+        private const val version = "5.0.1"
+        const val timber = "com.jakewharton.timber:timber:${version}"
     }
 }
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -60,4 +60,9 @@ object Samples {
         const val compiler = "androidx.room:room-compiler:${version}"
         const val rxjava = "androidx.room:room-rxjava2:${version}"
     }
+
+    object OkHttp {
+        private const val version = "4.9.3"
+        const val okhttp = "con.squareup.okhttp3:okhttp:${version}"
+    }
 }
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -70,4 +70,9 @@ object Samples {
         private const val version = "5.0.1"
         const val timber = "com.jakewharton.timber:timber:${version}"
     }
+
+    object Fragment {
+        private const val version = "1.3.5"
+        const val fragmentKtx = "androidx.fragment:fragment-ktx:${version}"
+    }
 }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -292,9 +292,7 @@ class SentryPlugin : Plugin<Project> {
                     project.logger.info { "uploadSentryNativeSymbols won't be executed" }
                 }
             }
-        }
 
-        if (extension.autoInstallation.enabled.get()) {
             project.installDependencies(extension)
         }
     }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
@@ -16,6 +16,8 @@ import io.sentry.android.gradle.SentryTasksProvider.getPackageBundleTask
 import io.sentry.android.gradle.SentryTasksProvider.getPackageProvider
 import io.sentry.android.gradle.SentryTasksProvider.getPreBundleTask
 import io.sentry.android.gradle.SentryTasksProvider.getTransformerTask
+import io.sentry.android.gradle.autoinstall.installDependencies
+import io.sentry.android.gradle.extensions.SentryPluginExtension
 import io.sentry.android.gradle.instrumentation.SpanAddingClassVisitorFactory
 import io.sentry.android.gradle.services.SentrySdkStateHolder
 import io.sentry.android.gradle.tasks.SentryGenerateProguardUuidTask
@@ -291,6 +293,10 @@ class SentryPlugin : Plugin<Project> {
                 }
             }
         }
+
+        if (extension.autoInstallation.enabled.get()) {
+            project.installDependencies(extension)
+        }
     }
 
     private fun isVariantAllowed(
@@ -307,6 +313,7 @@ class SentryPlugin : Plugin<Project> {
     companion object {
         const val SENTRY_ORG_PARAMETER = "sentryOrg"
         const val SENTRY_PROJECT_PARAMETER = "sentryProject"
+        internal const val SENTRY_SDK_VERSION = "5.6.1"
 
         internal val sep = File.separator
 
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt
@@ -1,0 +1,65 @@
+package io.sentry.android.gradle.autoinstall
+
+import io.sentry.android.gradle.autoinstall.okhttp.OkHttpInstallStrategy
+import io.sentry.android.gradle.extensions.SentryPluginExtension
+import io.sentry.android.gradle.util.info
+import org.gradle.api.Project
+import org.gradle.api.artifacts.DependencySet
+
+private const val SENTRY_GROUP = "io.sentry"
+private const val SENTRY_ANDROID_ID = "sentry-android"
+private const val SENTRY_OKHTTP_ID = "sentry-android-okhttp"
+private const val SENTRY_TIMBER_ID = "sentry-android-timber"
+private const val SENTRY_FRAGMENT_ID = "sentry-android-fragment"
+
+private val strategies = listOf<InstallStrategyRegistrar>(
+    OkHttpInstallStrategy.Registrar
+)
+
+fun Project.installDependencies(extension: SentryPluginExtension) {
+    configurations.named("implementation").configure { configuration ->
+        configuration.withDependencies { dependencies ->
+            var sentryVersion = dependencies.findSentryAndroidVersion()
+            sentryVersion = installSentrySdk(sentryVersion, dependencies, extension)
+
+            val installOkHttp = !dependencies.isModuleAvailable(SENTRY_OKHTTP_ID)
+            AutoInstallState.apply {
+                this.sentryVersion = sentryVersion
+                this.installOkHttp = installOkHttp
+                this.installFragment = false
+                this.installTimber = false
+            }
+        }
+    }
+    project.dependencies.components { component ->
+        strategies.forEach { it.register(component) }
+    }
+}
+
+private fun Project.installSentrySdk(
+    sentryVersion: String?,
+    dependencies: DependencySet,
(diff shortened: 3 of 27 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/fragment/FragmentInstallStrategy.kt
```diff
@@ -0,0 +1,60 @@
...
+            return
+        }
+
+        // TODO: technically we support all versions of androidx.fragment, but there can be a case
+        // when we bump the version of androidx.fragment in the sentry-android SDK to 1.4+, which
+        // requires compileSdkVersion 31+, so the user's app might stop compiling. Should we make the
+        // transitive dependency of androidx.fragment `compileOnly` in the sentry-android SDK to avoid this?
+        // Another option: we could check the user's compileSdkVersion and do not add sentry-android-fragment
+        // (this is, when we bump the transitive androidx.fragment version there to 1.4.1, so
+        // no action item for now)
```
romtsn, line 32: @marandaneto @bruno-garcia need your opinion here

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/timber/TimberInstallStrategy.kt
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        // TODO: technically we do not support version < 4.6.0 of Timber, because they introduced
+        // nullability annotations for Kotlin, so if somebody used to extend Timber.Tree in their
+        // kotlin codebase, after adding sentry-android-timber the code won't be compile-able (source-incompatible)
+        // because we bring-in a transitive version 4.7.1. However, I think it's a fairly small
+        // price to pay, the users would need to remove a single `?` to make one parameter non-nullable and make it work.
+        // On the other hand version 4.6.0 is nearly 5 years old, so probably it's safe to drop4
+        // everything below that version.
+
```
romtsn, line 33: @marandaneto @bruno-garcia and here

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -277,6 +279,10 @@ class SentryPlugin : Plugin<Project> {
                 }
             }
         }
+
+        if (extension.autoInstallation.enabled.get()) {
```
cortinico, line 283: I'm not entirely sure, but I believe this won't work. Have you verified that it works? The reason is that the value for `autoInstallation` is not available at configuration time

romtsn: No, haven't tried yet, thanks for the hint, I will try

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/timber/TimberInstallStrategy.kt
```diff
@@ -0,0 +1,74 @@
...
+            return
+        }
+
+        context.details.allVariants { metadata ->
+            metadata.withDependencies { dependencies ->
+                val sentryVersion = autoInstallState.get().sentryVersion
+                dependencies.add("$SENTRY_GROUP:$SENTRY_TIMBER_ID:$sentryVersion")
+
+                logger.info {
+                    "$SENTRY_TIMBER_ID is successfully installed with version: $sentryVersion"
```
bruno-garcia, line 53:
```suggestion
                    "$SENTRY_TIMBER_ID was successfully installed with version: $sentryVersion"
```

bruno-garcia (commented):
Maybe I'm wrong to suggest `was` there instead of `is`. Or perhaps we should log: `installing ayz package` since I believe at that point we're just adding it to the list of dependencies.

Perhaps even: `Adding package xyz to the dependencies list`

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -307,6 +311,7 @@ class SentryPlugin : Plugin<Project> {
     companion object {
         const val SENTRY_ORG_PARAMETER = "sentryOrg"
         const val SENTRY_PROJECT_PARAMETER = "sentryProject"
+        internal const val SENTRY_SDK_VERSION = "5.6.1"
```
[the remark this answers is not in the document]
romtsn: yep, this line has to be updated to bump the SDK. I introduce a configuration option called `sentryVersion` which always can be overridden by the users, if they wanna use a version different from the one here.

Alternatively, they can also just add a direct dependency to the SDK and let Gradle resolve it to the latest version. There are many ways this can be configured :)

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+        logger.info {
+            "sentry-android won't be installed because it was already installed directly"
+        }
+        sentryVersion
+    }
+}
+
+private fun DependencySet.findSentryAndroidVersion(): String? =
+    find { it.group == SENTRY_GROUP && it.name == SENTRY_ANDROID_ID }?.version
```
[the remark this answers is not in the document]
romtsn: yep good point, I'll add it as well.

romtsn:
Had to revert back to using singleton to share AutoInstallState, cus build services cannot be serialized (and the ComponentMetadataRules are serialized). Did thorough testing - no problems occurred.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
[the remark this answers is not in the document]
bruno-garcia: I see. Worth noting that having many ways can possible introduce confusion though. Like "I'm setting version X but it's breaking" And after digging we found out version Y was being set elsewhere which took precedence so worth documenting things thoroughly.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -60,4 +60,19 @@ object Samples {
...
         const val rxjava = "androidx.room:room-rxjava2:${version}"
     }
+
+    object OkHttp {
+        private const val version = "4.9.3"
+        const val okhttp = "com.squareup.okhttp3:okhttp:${version}"
+    }
+
+    object Timber {
+        private const val version = "5.0.1"
```
marandaneto, line 70: Our lib currently uses `4.7.1`

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/fragment/FragmentInstallStrategy.kt
```diff
@@ -0,0 +1,57 @@
...
+
+// @Inject is needed to avoid Gradle error
+// @CacheableRule // TODO: make it cacheable somehow (probably depends on parameters)
+abstract class FragmentInstallStrategy @Inject constructor() : ComponentMetadataRule {
+
+    private var logger: Logger = SentryPlugin.logger
+
+    constructor(logger: Logger) : this() {
+        this.logger = logger
+    }
```
marandaneto, line 22: Do we need a logger ctor or just for testing? otherwise no need to assign `SentryPlugin.logger` directly?

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/okhttp/OkHttpInstallStrategy.kt
```diff
@@ -0,0 +1,69 @@
...
+// @CacheableRule
+abstract class OkHttpInstallStrategy @Inject constructor() : ComponentMetadataRule {
+
+    private var logger: Logger = SentryPlugin.logger
+
+    constructor(logger: Logger) : this() {
+        this.logger = logger
+    }
+
+    override fun execute(context: ComponentMetadataContext) {
```
marandaneto, line 26: Can we somehow DRY for the `execute` method? OkHttp and others look like the same, maybe we get the configuration in a generic way?

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/fragment/FragmentInstallStrategy.kt
romtsn: just for testing. Not sure wdym by no need to assign directly - how'd you see it otherwise?

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/okhttp/OkHttpInstallStrategy.kt
romtsn: Yeah I think we can, will try to come up with something

buildSrc/src/main/java/Dependencies.kt
marandaneto: not sure, if our CI compiles with a different version it means we cannot guarantee that it works with the transitive dependency, maybe a matrix then?

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/fragment/FragmentInstallStrategy.kt
marandaneto: its just a nit, you are assigning twice during tests, since it assigns already in the declaration, and you cannot make it a `val` but only `var`.
I'd rather have an optional parameter with a pre-assigned value, making it a `val`.

romtsn: mm, but this ctor is not gonna be called by Gradle, because it's Gradle who instantiates this class through `newInstance` using primary ctor. And all parameters in the ctor should be injected through `@Inject`, but we can't inject the logger unfortunately.

buildSrc/src/main/java/Dependencies.kt
romtsn: hmm, I think I'm missing the point. The transitive dependency works, because this is how it is in the sentry-java repo now, right?

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/fragment/FragmentInstallStrategy.kt
marandaneto: Yep, I know it, again it's just a nit, but See the example https://github.com/getsentry/sentry-java/blob/e1a7226bead691b2a824332f3d3a0930e68e33cc/sentry-servlet/src/main/java/io/sentry/servlet/SentryServletRequestListener.java#L21-L29

There's an empty ctor, which defaults to a value and calls the other ctor with the given value, my idea would be to do the same.
Feel free to ignore if I'm not getting the point.

buildSrc/src/main/java/Dependencies.kt
romtsn: I actually added a test that uses different versions of okhttp/timber/fragment. I think that should be enough? https://github.com/getsentry/sentry-android-gradle-plugin/blob/feat/add-dependencies/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginAutoInstallTest.kt

romtsn:
@marandaneto addressed the comments, you could take another look if you will.

marandaneto (approved):

PR #288, opened 2022-03-03
Add SentryOkHttpInterceptor instrumentation

### LINKED ISSUE #222: Automatically attach SentryInterceptor for OkHttp and Apollo
Since we have some basis for bytecode manipulation now, we can utilise that knowledge and attach SentryInterceptor for okhttp and apollo clients automatically to avoid any manual work from the user's site

romtsn:
_#skip-changelog_

## :scroll: Description

* Automatically adds `SentryOkHttpInterceptor` to OkHttp requests. We instrument the okhttp library itself, therefore covering all okhttp clients within the app. We also check for the presence of SentryOkHttpInterceptor in the interceptor list, to make sure we do not instrument multiple times

image

image

Closes #222 (apollo out of scope)

## :pencil: Checklist

- [x] No breaking changes


examples/android-instrumentation-sample/src/main/java/io/sentry/samples/instrumentation/SampleApp.kt
```diff
@@ -30,18 +24,5 @@ class SampleApp : Application() {
             .build()
 
         analytics = getSharedPreferences("analytics", Context.MODE_PRIVATE)
-
-        GlobalScope.launch(Dispatchers.IO) {
```
romtsn, line 34: Removed as it's not really needed to test file I/O

romtsn, line 34: Removed as it's not really needed to test file I/O

plugin-build/buildSrc/src/main/kotlin/io/sentry/android/gradle/internal/ASMifyTask.kt
```diff
@@ -45,7 +45,7 @@ abstract class ASMifyTask : Exec() {
     private val tmpDir: String get() = "${project.buildDir}/tmp/asmified"
 
     override fun exec() {
-        val asmJars = project.configurations.getByName("runtimeClasspath")
+        val asmJars = project.configurations.getByName("compileClasspath")
```
romtsn, line 48: This no longer worked, because we had made ASM dependencies `compileOnly`, so this had to change to compileClasspath

examples/android-instrumentation-sample/src/main/java/io/sentry/samples/instrumentation/ui/MainActivity.kt
```diff
@@ -24,14 +28,21 @@ class MainActivity : ComponentActivity() {
         list.adapter = TrackAdapter()
 
         lifecycleScope.launchWhenStarted {
+            Sentry.getSpan()?.finish()
+            val transaction = Sentry.startTransaction(
+                "Track Interaction",
+                "ui.action.load",
+                true
+            )
```
marandaneto, line 36: If the `collect` does not get called because of no items, the transaction will never be finished, or does it get called anyway?

marandaneto (approved):
LGTM

examples/android-instrumentation-sample/src/main/java/io/sentry/samples/instrumentation/ui/MainActivity.kt
romtsn: mm, I think it will always emit the latest value + we prefill the db from file, so it won't ever be empty. Should be fine I think

PR #292, opened 2022-03-10
auto assign android label

bruno-garcia:
So we can see on our GH project board filering `Platform: Android` across all repos


bruno-garcia:
Perhaps not needed because @philipphofmann added an issue template: https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/ISSUE_TEMPLATE/bug_report.md

PR #297, opened 2022-03-18
Remove cortinico from CODEOWNERS

romtsn:
@cortinico so you are not bothered by everything, but just where you're tagged/interested in :) 

_#skip-changelog_


cortinico (approved):
Yup thank you :)

PR #320, opened 2022-05-24
Bump AGP versions

romtsn:
_#skip-changelog_


plugin-build/build.gradle.kts
```diff
@@ -81,7 +81,7 @@ tasks.withType<KotlinCompile>().configureEach {
 
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
-        freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn")
+        freeCompilerArgs = listOf("-Xopt-in=kotlin.RequiresOptIn", "-Xjvm-default=enable")
```
marandaneto, line 84: Why do we need that? :)

marandaneto (approved):

plugin-build/build.gradle.kts
romtsn: Kotlin compiler [issue](https://youtrack.jetbrains.com/issue/KT-47000), we either need to use 1.6.20 where it was fixed, or use this flag

PR #321, opened 2022-06-09
Update sentry-android runtime SDK version

romtsn:


romtsn:
So I believe I'm hitting a bug in R8 (which only happens in AGP 7.3.0-betaX and 7.4.0-alphaX, since they use a newer version of R8). I've filed it here - https://issuetracker.google.com/issues/235733922. It's related to 6.0.0 proguard-rules. 

Wdyt we should do in the meantime? There's a workaround I found (needs a new release of sentry-android), or we can disable AGP beta/alpha tests for now, and force release a new version of SAGP.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Features
+* Bump sentry runtime SDK version to `6.0.0` ([#321](https://github.com/getsentry/sentry-android-gradle-plugin/pull/321))
```
marandaneto, line 6:
```suggestion
### Features

* Bump sentry runtime SDK version to `6.0.0` ([#321](https://github.com/getsentry/sentry-android-gradle-plugin/pull/321))
```

marandaneto (approved):

marandaneto:
> So I believe I'm hitting a bug in R8 (which only happens in AGP 7.3.0-betaX and 7.4.0-alphaX, since they use a newer version of R8). I've filed it here - [issuetracker.google.com/issues/235733922](https://issuetracker.google.com/issues/235733922). It's related to 6.0.0 proguard-rules.
> 
> Wdyt we should do in the meantime? There's a workaround I found (needs a new release of sentry-android), or we can disable AGP beta/alpha tests for now, and force release a new version of SAGP.

What would be the fix on sentry-android? It seems that we should fix it on our end since we can't guarantee a fast response on the AGP side.

romtsn:
Add this to proguard rules:
```
-keepclassmembers enum io.sentry.** { *; }
```

marandaneto:
> Add this to proguard rules:
> 
> ```
> -keepclassmembers enum io.sentry.** { *; }
> ```

Since there are just a few enums (~10), I'd be fine with it for now.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -324,7 +324,7 @@ class SentryPlugin : Plugin<Project> {
     companion object {
         const val SENTRY_ORG_PARAMETER = "sentryOrg"
         const val SENTRY_PROJECT_PARAMETER = "sentryProject"
-        internal const val SENTRY_SDK_VERSION = "5.7.0"
+        internal const val SENTRY_SDK_VERSION = "6.0.0"
```
romtsn, line 327:
```suggestion
        internal const val SENTRY_SDK_VERSION = "6.1.0"
```

PR #328, opened 2022-06-14
Danger PR number as markdown link 

romtsn:
_#skip-changelog_


bruno-garcia:
Could you please give a short description what this does? These danger files are generally aligned across repos so maybe something we should change elsewhere

romtsn:
Sry, this is actually to align with sentry-java 🙈 https://github.com/getsentry/sentry-java/pull/1745

Issue #309, opened 2022-04-25
MetaInfStripTransform issue with signed jars


romtsn:
The `MetaInfStripTransform` should also clean up the `MANIFEST.MF` file for transformed jars, if they are signed, because it contains references to stripped out files.

This only affects AGP versions between 7.0.0 and 7.1.2 where we run the transform.

malbanese:
If interested, here's a sample project I put together that exhibits the issue mentioned in the other thread. The error can be encountered by running any of the test gradle tasks.

https://github.com/malbanese/sentry-gradle-3-0-1-test-execution-bug

bruno-garcia:
7.0.0: Dec 1, 2020
7.1.2 (February 2022)

Basically 1 year worth of AGP. But the solution is telling users to upgrade AGP

Ideas: 
* Create a troubleshoot page entry describing this and linking to this issue
* SAGP check for the version and warns on the terminal users should upgrade AGP
* Proper fix: Clean up manifest

PR #334, opened 2022-06-27
Skip jar processing on AGP < 7.1.2 if it is a signed multi release ja…

lbloder:
## :scroll: Description
Detect signed multi release jars and let the user know that these cause problems and that they should upgrade to AGP >= 7.1.2

Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/309

Added TestCase to ensure that a signed jar is not touched by the transform.

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
Add Entry to Android Troubleshooting Guide

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransform.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransform.kt
@@ -20,6 +20,7 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
+import java.util.logging.Logger
 
 /**
  * Gradle's [TransformAction] that strips out unsupported Java classes from
@@ -63,6 +64,13 @@ abstract class MetaInfStripTransform : TransformAction<MetaInfStripTransform.Par
                         continue
                     }
 
+                    if (jarEntry.isSignatureEntry) {
+                        logger.warning("Signed Multirelease Jar (${jarFile.name}) found, skipping transform. This might lead to duplicate class errors due to a bug in AGP (https://issuetracker.google.com/issues/206655905). Please update to AGP >= 7.1.2 (https://developer.android.com/studio/releases/gradle-plugin)")
+                        output.delete()
+                        outputs.file(inputArtifact)
+                        return
+                    }
+
                     if (jarEntry.name.startsWith(versionsDir, ignoreCase = true)) {
                         val javaVersion = jarEntry.javaVersion
                         if (javaVersion > MIN_SUPPORTED_JAVA_VERSION) {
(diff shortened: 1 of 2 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransform.kt
```diff
@@ -63,6 +64,13 @@ abstract class MetaInfStripTransform : TransformAction<MetaInfStripTransform.Par
                         continue
                     }
 
+                    if (jarEntry.isSignatureEntry) {
+                        logger.warning("Signed Multirelease Jar (${jarFile.name}) found, skipping transform. This might lead to duplicate class errors due to a bug in AGP (https://issuetracker.google.com/issues/206655905). Please update to AGP >= 7.1.2 (https://developer.android.com/studio/releases/gradle-plugin)")
```
romtsn, line 68:
```suggestion
                        SentryPlugin.logger..warning { "Signed Multirelease Jar (${jarFile.name}) found, skipping transform. This might lead to auto-instrumentation issues due to a bug in AGP (https://issuetracker.google.com/issues/206655905). Please update to AGP >= 7.1.2 (https://developer.android.com/studio/releases/gradle-plugin) in order to keep using `autoInstrumentation` option." }
```

* Let's use the logger from `SentryPlugin` to not create new instances.
* Let's also use a `warning {}` extension, that would prepend `[sentry]` to each log call
* I rephrased a bit the message

plugin-build/src/test/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransformTest.kt
```diff
@@ -31,20 +31,22 @@ class MetaInfStripTransformTest {
...
+                mimicSignedJar = mimicSignedJar
             )
 
             whenever(provider.get()).thenReturn(RegularFile { jar })
 
             return TestMetaInfStripTransform(provider)
         }
 
-        private fun File.toJar(multiRelease: Boolean, includeSupportedVersion: Boolean): File {
+        private fun File.toJar(multiRelease: Boolean, includeSupportedVersion: Boolean, mimicSignedJar: Boolean): File {
```
romtsn, line 49:
```suggestion
        private fun File.toJar(multiRelease: Boolean, includeSupportedVersion: Boolean, signed: Boolean): File {
```

Let's stick to the naming pattern (e.g. we could also say `mimicMultiRelease` but we just use `multiRelease`). I think it's fine to name them like this, because it's just for tests

romtsn (approved):

lbloder:
> sorry my suggestion caused the formatting to fail 🙈
:D no worries, should be fixed now.

Any idea on the Test Publishing UUID job failing? Seems to be failing on main as well: https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/2583945350

romtsn:
Yeah, I'm certain that the problem resides in the new sentry-cli version, because from the logs it says:
```
> Found 12 debug information files
> Resolved source code for 4 debug information files
> Prepared debug information files for upload
> Nothing to upload, all files are on the server
```

I feel like there was a new feature introduced, where we check if the files are already on the server before uploading it, hence we don't get the `Uploaded X missing debug information files`

kamilogorek:
We had this feature for as long as I can remember. Not sure why it didn't duplicate those files before 👀

romtsn:
So 2.3.0 seem to have fixed it, merging in then :)

Issue #359, opened 2022-07-22
Index 0 out of bounds for length 0


wzieba:
### Gradle Version

7.4.2

### AGP Version

7.2.1

### Code Minifier/Optimizer

R8

### Version

3.0.1

### Sentry SDK Version

6.3.0

### Steps to Reproduce

1. Check out the branch available here: https://github.com/woocommerce/woocommerce-android/compare/sentry_performance_issue?expand=1
2. In the main directory prepare repository by executing: `cp gradle.properties-example gradle.properties`
3. Run build: `./gradlew assembleWasabiDebug`

### Expected Result

The build pass.

### Actual Result

The build fails with:

<details>
   <summary> ArrayIndexOutOfBoundsException </summary>

```
Unable to find common super type for java/lang/Exception and com/stripe/proto/model/rest/ErrorResponse.
Unable to find common super type for com/stripe/proto/model/rest/ErrorResponse and java/lang/Object.
Unable to find common super type for com/stripe/core/crpcserver/ServiceLogger and java/util/Iterator.
> Task :WooCommerce:desugarWasabiDebugFileDependencies UP-TO-DATE
Unable to find common super type for com/stripe/proto/api/attestation/SignedRequest and java/lang/String.
Unable to find common super type for com/stripe/proto/api/attestation/SignedRequest and java/lang/String.
Unable to find common super type for com/stripe/core/crpcserver/ServiceLogger and java/util/Iterator.
Unable to find common super type for java/lang/String and com/stripe/proto/model/common/VersionInfoPb.
Unable to find common super type for java/lang/Object and com/stripe/proto/model/common/VersionInfoPb.
Unable to find common super type for com/stripe/proto/api/rest/ReportedReaderConfig and java/lang/String.
Unable to find common super type for com/squareup/wire/Message and java/lang/Object.
Unable to find common super type for java/lang/String and com/stripe/proto/model/common/DeviceInfo.
Unable to find common super type for java/lang/Object and com/stripe/proto/model/common/DeviceInfo.
Unable to find common super type for com/stripe/proto/model/common/VersionInfoPb and com/stripe/proto/api/rest/ConnectionType.
Unable to find common super type for java/lang/String and com/stripe/proto/model/common/VersionInfoPb.
Unable to find common super type for java/lang/Object and com/stripe/proto/model/common/VersionInfoPb.
Unable to find common super type for com/stripe/proto/api/rest/ReportedReaderConfig and java/lang/String.
[...]

Issue #360, opened 2022-07-24
Failure in mergeLibDexDebug task when upgrading from 2.1.5 to 3.1.3


mshafrir-stripe:
### Gradle Version

7.5

### AGP Version

7.2.1

### Code Minifier/Optimizer

R8

### Version

3.1.3

### Sentry SDK Version

5.7.4

### Steps to Reproduce

1. Upgrade `io.sentry.android.gradle` from `2.1.5` to `3.1.3`
2. Run `./gradlew build`

### Expected Result

Complete successfully

### Actual Result

```
00:09:49.145	[sentry] Error while instrumenting r.a (Ljava/lang/String;)V.
00:09:49.145	Please report this issue at https://github.com/getsentry/sentry-android-gradle-plugin/issues
00:09:49.147	java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
00:09:49.147		at org.objectweb.asm.Frame.merge(Frame.java:1268)
00:09:49.147		at org.objectweb.asm.Frame.merge(Frame.java:1233)
00:09:49.147		at org.objectweb.asm.MethodWriter.computeAllFrames(MethodWriter.java:1610)
00:09:49.147		at org.objectweb.asm.MethodWriter.visitMaxs(MethodWriter.java:1546)
00:09:49.147		at org.objectweb.asm.MethodVisitor.visitMaxs(MethodVisitor.java:773)
00:09:49.147		at org.objectweb.asm.commons.LocalVariablesSorter.visitMaxs(LocalVariablesSorter.java:148)
00:09:49.147		at org.objectweb.asm.MethodVisitor.visitMaxs(MethodVisitor.java:773)
00:09:49.147		at io.sentry.android.gradle.instrumentation.util.CatchingMethodVisitor.visitMaxs(CatchingMethodVisitor.kt:24)
00:09:49.147		at org.objectweb.asm.tree.MethodNode.accept(MethodNode.java:769)
00:09:49.147		at org.objectweb.asm.tree.MethodNode.accept(MethodNode.java:649)
00:09:49.147		at org.objectweb.asm.tree.ClassNode.accept(ClassNode.java:452)
00:09:49.147		at io.sentry.android.gradle.instrumentation.util.AnalyzingVisitor.visitEnd(AnalyzingVisitor.kt:14)
00:09:49.147		at org.objectweb.asm.ClassVisitor.visitEnd(ClassVisitor.java:377)
[...]

PR #362, opened 2022-08-01
Detect minified classes and skip instrumentation to avoid build problems

adinauer:
## :scroll: Description

Detect minified classes using a Regex and skip visiting it to avoid `ArrayIndexOutOfBoundsException` due to empty `dstFrame.inputStack` Array in `org.objectweb.asm.Frame:1233`.

Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/359 and https://github.com/getsentry/sentry-android-gradle-plugin/issues/360

Local build and added tests including a minimal reproduction JAR

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/util/MinifiedClassDetectionTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        classNames.forEach {
+            assertFalse(classNameLooksMinified(it, "com/example/$it"), it)
+        }
+    }
+
+    @Test
+    fun `tests that something happens`() {
+        """^\w(\\${'$'}(\w))*${'$'}""".toRegex().matches("a${'$'}a")
+    }
```
marandaneto, line 56: The test name here is not clear.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginWithMinifiedLibsTest.kt
```diff
@@ -23,6 +23,7 @@ class SentryPluginWithMinifiedLibsTest :
               implementation 'com.google.android.gms:play-services-mlkit-text-recognition:18.0.0'
               implementation 'com.adcolony:sdk:4.7.1'
               implementation 'com.appboy:android-sdk-ui:19.0.0'
+              implementation files("testlib/stripe-problematic.jar")
```
marandaneto, line 26: Can we actually add a dependency directly to the stripe lib that is problematic?

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/ConstantPoolHelpers.kt
```diff
@@ -68,6 +76,25 @@ internal fun ClassReader.isMinifiedClass(): Boolean {
...
+private val MINIFIED_CLASSNAME_REGEX = """^(((([a-zA-z])\4{1,}|[a-zA-Z]{1,2})([0-9]{1,})?(([a-zA-Z])\7{1,})?)|([a-zA-Z]([0-9])?))(${'\\'}${'$'}((((\w)\14{1,}|[a-zA-Z]{1,2})([0-9]{1,})?(([a-zA-Z])\17{1,})?)|(\w([0-9])?)))*${'$'}""".toRegex()
+
+/**
+ * See https://github.com/getsentry/sentry/blob/c943de2afc785083554e7fdfb10c67d0c0de0f98/static/app/components/events/eventEntries.tsx#L57-L58
+ */
+private val MINIFIED_CLASSNAME_SENTRY_REGEX =
+    """^(([\w\${'$'}]\/[\w\${'$'}]{1,2})|([\w\${'$'}]{2}\/[\w\${'$'}]\/[\w\${'$'}]))(\/|${'$'})""".toRegex()
+/* ktlint-enable max-line-length */
+
+fun classNameLooksMinified(simpleClassName: String, fullClassName: String): Boolean {
```
marandaneto, line 93:
```suggestion
private fun classNameLooksMinified(simpleClassName: String, fullClassName: String): Boolean {
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/ConstantPoolHelpers.kt
```diff
@@ -44,17 +44,25 @@ internal fun ClassWriter.findClassReader(): ClassReader? {
...
+internal fun ClassReader.getSimpleClassName(): String {
+    return className.substringAfterLast("/")
+}
+
 /**
  * Looks at the constant pool entries and searches for R8 markers
  */
 internal fun ClassReader.isMinifiedClass(): Boolean {
-    val charBuffer = CharArray(maxStringLength)
+    return isR8Minified(this) || classNameLooksMinified(this.getSimpleClassName(), this.className)
```
marandaneto, line 55: Since there are many different ways of checking it now, I am wondering if we'd not have false positives.

marandaneto: Do we call our logger in case `isMinifiedClass` returns true? Let's do it if not, so people know why something isn't being instrumented.

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/util/MinifiedClassDetectionTest.kt
adinauer: Oops that was just there for me to figure out why the regex failed to match.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginWithMinifiedLibsTest.kt
adinauer: Yes, can update

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/ConstantPoolHelpers.kt
adinauer: Can't access it directly from test if I do that.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/ConstantPoolHelpers.kt
adinauer: Log statement is in place:
```
SentryPlugin.logger.info {
                "$simpleClassName skipped from instrumentation because it's a minified class."
            }
```

I guess it's better to skip instrumentation for too many classes rather than break the build. But yes, we'll surely have some false positives due to the regex.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/ConstantPoolHelpers.kt
marandaneto: Maybe we add the annotation that is only visible for the test then?

marandaneto:
Let's upgrade the Gradle and AGP testing matrix with the latest version to be sure that we are not breaking something new either.
You can check the latest version of AGP [here](https://maven.google.com/web/index.html#com.android.tools.build:gradle-api)
Gradle [here](https://github.com/gradle/gradle)
Also the Gradle vs AGP compatibility [here](https://developer.android.com/studio/releases/gradle-plugin#updating-gradle)
It can be a separate PR on top of this one but I'd do it before merging/releasing.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/util/ConstantPoolHelpers.kt
adinauer: I don't think they're available in SAGP (yet).

adinauer:
Created https://github.com/getsentry/sentry-android-gradle-plugin/pull/363 for the build matrix.

marandaneto (approved):

marandaneto:
Thanks for digging into the dark side of bytecode man. @adinauer .

PR #396, opened 2022-10-28
Populate events with dependencies metadata

### LINKED ISSUE #236: Send modules along with SentryEvent
There's a capability in the Sentry product to display installed modules/dependencies, for instance, this is how it looks for Ruby
image

Would be nice to do something similar for the Android SDK - we can collect the dependencies of an app through SAGP and make them available at runtime for the SDK, for example as a file under `res/raw` folder, and then the SDK can send it alongside protocol. This is already available in the [protocol](https://develop.sentry.dev/sdk/event-payloads/)

romtsn:
## :scroll: Description

* Adds a new task to collect dependencies and include them into apk/jar as part of assets/resources
* Some changes to the existing tasks to suppress some warnings and for common output handling

Closes #236 

Manually and automated

## :pencil: Checklist

- [x] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/util/SentryPluginUtils.kt
```diff
@@ -49,4 +52,10 @@ internal object SentryPluginUtils {
         }
         return variant.buildType.isMinifyEnabled
     }
+
+    fun getAndDelete(property: Provider<RegularFile>): File {
```
markushi, line 56: `l`: I'd make the method name more descriptable, how about `getAndDeleteFile()`?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryExternalDependenciesReportTask.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    @get:OutputFile
+    abstract val output: RegularFileProperty
+
+    @TaskAction
+    fun action() {
+        val outputFile = SentryPluginUtils.getAndDelete(output)
+        outputFile.parentFile.mkdirs()
+
+        val dependencies = runtimeConfiguration
```
markushi, line 62: nice one!

markushi (commented):
LGTM, please check my comments!

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -308,6 +345,71 @@ class SentryPlugin : Plugin<Project> {
...
+            }
+            if (configuredForJavaProject.getAndSet(true)) {
+                logger.info { "The Sentry Gradle plugin was already configured" }
+                return@withType
+            }
+
+            val javaExtension = project.extensions.getByType(JavaPluginExtension::class.java)
+
+            val sentryResDir = project.layout.buildDirectory.dir("generated${sep}sentry")
+            javaExtension.sourceSets.getByName("main").resources { sourceSet ->
```
adinauer, line 363: `m` Should we also allow for devs to configure the source set they want?
Maybe even support multiple and have `listOf("main")` as default?
This could be a separate issue as well.

adinauer (approved):
Only gave this a quick review. LGTM just one question regarding the hard coded `main` source set.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
romtsn: hmm, but what would be the usecase? All the sourcesets get packaged into the final jar anyway, and that's what we're interested in

markushi (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
adinauer: I was thinking about people renaming the source sets because they're following some conventions or have to support some legacy code base where names are different.

PR #409, opened 2022-11-25
Add UI transactions for compose clickables

markushi:
## :scroll: Description

Instruments the `Modifier.clickable` method to wrap the clickable lamda using `SentryClickableIntegration.wrapClickable` (which is part of `sentry-compose`).

We want to have automatic UI transactions and clickable breadcrumb events. 

## :pencil: Checklist

- [ ] No breaking changes


markushi:
We decided to determine the click/scroll targets at runtime, thus there's no need for bytecode instrumentation anymore.

Issue #313, opened 2022-05-03
Write the proguard uuid to the Android manifest


bruno-garcia:
### Description

To avoid reading the sentry.properties at runtime, given that the manifest properties are anyway always read.
Change the plugin so it writes the proguard uuid into the android manifest

markushi:
Should be done together with #308

markushi:
Let's come back to this later once we know how many people are actually using the manifest init method.

smeubank:
> Let's come back to this later once we know how many people are actually using the manifest init method.

is there a way to find that out via telemetry? We have this [dashboard](https://sentry.sentry.io/dashboard/80023/?project=4506241308295168&query=java&statsPeriod=90d) please consider adding a tag to the transactions to quantify usage, and then you can decide whether to close or complete this task

PR #421, opened 2022-12-29
initial implementation to write enabled instrumentations to file

### LINKED ISSUE #308: Inject enabled features into the sentry-android sdk
We need to find a way to inject enabled features of the gradle plugin (auto-instrumentation, auto-installation, dexguard, etc.) into the Android SDK as part of the `integrations` list (https://github.com/getsentry/sentry-java/issues/1885).

Potential solutions:
* Inject into the AndroidManifest and read it at runtime
* Write into a file (similar to proguard-uuid) and read it at runtime
* Use bytecode manipulation to update the integrations list

lbloder:
## :scroll: Description

Fixes #308 

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/examples/android-instrumentation-sample/build.gradle.kts
+++ b/examples/android-instrumentation-sample/build.gradle.kts
@@ -1,3 +1,6 @@
+import io.sentry.android.gradle.extensions.InstrumentationFeature
+import io.sentry.android.gradle.extensions.TracingInstrumentationExtension
+
 plugins {
     id("com.android.application")
     kotlin("android")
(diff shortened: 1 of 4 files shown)


lbloder:
@romtsn just came across this issue https://github.com/getsentry/sentry-android-gradle-plugin/issues/313. Wdyt, should I write the integrations to manifest rather than its own file?

romtsn:
@lbloder yes, that'd be definitely the preferred way, but it requires quite some refactoring I believe. If you wanna give it a stab, I'd look into [Bugsnag's](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/0d5873b9e50fd271ad7bbb6e6fc977034b964b95/src/main/kotlin/com/bugsnag/android/gradle/internal/ManifestUuidTaskV2Compat.kt) implementation as prior art. Otherwise you can keep it as is, until we come to this at a later point and refactor all at once (including writing the uuid to the manifest)

PR #455, opened 2023-03-23
feat: logcat integration

buenaflor:
## :scroll: Description
Adds a `logcat` subextension that has following options:
- `enabled`
- `minLevel`

`minLevel` determines which Log functions will be replaced by `SentryLogcatAdapter` via bytecode manipulation.

`SentryLogcatAdapter` will add breadcrumbs to the respective log.

Ref: https://github.com/getsentry/sentry-java/issues/1620

 - Local/manual testing
 - Jadx
 
## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
Update sentry-docs

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/LogcatExtension.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/LogcatExtension.kt
@@ -1,0 +1,21 @@
+package io.sentry.android.gradle.extensions
+
+import io.sentry.android.gradle.instrumentation.logcat.LogcatLevel
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import javax.inject.Inject
+
+open class LogcatExtension @Inject constructor(objects: ObjectFactory) {
+    /**
+     * Enables or disables the Logcat feature.
+     * Defaults to true.
+     */
+    val enabled: Property<Boolean> = objects.property(Boolean::class.java).convention(true)
+
+    /**
+     * The minimum log level to capture.
+     * Defaults to Level.WARNING.
+     */
+    val minLevel: Property<LogcatLevel> =
+        objects.property(LogcatLevel::class.java).convention(LogcatLevel.WARNING)
+}
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
@@ -77,5 +77,12 @@ enum class InstrumentationFeature {
      * This feature uses bytecode manipulation and adds an OnDestinationChangedListener to all
      * navigation controllers used in Jetpack Compose.
      */
-    COMPOSE
+    COMPOSE,
+
+    /**
+     * When enabled the SDK will create breadcrumbs when using Logcat
+     * This feature uses bytecode manipulation and replaces Logcat
+     * with Sentry-specific implementations.
+     */
+    LOGCAT
 }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
@@ -1,5 +1,6 @@
 package io.sentry.android.gradle.extensions
 
+import org.gradle.api.Action
 import javax.inject.Inject
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
@@ -44,8 +45,19 @@ open class TracingInstrumentationExtension @Inject constructor(objects: ObjectFa
                 InstrumentationFeature.FILE_IO,
                 InstrumentationFeature.OKHTTP,
                 InstrumentationFeature.COMPOSE,
+                InstrumentationFeature.LOGCAT
             )
         )
+
+    val logcat: LogcatExtension = objects.newInstance(
+        LogcatExtension::class.java
+    )
+
+    fun logcat(
+        logcatAction: Action<LogcatExtension>
+    ) {
+        logcatAction.execute(logcat)
+    }
 }
 
 enum class InstrumentationFeature {
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatClassVisitor.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatClassVisitor.kt
@@ -1,0 +1,55 @@
+package io.sentry.android.gradle.instrumentation.logcat
+
+import org.objectweb.asm.ClassVisitor
+import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.commons.AdviceAdapter
+
+class LogcatClassVisitor(
+    apiVersion: Int,
+    originalVisitor: ClassVisitor,
+    private val minLevel: LogcatLevel
+) :
+    ClassVisitor(apiVersion, originalVisitor) {
+    override fun visitMethod(
+        access: Int,
+        name: String?,
+        descriptor: String?,
+        signature: String?,
+        exceptions: Array<out String>?
+    ): MethodVisitor {
+        val mv = super.visitMethod(access, name, descriptor, signature, exceptions)
+        return LogMethodVisitor(mv, access, name, descriptor, minLevel)
+    }
+}
+
+class LogMethodVisitor(
+    mv: MethodVisitor?,
+    access: Int,
+    name: String?,
+    desc: String?,
+    private val minLevel: LogcatLevel
+) :
+    AdviceAdapter(Opcodes.ASM7, mv, access, name, desc) {
+    override fun visitMethodInsn(
+        opcode: Int,
+        owner: String,
+        name: String,
+        desc: String?,
+        itf: Boolean
+    ) {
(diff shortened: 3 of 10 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/LogcatExtension.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import javax.inject.Inject
+
+open class LogcatExtension @Inject constructor(objects: ObjectFactory) {
+    /**
+     * Enables or disables the Logcat feature.
+     * Defaults to true.
+     */
+    val enabled: Property<Boolean> = objects.property(Boolean::class.java).convention(true)
```
markushi, line 13: Since this feature causes more data to be sent, we should probably disable it by default. Maybe we can enable it by default with the next major bump.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/LogcatExtension.kt
```diff
@@ -0,0 +1,21 @@
+package io.sentry.android.gradle.extensions
+
+import io.sentry.android.gradle.instrumentation.logcat.LogcatLevel
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import javax.inject.Inject
+
+open class LogcatExtension @Inject constructor(objects: ObjectFactory) {
+    /**
+     * Enables or disables the Logcat feature.
```
markushi, line 10: Maybe you can add a comment that once enabled, this will replace logcat calls and add breadcrumbs?

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
```diff
@@ -77,5 +89,12 @@ enum class InstrumentationFeature {
...
      */
-    COMPOSE
+    COMPOSE,
+
+    /**
+     * When enabled the SDK will create breadcrumbs when using Logcat
+     * This feature uses bytecode manipulation and replaces Logcat
+     * with Sentry-specific implementations.
+     */
+    LOGCAT
```
markushi, line 99: I don't think we need this as we already have `logcat.enabled` now right?

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatLevel.kt
```diff
@@ -0,0 +1,25 @@
+package io.sentry.android.gradle.instrumentation.logcat
+
+enum class LogcatLevel {
+    VERBOSE,
+    DEBUG,
+    INFO,
+    WARNING,
+    ERROR;
+
+    fun allowedLogFunctions(): List<String> {
```
markushi, line 10: This function will probably get called many times, so it could make sense to optimize this avoid the list allocations. How about assigning a number to every level and then just comparing those? Something like: 
````
enum class LogcatLevel(val level : Int) {
    Verbose(0),
    Debug(1);

    fun supports(other: LogcatLevel) : Boolean {
        return level >= other.level
    }
}
````

File: plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatClassVisitor.kt [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+) :
+    AdviceAdapter(Opcodes.ASM7, mv, access, name, desc) {
+    override fun visitMethodInsn(
+        opcode: Int,
+        owner: String,
+        name: String,
+        desc: String?,
+        itf: Boolean
+    ) {
+        if (owner.contains("android/util/Log") && minLevel.allowedLogFunctions().contains(name)) {
```
markushi, line 41: I think it's safe to use equals here
```suggestion
        if (owner.equals("android/util/Log") && minLevel.allowedLogFunctions().contains(name)) {
```

markushi (commented):
This is great, well done! 🚀  please have a look at my comments

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
romtsn: yeah I was about to comment the same, we probably don't need it

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatInstrumentable.kt
```diff
@@ -0,0 +1,23 @@
...
+        instrumentableContext: ClassContext,
+        apiVersion: Int,
+        originalVisitor: ClassVisitor,
+        parameters: SpanAddingClassVisitorFactory.SpanAddingParameters
+    ): ClassVisitor {
+        return LogcatClassVisitor(apiVersion, originalVisitor, minLevel)
+    }
+
+    override fun isInstrumentable(data: ClassContext) =
+        !data.isSentryClass() && !data.currentClassData.className.contains("SentryLogcatAdapter\$Companion")
```
romtsn, line 22:
```suggestion
        !data.isSentryClass()
```
I think `isSentryClass()` should already cover it, because it checks if the class name starts with `io.sentry`, which `SentryLogcatAdapter` falls under. Or am I missing something?

buenaflor: It will run into a recursion because`Log.d` calls `SentryLogcatAdapter.d` which again calls `Log.d` and never break out of it but I'll check again if it's the case.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatClassVisitor.kt
```diff
@@ -0,0 +1,55 @@
...
+}
+
+class LogMethodVisitor(
+    mv: MethodVisitor?,
+    access: Int,
+    name: String?,
+    desc: String?,
+    private val minLevel: LogcatLevel
+) :
+    AdviceAdapter(Opcodes.ASM7, mv, access, name, desc) {
```
romtsn, line 33: we should also use here the same apiVersion as in the ClassVisitor instead of `Opcodes.ASM7`, so have to pass it as a parameter here

```suggestion
    AdviceAdapter(apiVersion, mv, access, name, desc) {
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatClassVisitor.kt
```diff
@@ -0,0 +1,55 @@
+package io.sentry.android.gradle.instrumentation.logcat
+
+import org.objectweb.asm.ClassVisitor
+import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.commons.AdviceAdapter
+
+class LogcatClassVisitor(
```
romtsn, line 8: we could utilise `CommonClassVisitor` instead, because it does automatic logging and catching exceptions during visiting, but up to you, this is also fine as it is now. You could look at other instrumentables for examples

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatInstrumentable.kt
romtsn: if you've tested it locally and put the adapter into the `io.sentry.samples.**` package, then indeed it'd go into a loop, because we exclude `io.sentry.samples.**` package from `isSentryClass` check

romtsn (approved):
LGTM, besides a couple of things.

We should probably also add an integration test that verifies the instrumented bytecode can be accepted by the JVM. You could look at `VisitorTest` and just add a new `.class` fixture with some logcat calls in it and then test it analogous to other fixtures in that test.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/logcat/LogcatInstrumentable.kt
buenaflor: Oh yeah, true, that might be the case, my first local tests were done adding it in the samples.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/LogcatExtension.kt
markushi: If we do the same for Timber I'm fine with keeping it enabled 🚀

markushi:
Nicely done! You can definitely mark the PR as "read for review" now.
Before merging, let's ensure we don't forget about
- [ ] running `make format` to format the code 
- [ ] extending the [changelog](https://github.com/getsentry/sentry-android-gradle-plugin/pull/455#issuecomment-1480946991)
- [ ] adding an [integration test](https://github.com/getsentry/sentry-android-gradle-plugin/pull/455#pullrequestreview-1356380488)

marandaneto:
@buenaflor is this going to work for dependencies and transitive dependencies?
the question is: is the bytecode man. able to wrap the Log calls from 3rd party libs/sdks?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
```
romtsn, line 5: btw the changelog is outdated, you'd need to merge against main again

Issue #40, opened 2021-02-08
Add app context to ProGuard mappings uploaded.


bruno-garcia:
Pass additional information to sentry-cli to augment what we know about the `mappings.txt` we're uploading.
The goal is to surface more information in the `ProGuard` page in Sentry Settings.

Data to pass to sentry-cli, app's:

* name
* id
* version

See also: https://github.com/getsentry/sentry-java/issues/954

sentry-cli already accepts some extra fields:
https://github.com/getsentry/sentry-cli/blob/master/src/commands/upload_proguard.rs#L41-L72

But we might need to extend it, and add support on the server. Needs discussion.

marandaneto:
turns out that https://github.com/getsentry/sentry-cli/blob/master/src/commands/upload_proguard.rs#L41-L72 is dead code so sentry-cli needs to support such new properties and Sentry UI should offer a way to display and filter such new properties

markushi:
@jan-auer it looks like we had a similar feature in the distant past: https://github.com/getsentry/sentry/pull/10246 Any chance you remember why it was removed?

PR #476, opened 2023-04-21
feat: add release information args to proguard mapping upload task

buenaflor:
## :scroll: Description
Release information can now be attached to the `sentry-cli` call during the proguard mapping upload task. 

`sentry-cli` already has a mechanism to add those args: `--app-id, --version, --version-code` so no work there needs to be done right now.

This is needed for associating a release to proguard mapping files. See https://github.com/getsentry/sentry-android-gradle-plugin/issues/40

manually with different flavors and unit test

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/util/ReleaseInfo.kt
```diff
@@ -0,0 +1,7 @@
+package io.sentry.android.gradle.util
+
+data class ReleaseInfo(
+    val applicationId: String,
+    val versionName: String,
+    val versionCode: Int? = null
+)
```
romtsn, line 7:
```suggestion
) : Serializable
```
`m`: I think it has to be serializable if we pass it as a task input

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -238,3 +241,18 @@ private fun AndroidComponentsExtension<*, *, *>.configureVariants(callback: (Var
         onVariants70(this, callback)
     }
 }
+
+private fun getReleaseInfo(project: Project, variant: Variant): ReleaseInfo {
+    val appExtension = project.extensions.getByType(AppExtension::class.java)
+    var applicationId =
+        appExtension.defaultConfig.applicationId ?: appExtension.namespace.toString()
+    var versionName = appExtension.defaultConfig.versionName ?: "1.0.0"
```
romtsn, line 249:
```suggestion
    var versionName = appExtension.defaultConfig.versionName ?: "undefined"
```
Maybe that is a bit more explicit than `1.0.0`? Idk if we have server-side validation of semver though, but I'd try to send undefined and see what happens

romtsn (approved):
lgtm besides some minor comments 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
buenaflor: After implementing the backend, we don't have a semver validation here so we're I guess we're free to pass in what makes sense in that case. I assume `undefined` seems reasonable if a version can't be determined.

buenaflor:
@romtsn @markushi do you know what I changed that might make the integration test fail? or is it flakiness? 

It seems like all tests are failing of versions [AGP 7.3.0, Gradle 7.5] or lower

Issue #474, opened 2023-04-20
Build fails on android-Tiramisu sdk installed


stefanosiano:
### Description

In `BootstrapAndroidSdk` we look for installed sdks in the format `android-33`. In case an sdk is installed in the format `android-Tiramisu` the build fails due to `NumberFormatException`.
We should account for the possibility to have sdks with the Android version name in the folder name instead of the sdk code, by either considering only sdk folders with a number in them or by manually mapping the names to numbers.

PR #478, opened 2023-04-26
Parse the sdk name in BootstrapAndroidSdk

stefanosiano:
## :scroll: Description
Parse the sdk name in BootstrapAndroidSdk

Some sdks like Tiramisu or UpsideDownCake are installed creating a folder with the full name instead of the version number. Our plugin check the latest sdk version, but parsing the name as integer make it crash.
Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/474

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Parse the sdk name in BootstrapAndroidSdk ([#478](https://github.com/getsentry/sentry-android-gradle-plugin/pull/478))
+
 ### Dependencies
 
 - Bump CLI from v2.16.1 to v2.17.4 ([#469](https://github.com/getsentry/sentry-android-gradle-plugin/pull/469), [#475](https://github.com/getsentry/sentry-android-gradle-plugin/pull/475))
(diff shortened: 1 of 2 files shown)


romtsn (approved):
@stefanosiano I've patched the PR, now we just do `maxOf` over the entire SDK name without parsing it to integer at all (maxOf will take care of ordering strings, and Android is naming new versions in alphabetical order, so we are good here).

Your approach wouldn't actually fix it, because locally you still have the sdk under `platforms/android-UpsideDownCake` and you were trying to transform it to `platforms/android-34` which would not exist then.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Parse the sdk name in BootstrapAndroidSdk ([#478](https://github.com/getsentry/sentry-android-gradle-plugin/pull/478))
```
romtsn, line 7:
```suggestion
- Do not transform the sdk name to Int in BootstrapAndroidSdk ([#478](https://github.com/getsentry/sentry-android-gradle-plugin/pull/478))
```

PR #497, opened 2023-05-30
Add license headers

### LINKED ISSUE #453: Add attribution and the text of their license at the top of the file
Example [getsentry/sentry@master/src/sentry/db/models/fields/jsonfield.py](https://github.com/getsentry/sentry/blob/master/src/sentry/db/models/fields/jsonfield.py?rgh-link-date=2023-03-22T15%3A14%3A40Z)

romtsn:
_#skip-changelog_

Closes #453


romtsn:
looks like benchmark job is failing because of the same reason as was failing the kotlin-compiler plugin (older Gradle version doesn't support composite builds). I think we'd have to update the commit hash of the duckduckgo patch, but I'd do it in a different PR

adinauer (approved):

PR #500, opened 2023-05-31
Allow setting sentry properties via plugin extension

### LINKED ISSUE #440: Add the ability to supply auth token through the Gradle Plugin extension
### Problem Statement

We depend on hosting our project's credentials via a single .properties  file which includes all the credentials/token needed to build the project, however, for Sentry it seems that we have to maintain a separate `sentry.properties` to include the token there, so I was asking if we could basically supply the auth token via the Sentry Plugin extension and read it from our central `.properties` file instead?

### Solution Brainstorm

I know that Sentry supports specifying an auth token, orgId and projectId via the sentry.properties file, but I think being able to specify these three params via the extension provides more flexibility for the user to supply these parameters via different means e.g. reading it from a file other than `sentry.properties`. 

So I can imagine something like 

```
sentry {
    orgId = providers.gradleProperty('sentryOrgId')
    projectId = providers.gradleProperty('sentryProjectId')
    authToken = providers.gradleProperty('sentryAuthToken')
}
```

romtsn:
## :scroll: Description

* Allows setting sentry properties via plugin extension
* We set the auth token as an environment variable to avoid exposing it when debug mode is enabled (otherwise it'd be printed out as a sentry-cli command argument)

Closes #440 

Manually and automated

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
Update docs


adinauer (approved):
LGTM 🚀

PR #501, opened 2023-06-01
No longer ignore `org` and `project` in `sentry` config block

adinauer:
## :scroll: Description

Pass `Provider<String>` instead of `String?` into tasks so they can lazily resolve the sentry `org` and `project` properties.

`org` and `project` in `sentry` config block was ignored.

Manually

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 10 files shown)


romtsn (approved):
Looks good 👍  Could you also update `SentrySourceContextNonAndroidTest` please to make sure we don't have this problem for non-android builds as well?

markushi (approved):

Issue #401, opened 2022-11-04
Database instrumentation enhancements


romtsn:
### Description

* New Room version ([2.5.0](https://developer.android.com/jetpack/androidx/releases/room#version_250_2)) has a new `@Upsert` query - we should check whether our instrumentation works with that already or needs to be adapted.

* Room was also converted to Kotlin - this potentially may break our instrumentation
* Moreover, androidx.sqlite has also been converted to Kotlin, so we should also check if it's still working

Ideally, we'd set up a Room version matrix on CI to test forward compatibility.

kahest:
Check the above and create issues if there are action items.

stefanosiano:
@kahest Let's consider it for the next quarter planning.

romtsn:
Closed by #502

PR #502, opened 2023-06-01
New Sqlite instrumentation

stefanosiano:
## :scroll: Description
auto install of sentry-android-sqlite module
added Sqlite new instrumentation

We want to refactor the database instrumentation moving as much as we can to the Java side, decreasing the need of bytecode manipulation.
The sdk side was convered in https://github.com/getsentry/sentry-java/pull/2722
Here is the tracking issue: https://github.com/getsentry/sentry-android-gradle-plugin/issues/401

## :pencil: Checklist

- [ ] No breaking changes


stefanosiano:
I need to wrap up and fix the last things on auto-installing the `sentry-android-sqlite` module (it's working, but tests are failing)

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/visitor/SQLiteOpenHelperMethodVisitor.kt
```diff
@@ -0,0 +1,63 @@
...
+    var alreadyInstrumented = false
+
+    override fun visitMethodInsn(
+        opcode: Int,
+        owner: String?,
+        name: String?,
+        descriptor: String?,
+        isInterface: Boolean
+    ) {
+        if (
```
romtsn, line 37: nit: I think there's one uncovered case with this approach, for example:
```kotlin
private val delegate: OpenHelper.Factory

private fun wrap(configuration: OpenHelper.Configuration): OpenHelper {
  return delegate.create(configuration)
}

override fun create(configuration: OpenHelper.Configuration): OpenHelper = wrap(configuration)
```

in this case we'd still instrument it, because `delegate.create(configuration)` is outside the `create` method scope. But I think this is still fine and it's more of an edge case; we can always come back to it if someone reports an issue.

romtsn: Another weird thing - I tried to run it locally and decompile afterwards, and I see this weird `bridge synthetic create` method in `AutoClosingRoomOpenHelperFactory`:
image

However, I have no idea where it's coming from, the [implementation](https://github.com/androidx/androidx/blob/0cd9a1b62bab73952022bbd055bcd27efa8240a9/room/room-runtime/src/main/java/androidx/room/AutoClosingRoomOpenHelperFactory.kt) doesn't look much different from the other open helpers they have. So I'm guessing it's still fine, because the other `create` method is properly not instrumented. Maybe it's a bug in the decompiler tool not sure

romtsn: bottom line: nothing to address here, just we should be aware of both of these things :)

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginAutoInstallTest.kt
```diff
@@ -162,6 +167,49 @@ class SentryPluginAutoInstallTest(
...
+    fun `sqlite is not added for lower sentry versions`() {
+        appBuildFile.appendText(
+            // language=Groovy
+            """
+            dependencies {
+              implementation 'androidx.sqlite:sqlite:2.0.0'
+            }
+
+            sentry.autoInstallation.enabled = true
+            sentry.autoInstallation.sentryVersion = "6.20.0"
```
romtsn, line 180: `l`: I guess we can be in an unlucky situation where the gradle plugin has been updated to the latest version, but the sentry version is not 6.21.0, so our users will lose on the database instrumentation. But this is probably a documentation thing, and maybe we should also mention in the changelog and migration guide.

romtsn: actually, now when I'm thinking about it - maybe we shouldn't wipe the old implementation completely, but rather keep it for compatibility? We could have two checks `isNewDatabaseInstEnabled` and `isOldDatabaseInstEnabled` - the first one would check for the `sentry-android-sqlite` presence like it does now, and if it's not available, we'd fallback to the old instrumentation (in case it's enabled) - how does that sound? 

We can think of removing it completely for the next major, for example.

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/VisitorTest.kt
```diff
@@ -101,13 +98,6 @@ class VisitorTest(
         @Parameterized.Parameters(name = "{0}/{1}")
         @JvmStatic
         fun parameters() = listOf(
-            arrayOf("androidxSqlite", "FrameworkSQLiteDatabase", AndroidXSQLiteDatabase(), null),
```
romtsn, line 104: `m`: ideally we'd also have a test for `FrameworkSQLiteOpenHelperFactory`, for example. We could pull the fixture from https://dl.google.com/android/maven2/androidx/sqlite/sqlite-framework/2.3.1/sqlite-framework-2.3.1.aar, extract a compiled `.class` from it and add it here + a test for this

romtsn (approved):
Awesome job! There's 2 points I think that need attention - compatibility and a bytecode visitor test, but the rest is looking really good.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/TracingInstrumentationExtension.kt
```diff
@@ -61,8 +61,8 @@ open class TracingInstrumentationExtension @Inject constructor(objects: ObjectFa
 
 enum class InstrumentationFeature(val integrationName: String) {
     /**
-     * When enabled the SDK will create spans for any CRUD operation performed by 'androidx.sqlite'
-     * and 'androidx.room'. This feature uses bytecode manipulation.
+     * When enabled the SDK will create spans for any CRUD operation performed by
+     * 'androidx.sqlite.db.SupportSQLiteOpenHelper'. This feature uses bytecode manipulation.
```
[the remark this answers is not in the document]
stefanosiano: yeah, there are just the Room Daos that get a special treatment, that's why I kept it  in the comment.
Not sure if it makes sense, though

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginAutoInstallTest.kt
stefanosiano: yeah, it makes sense

PR #504, opened 2023-06-05
Add SentryOkHttpEventListener automatically

stefanosiano:
## :scroll: Description
added SentryOkHttpEventListener automatic instrumentation when using the OkHttp feature

We want to improve the coverage of the http spans when using OkHttp, using the new SentryOkHttpEventListener added in https://github.com/getsentry/sentry-java/pull/2659

## :pencil: Checklist

- [x] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/okhttp/visitor/OkHttpEventListenerMethodVisitor.kt
```diff
@@ -0,0 +1,63 @@
...
+        // Add the following call at the beginning of the constructor with the Builder parameter:
+        // builder.eventListener(new SentryOkHttpEventListener(builder.eventListenerFactory));
+
+        // OkHttpClient.Builder is the parameter, retrieved here
+        visitVarInsn(Opcodes.ALOAD, 1)
+
+        // Let's declare the SentryOkHttpEventListener variable
+        visitTypeInsn(Opcodes.NEW, "io/sentry/android/okhttp/SentryOkHttpEventListener")
+
+        // Than we just copy it - I don't really know why it's needed, i just know it's needed
```
markushi, line 31: the INVOKESPECIAL below will call the init (constructor) method of the `SentryOkHttpEventListener`; since it will consume the elements on the stack and won't push anything back on the stack (`<init>` returns void) dup is required to have a reference to the `SentryOkHttpEventListener` after the constructer has been called.

markushi (approved):
LGTM!

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/okhttp/visitor/OkHttpEventListenerMethodVisitor.kt
stefanosiano: thanks for the explanation! I'll add it to the comment 😄

CHANGELOG.md
```diff
@@ -1,13 +1,16 @@
...
+## Unreleased
 
 ### Features
 
+- Add OkHttpEventListener automatically ([#504](https://github.com/getsentry/sentry-android-gradle-plugin/pull/504))
 - New Sqlite instrumentation ([#502](https://github.com/getsentry/sentry-android-gradle-plugin/pull/502))
     - This integration replaces the old database instrumentation with the `sentry-android-sqlite` integration
     - Any implementation of SupportSQLiteOpenHelper.Factory is now supported
 
+## 3.10.0
```
stefanosiano, line 12: looks like the new db instrumentation was not added in 3.10.0 (i downloaded the [source](https://github.com/getsentry/sentry-android-gradle-plugin/releases/tag/3.10.0) and checked it)
Probably it was just a wrong changelog update after a merge with master

markushi: Oh my bad then, I assumed the one on main branch was correct. Thanks for fixing 👍

stefanosiano: yeah, it stealthily slipped through to main when i merged the other pr 😅

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SpanAddingClassVisitorFactory.kt
```diff
@@ -99,6 +100,9 @@ abstract class SpanAddingClassVisitorFactory :
                         sentryModulesService.isNewDatabaseInstrEnabled() ||
                             sentryModulesService.isOldDatabaseInstrEnabled()
                     },
+                    OkHttpEventListener().takeIf {
+                        sentryModulesService.isOkHttpInstrEnabled()
```
romtsn, line 104: a little bit late to the party, but I think this one is not what we need. We added the new `SentryOkHttpEventListener` in version 6.20.0, whereas this one checks for version 5.0.0. so we'd have to introduce new `isOkHttpListenerInstrEnabled` or something like that

stefanosiano: you're right 🤔 
how I missed it...

stefanosiano: https://github.com/getsentry/sentry-android-gradle-plugin/pull/508

PR #521, opened 2023-06-29
Auto install Sentry integrations for Java Backend, Desktop, etc.

### LINKED ISSUE #503: Auto install Sentry SDK and integrations for Java
### Problem Statement

For Android we already auto install the Sentry SDK and other Sentry dependencies if the project uses libraries we have integrations for. We should provide the same functionality for Backend, Desktop, etc.

### Solution Brainstorm

Copy what's available for Android.

adinauer:
## :scroll: Description

Automatically install Sentry Java SDK and integration dependencies.

Fixes #503

## :pencil: Checklist

- [x] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/jdbc/JdbcInstallStrategy.kt
```diff
@@ -0,0 +1,92 @@
...
+            component.withModule(
+                "$MARIADB_GROUP:$MARIADB_ID",
+                JdbcInstallStrategy::class.java
+            ) {}
+            component.withModule(
+                "$POSTGRES_GROUP:$POSTGRES_ID",
+                JdbcInstallStrategy::class.java
+            ) {}
+
+            (5..14).forEach {
```
romtsn, line 77: why 5..14? 🙈

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/jdbc/JdbcInstallStrategy.kt
```diff
@@ -0,0 +1,92 @@
...
+    @Inject // inject is needed to avoid Gradle error
+    constructor() : this(SentryPlugin.logger)
+
+    override val sentryModuleId: String get() = SENTRY_JDBC_ID
+
+    override val shouldInstallModule: Boolean get() = AutoInstallState.getInstance().installJdbc
+
+    override val minSupportedThirdPartyVersion: SemVer get() = MIN_SUPPORTED_VERSION
+
+    override val minSupportedSentryVersion: SemVer get() = SemVer(4, 4, 0)
```
romtsn, line 29: you probably copy-pasted this, but we should set the min sentry version here, basically when this integration was introduced. E.g. jdbc was introduced in 5.3.0. We need this in case user uses/specifies an older version of the io.sentry:sentry module

We have to change this for all new strategies

romtsn:
looking good from the first skim, needs tests still and the changes I mentioned, but overall great stuff 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/jdbc/JdbcInstallStrategy.kt
adinauer: Because there's `ojdbc5`, `ojdbc14` and a bunch in between. Not sure if all of those exists but I guess it doesn't hurt to check here.

adinauer:
Specified versions by looking at [the registry](https://github.com/getsentry/sentry-release-registry/tree/master/packages/maven/io.sentry) and added some tests.

adinauer:
@romtsn wanna give this another pass? Should be good to go now. Just waiting for CI.

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AbstractInstallStrategy.kt
```diff
@@ -68,4 +84,22 @@ abstract class AbstractInstallStrategy : ComponentMetadataRule {
             }
         }
     }
+
+    private fun parseVersion(version: String): SemVer? {
+        val semVer = parseVersionSafely(version)
+        if (semVer == null && version.endsWith(".RELEASE")) {
```
romtsn, line 90: hm, should we just rather do `parseVersionSafely(version.removeSuffix(".RELEASE")` to avoid double-parsing altogether?

adinauer: Thinking about this again it seems the right thing. Will change it.

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AbstractInstallStrategy.kt
```diff
@@ -27,13 +29,27 @@ abstract class AbstractInstallStrategy : ComponentMetadataRule {
...
         }
-        val thirdPartySemVersion = SemVer.parse(context.details.id.version)
-        if (thirdPartySemVersion < minSupportedThirdPartyVersion) {
-            logger.warn {
-                "$sentryModuleId won't be installed because the current version is " +
-                    "lower than the minimum supported version ($minSupportedThirdPartyVersion)"
-            }
-            return
+        minSupportedThirdPartyVersion?.let {
+            parseVersion(context.details.id.version)?.let { thirdPartySemVersion ->
```
romtsn, line 33: `l`: we could also extract `parseVersion(context.details.id.version)` into a variable to avoid double-parsing the same thing

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginTest.kt
```diff
@@ -272,8 +272,9 @@ class SentryPluginTest(
             .appendArguments(":app:assembleDebug", "--info")
             .build()
 
+        val instr = "OkHttpEventListener, OkHttp"
```
romtsn, line 275: hm, has this changed 🤔 ?

romtsn (approved):
LGTM, besides a couple of comments.

Also, a very important thing that we have to do  - we should switch all of our upstream dependencies (e.g. `kotlinx-coroutines-core` for `sentry-kotlin-extensions`) from `implementation` to `compileOnly`. Auto-install works just by adding an integration into `pom.xml` of the upstream dependency, but if we don't make the upstream dependency itself a `compileOnly` one, this will result in a circular dependency and break some of the community plugins that collect dependencies. 

You can check [this test](https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginWithDependencyCollectorsTest.kt) for more details (and maybe update it with the new deps that you've introduced in this PR)

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginTest.kt
adinauer: in theory this test should've failed on `main` as well.

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AbstractInstallStrategy.kt
adinauer: my idea was to not even attempt parsing, if we don't care about the version anyways

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginTest.kt
romtsn: yeah, the problem is that it never finishes the CI run but aborts lol. good catch anyway

plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AbstractInstallStrategy.kt
romtsn: sounds good 👍

adinauer:
I've opened https://github.com/getsentry/sentry-java/pull/2837 to switch dependencies to `compileOnly`.

adinauer:
Added auto install for graphql as we've got some stuff cooking for that. Also added a bunch more tests.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginWithDependencyCollectorsNonAndroidTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+            sentry {
+              autoUploadProguardMapping = false
+              autoInstallation.enabled = true
+            }
+            """.trimIndent()
+        )
+
+        val result = runner
+            .appendArguments("app:build")
```
romtsn, line 37: I think this won't actually automatically call `collectDependencies` task from the plugin, you'd have to do it manually for non-android project, see https://github.com/mikepenz/AboutLibraries/tree/develop#gradle-api

adinauer:
CI is having issues. Merging anyways.

PR #535, opened 2023-07-28
Use Hub.getSpan instead of Sentry.getSpan

markushi:
## :scroll: Description
Switch from `Sentry.getSpan()` to `HubAdapter.getInstance().getTransaction() / getSpan()` eventually.

https://github.com/getsentry/sentry-java/pull/2855
https://github.com/getsentry/sentry-java/issues/1828

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/AbstractSpanAddingMethodVisitor.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/AbstractSpanAddingMethodVisitor.kt
@@ -64,20 +64,13 @@ abstract class AbstractSpanAddingMethodVisitor(
         visitVarInsn(Opcodes.ALOAD, hubIdx) // hub
 
         visitMethodInsn(
-            Opcodes.INVOKEINTERFACE,
+            Opcodes.INVOKEVIRTUAL,
             "io/sentry/IHub",
             "getSpan",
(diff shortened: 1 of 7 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/AbstractSpanAddingMethodVisitor.kt
```diff
@@ -41,23 +41,45 @@ abstract class AbstractSpanAddingMethodVisitor(
...
-    ISpan span = Sentry.getSpan()
+    IHub hub = HubAdapter.getInstance();
+    ISpan span = hub.getTransaction() ||  hub.getSpan();
     ISpan child = null;
     if (span != null) {
       child = span.startChild("db", <description>);
     }
     */
     protected fun MethodVisitor.visitStartSpan(
+        attachToRoot: Boolean,
```
markushi, line 52: Not sure if we need this flag. But in theory we could always attach the rooms spans to the root and the sqlite spans to the last active span

romtsn: yeah, I think I'd just do that without the flag

romtsn (approved):

PR #565, opened 2023-10-02
Instrument ContentProvider/Application onCreate calls to measure app-start performance

markushi:
## :scroll: Description

A proof-of-concept on how we could capture and create spans for Application and ContentProvider `.onCreate` calls.

A few points/questions to be answered before we can use this:
* Any class init / ctor timing is not being tracked - I guess that's fine for v1
* The Java class hierarchy (e.g. `class ContentProviderB extends ContentProviderA`) is not tracked - I guess that's fine too
* Maybe pass class names to the `.onXCreate()` calls as well? - I also don't think this is required as both Application and Content Provider classes should be not be renamed by R8/ProGuard
* Add tests

Implements parts of https://github.com/getsentry/sentry-java/issues/3035

unit / integration tests

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps

Requires a new release of `sentry-java` where the corresponding AppStartMetrics class exists - DO NOT MERGE until this is provided.
 - [ ] `sentry-java` released and min-version adapted


examples/android-instrumentation-sample/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
romtsn, line 1: `l`: don't forget to remove this class after updating the sentry-java version

- [ ] Remove it

romtsn (approved):
LGTM, great stuff and also is very concise 🚀

PR #582, opened 2023-10-23
Analytics / Telemetry

### LINKED ISSUE #223: Add analytics for the gradle plugin
We could utilise `sentry-java` for that, so we can get:
* Automatic error monitoring
* Transactions for task execution
* Transactions for tracing instrumentation
* Check adoption via release health feature (?) (what will be a "user" in our case? a project? an org? a developer?)
* Send sentry log files and the failed artifacts as attachments

Some caveats:
* It should be an opt-in feature, because it potentially will slow down the builds.

adinauer:
## :scroll: Description

We now send telemetry data unless a self hosted `url` has been set. This can be disabled via the `telemetry` option.

Closes #223 

Manually using samples

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
- Fix tests


plugin-build/build.gradle.kts
```diff
@@ -65,6 +65,11 @@ dependencies {
 
     compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin:${KotlinCompilerVersion.VERSION}")
 
+    // compileOnly since we'll be shading the common dependency into the final jar
+    // but we still need to be able to compile it (this also excludes it from .pom)
+    compileOnly(Libs.SENTRY)
```
romtsn, line 70: I think this could be just `implementation` btw. I don't think there are gradle plugins that have Sentry on their classpath, and even if there are, we'll be just using whatever is the latest version of sentry.java, which is probably fine?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,460 @@
...
+                Sentry.init { options ->
+                    options.dsn = parameters.dsn.get()
+                    options.isDebug = true
+                    options.isEnablePrettySerializationOutput = false
+                    options.tracesSampleRate = 1.0
+                    options.release = BuildConfig.Version
+                    options.isSendModules = false
+                    options.environment = parameters.buildType.get()
+                    options.setTag("SDK_VERSION", BuildConfig.SdkVersion)
+                    options.setTag("AGP_VERSION", BuildConfig.AgpVersion)
```
romtsn, line 105: `l`: we should actually read AGP version from the runtime classpath ehre, because `BuildConfig.AgpVersion` is what we package within the plugin, but since we use `compileOnly` for that, it will be overridden by whichever version users have on their classpath. You can get it via `AgpVersions.CURRENT`

plugin-build/build.gradle.kts
adinauer: Not sure, better safe than sorry here? Or do you see problems with shading?

plugin-build/src/main/kotlin/io/sentry/android/gradle/AppConfig.kt
```diff
@@ -271,6 +333,7 @@ private fun ApplicationVariant.configureProguardMappingsTasks(
 private fun ApplicationVariant.configureNativeSymbolsTask(
```
romtsn, line 333: Need to merge with `main` and change this a bit after resolving conflicts, we extracted task configuration into its own `.register` method similar to the others

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+import org.gradle.internal.operations.OperationStartEvent
+import org.gradle.process.ExecOperations
+import org.gradle.util.GradleVersion
+
+abstract class SentryTelemetryService :
+    BuildService<None>,
+    BuildOperationListener,
+    AutoCloseable {
+
+    private var hub: IHub = NoOpHub.getInstance()
```
romtsn, line 54: `m`: I believe all of this should be atomic, because it can be accessed by multiple tasks in parallel as mentioned here https://docs.gradle.org/current/javadoc/org/gradle/api/services/BuildService.html

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+        if (started) {
+            return
+        }
+
+        started = true
+        try {
+            if (startParameters.saas == false) {
+                SentryPlugin.logger.info(
+                    "Sentry is running against a self hosted instance. Telemetry has been disabled."
+                )
```
romtsn, line 69:
```suggestion
                SentryPlugin.logger.info {
                    "Sentry is running against a self hosted instance. Telemetry has been disabled."
                }
```
this will prepend the message with `[sentry]`

romtsn: I'd also replace it with our logger extension for other logs down here

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                hub = NoOpHub.getInstance()
+            } else {
+                if (!Sentry.isEnabled()) {
+                    SentryPlugin.logger.info(
+                        "Sentry telemetry is enabled. To disable set `telemetry=false` " +
+                            "in the sentry config block."
+                    )
+                    Sentry.init { options ->
+                        options.dsn = startParameters.dsn
+                        options.isDebug = true
```
romtsn, line 82: `l`: I guess we're better off setting it to `options.isDebug = startParameters.debug` which we can infer from `extension.debug`, wdyt?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                        options.tracesSampleRate = 1.0
+                        options.release = BuildConfig.Version
+                        options.isSendModules = false
+                        options.environment = startParameters.buildType
+                        options.setTag("SDK_VERSION", BuildConfig.SdkVersion)
+                        options.setTag("AGP_VERSION", AgpVersions.CURRENT.toString())
+                        options.setTag("BUILD_SYSTEM", "gradle")
+                        options.setTag("BUILD_TYPE", startParameters.buildType)
+                        options.setTag("GRADLE_VERSION", GradleVersion.current().version)
+                        startParameters.cliVersion?.let { options.setTag("SENTRY_CLI_VERSION", it) }
```
romtsn, line 93:
```suggestion
                        options.setTag("SENTRY_CLI_VERSION", BuildConfig.CliVersion)
```
forgot to change it here actually, maybe we don't need it in `startParameters` then, if it's static

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                        options.isDebug = true
+                        options.isEnablePrettySerializationOutput = false
+                        options.tracesSampleRate = 1.0
+                        options.release = BuildConfig.Version
+                        options.isSendModules = false
+                        options.environment = startParameters.buildType
+                        options.setTag("SDK_VERSION", BuildConfig.SdkVersion)
+                        options.setTag("AGP_VERSION", AgpVersions.CURRENT.toString())
+                        options.setTag("BUILD_SYSTEM", "gradle")
+                        options.setTag("BUILD_TYPE", startParameters.buildType)
```
romtsn, line 91: `m`: do we need it as a tag, if we already send it as `environment` which is also indexed and searchable?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                        }
+                    }
+                }
+                hub = Sentry.getCurrentHub()
+                startRun("gradle build ${startParameters.buildType}")
+
+                hub.configureScope { scope ->
+                    scope.user = User().also { user ->
+                        startParameters.defaultSentryOrganization.let { user.id = it }
+                        startParameters.sentryOrganization.let { user.id = it }
```
romtsn, line 109:
```suggestion
                        startParameters.defaultSentryOrganization?.let { user.id = it }
                        startParameters.sentryOrganization?.let { user.id = it }
```
shouldn't it be like this rather?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+    override fun progress(identifier: OperationIdentifier, event: OperationProgressEvent) {}
+
+    override fun finished(
+        buildOperationDescriptor: BuildOperationDescriptor,
+        operationFinishEvent: OperationFinishEvent
+    ) {
+        val details = buildOperationDescriptor.details
+
+        operationFinishEvent.failure?.let { error ->
+            // TODO check if Sentry task, otherwise ignore
```
romtsn, line 129: not a TODO anymore, right?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+
+        if (details is RunRootBuildWorkBuildOperationType.Details) {
+            endRun()
+        }
+    }
+
+    private fun isSentryError(throwable: Throwable, details: Any?): Boolean {
+        val isSentryTaskName = (details as? ExecuteTaskBuildOperationDetails)
+            ?.let {
+                it.task.name.substringAfterLast(":")
```
romtsn, line 144: nit: we could also check for the task type here, not sure if it's more reliable, but this is also fine 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+        fun createParameters(
+            project: Project,
+            variant: SentryVariant?,
+            extension: SentryPluginExtension,
+            cliExecutable: String,
+            sentryOrg: String?,
+            buildType: String
+        ): SentryTelemetryServiceParams {
+            val tags = extraTagsFromExtension(project, extension)
+            var isSaas: Boolean? = null
```
romtsn, line 238: `m`: I guess we should check here for `extension.url.orNull != null` or something like that?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+            var defaultSentryOrganization: String? = null
```
romtsn, line 239: and this one we probably don't need at all?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+            extension: SentryPluginExtension,
+            variant: SentryVariant?,
+            isSaas: Boolean?,
+            defaultSentryOrganization: String?,
+            sentryOrg: String?,
+            buildType: String,
+            tags: Map<String, String>
+        ): SentryTelemetryServiceParams {
+            var isSaas1 = isSaas
+            var defaultSentryOrganization1 = defaultSentryOrganization
```
romtsn, line 277: These two are actually unused, I guess you forgot to pass them into `SentryTelemetryServiceParams`?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+
+        fun register(project: Project): Provider<SentryTelemetryService> {
+            return project.gradle.sharedServices.registerIfAbsent(
+                getBuildServiceName(SentryTelemetryService::class.java),
+                SentryTelemetryService::class.java
+            ) {}
+        }
+
+        private fun isExecAvailable(): Boolean {
+            return GradleVersions.CURRENT >= GradleVersions.VERSION_7_5
```
romtsn, line 316: The `ValueSource` class is actually available from Gradle 6.1, so we probably don't need this check at all? https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ValueSource.html

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                "SENTRY_ignoredBuildTypes_set",
+                extension.ignoredBuildTypes.get().isNotEmpty().toString()
+            )
+            // TODO PII?
+            tags.put(
+                "SENTRY_ignoredFlavors",
+                extension.ignoredFlavors.get().isNotEmpty().toString()
+            )
+            tags.put(
+                "SENTRY_experimentalGuardsquareSupport",
```
romtsn, line 357: note: this flag was renamed to `dexguardEnabled` on main

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                "SENTRY_tracing_debug",
+                extension.tracingInstrumentation.debug.get().toString()
+            )
+            tags.put(
+                "SENTRY_tracing_forceInstrumentDependencies",
+                extension.tracingInstrumentation.forceInstrumentDependencies.get().toString()
+            )
+            tags.put(
+                "SENTRY_tracing_features",
+                extension.tracingInstrumentation.features.get().toString()
```
romtsn, line 374: idk, but this should be `joinToString` perhaps? But I'm not sure

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                extension.includeDependenciesReport.get().toString()
+            )
+            tags.put("SENTRY_includeSourceContext", extension.includeSourceContext.get().toString())
+            // TODO PII?
+            tags.put(
+                "SENTRY_additionalSourceDirsForSourceContext_set",
+                extension.additionalSourceDirsForSourceContext.get().isNotEmpty().toString()
+            )
+            // TODO PII?
+            extension.org.orNull?.let { tags.put("SENTRY_org", it) }
```
romtsn, line 403: do we need this as a tag, if we already set it as `user.id`?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+            tags.put(
+                "SENTRY_autoInstallation_sentryVersion",
+                extension.autoInstallation.sentryVersion.get().toString()
+            )
+            tags.put(
+                "SENTRY_includeDependenciesReport",
+                extension.includeDependenciesReport.get().toString()
+            )
+            tags.put("SENTRY_includeSourceContext", extension.includeSourceContext.get().toString())
+            // TODO PII?
```
romtsn, line 397: this one, I think is fine, because it's relative to the `src/` folder, so I don't think there will be any PII. Or we could also change it to just a boolean, which indicates whether someone is using this option or not

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+                getBuildServiceName(SentryTelemetryService::class.java),
+                SentryTelemetryService::class.java
+            ) {}
+        }
+
+        private fun isExecAvailable(): Boolean {
+            return GradleVersions.CURRENT >= GradleVersions.VERSION_7_5
+        }
+
+        private fun extraTagsFromExtension(
```
romtsn, line 319: Any specific reason we have to prepend every tag with SENTRY?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+            // TODO PII?
+            tags.put(
+                "SENTRY_ignoredBuildTypes_set",
+                extension.ignoredBuildTypes.get().isNotEmpty().toString()
+            )
+            // TODO PII?
+            tags.put(
+                "SENTRY_ignoredFlavors",
+                extension.ignoredFlavors.get().isNotEmpty().toString()
+            )
```
romtsn, line 355: I think these are fine to send

romtsn (approved):
Looks good overall, just some comments on the SentryTelemetryService where I wasn't sure about some things, so I haven't changed them myself. 🚀

romtsn:
Oh and maybe one more thing - maybe we should have at least one test that disables telemetry and verifies that the hub is no-op (or just a log message at least)

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: wasn't sure how this'll be queryable in dashboards

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: Is there no way to override it?

adinauer: afaik you can set `cli.executable` in the properties file and it'll override what we have in `BuildConfig`, so I guess we should keep it as is?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: Hmm if there's no need to check I'd just keep it as is for now.

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: will check

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: Hmm can't remember exactly but I think there was some problem with CI matrix tests where this failed without the check.

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: no, do we want some prefix for config tags?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: will check

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: bool sounds better

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: probably not

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -0,0 +1,474 @@
...
+            tags.put("SENTRY_includeSourceContext", extension.includeSourceContext.get().toString())
+            // TODO PII?
+            tags.put(
+                "SENTRY_additionalSourceDirsForSourceContext_set",
+                extension.additionalSourceDirsForSourceContext.get().isNotEmpty().toString()
+            )
+            // TODO PII?
+            extension.org.orNull?.let { tags.put("SENTRY_org", it) }
+            // TODO PII?
+            extension.projectName.orNull?.let { tags.put("SENTRY_projectName", it) }
```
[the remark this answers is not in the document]
adinauer: will clarify with @smeubank on whether we should remove project

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
romtsn: oh, I wasn't aware this is possible, but looks like it is https://github.com/getsentry/sentry-android-gradle-plugin/blob/dd4a7de205b4bfb2c035c0bf6af36fd4840a9164/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt#L23-L26

Yeah this should be reverted then, sorry for that!

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
romtsn: yeah environment is queryable, so we probably don't need it to save our own space

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
romtsn: sure, feel free to ignore it

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
romtsn: ok, that's also fine as it is now

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
romtsn: I think they are unique to the project you're gonna be sending telemetry, or aren't they?

adinauer: i'll remove the prefix, was more for basically grouping them in Sentry UI, but not really important

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: oops

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: it's the one read from `sentry-cli info`, i think we should use it as a fallback

romtsn: I mean you are just assigning a variable to `null` and then passing it as a method argument, why not just pass null directly (or not pass it at all for that sake?)

adinauer: ah yeah I already removed it locally

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
adinauer: `"tracing_features": "[DATABASE, FILE_IO, OKHTTP, COMPOSE]"`

I think it's OK as is.

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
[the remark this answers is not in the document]
adinauer: will comment it out for now until discussed and update later so this PR isn't blocked

PR #586, opened 2023-11-02
Print a warning if the plugin is not applied on the app module

### LINKED ISSUE #301: Warn users if the plugin is not applied to the Android app project
We could check if the project has `com.android.application` in the plugins list, and if not, warn/fail the build. Follow-up from #298

romtsn:
## :scroll: Description

In theory we could even fail the build, but I don't dare doing that because not sure how it will behave for custom build logic, so warning is good enough

Closes #301 

Manually

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Print a warning if the Sentry plugin is not applied on the app module ([#586](https://github.com/getsentry/sentry-android-gradle-plugin/pull/586))
+
 ## 3.14.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


markushi (approved):
LGTM!
For instrumentation features we should support libraries at some point as well right?

romtsn:
> LGTM! For instrumentation features we should support libraries at some point as well right?

For instrumentation it should already work, because everything is packaged into the final APK. I guess you meant auto-installation of the `-core` package, which is only installed for the current project right now. This is probably a bigger initiative because the plugin will have to be applied to the root project instead, and it's a bit trickier to support that but possible :)

Issue #403, opened 2022-11-07
Support proguard mapping upload for any JVM app


romtsn:
### Description

Since compose-desktop receives more and more traction, and people usually like to minify their apps (to reduce the final `.jar` size, and also to optimize their bytecode), we should look into how to provide automatic support for uploading proguard/R8 mapping files. 

Since there's no standardized of using proguard/r8 with pure JVM apps, I guess we'll need to provide some configuration option, which task to depend on.

This is also valid for any JVM desktop app (JavaFX, Swing, etc.) that uses a minification tool.

alexstyl:
Thanks for opening this issue @romtsn 

For future reference, you can specify where you want the mapping file to be stored by passing a custom proguard configuration like this:

```kotlin
// desktop/build.gradle.kts

compose.desktop {
        // your app's configuration 

        buildTypes.release.proguard {
            obfuscate.set(true)
            configurationFiles.from(project.file("compose-desktop.pro"))
        }
    }
}
```

and in your `compose-desktop.pro` file. use the `-printmapping` option ([see the doc](https://www.guardsquare.com/manual/configuration/usage)):
```pro
-printmapping mapping.txt

# your other proguard configurations & rules
```

leonard84:
(Moved from #429)

### Problem Statement

Currently, sentry only has a Gradle plugin for Android builds that integrates tightly into AGP.

We are developing Java applications, and want to use the release announcement, the proguard mapping file upload, and source map upload functionality in a regular `java` Gradle build.

### Solution Brainstorm

Extract common functionality from the `sentry-android-gradle-plugin` into a shared component and create a `sentry-gradle-plugin` that offers tasks to announce a release as well as upload proguard mapping files. These tasks should be able to accept all the required and optional parameters of the CLI commands. It should also register an extension to configure common parameters across the default tasks, such as the auth token.

romtsn:
@leonard84 thanks for moving it here. Could you please give more details about how you use proguard/r8? Is it through a custom task or through a gradle plugin (if so, which one)? Just wanna know better which tasks do we have to hook with. Thanks

leonard84:
It is a custom subclass of the official `proguard.gradle.ProGuardTask` to fix some issues with build caching that the original has.

Nohus:
While I appreciate that having this upload functionality in a Gradle plugin could be useful for someone, why can't we just upload the mapping files on the web interface?

[...]

PR #592, opened 2023-11-07
Use new AGP api for native symbols upload

### LINKED ISSUE #109: Support new Android Gradle APIs

romtsn:
## :scroll: Description

* This was the last remaining thing which was using an older AGP API, now we have proper decoupling per agp-compat source set

Closes #109 

Manually + automated

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
* Eventually we'd need to configure tasks on the `SentryVariant` level to have it abstracted away, and then we could even move everything to the `common` module and it could be shared between Android and Jvm plugins. Probably gonna be part of #403

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -26,6 +26,7 @@ import io.sentry.android.gradle.tasks.SentryGenerateDebugMetaPropertiesTask
 import io.sentry.android.gradle.tasks.SentryGenerateIntegrationListTask
 import io.sentry.android.gradle.tasks.SentryGenerateProguardUuidTask
 import io.sentry.android.gradle.tasks.SentryUploadProguardMappingsTask
+import io.sentry.android.gradle.tasks.configureNativeSymbolsTask
 import io.sentry.android.gradle.tasks.dependencies.SentryExternalDependenciesReportTaskFactory
 import io.sentry.android.gradle.transforms.MetaInfStripTransform
 import io.sentry.android.gradle.util.AgpVersions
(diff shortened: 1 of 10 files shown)


plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -246,9 +246,8 @@ class SentryPluginTest :
             .appendArguments(":app:assembleDebug", "--info")
             .build()
 
-        val instr = "OkHttpEventListener, OkHttp"
         assertTrue {
-            "[sentry] Instrumentable: ChainedInstrumentable(instrumentables=$instr)" in build.output
+            "[sentry] Instrumentable: ChainedInstrumentable(instrumentables=OkHttp)" in build.output
```
romtsn, line 250: I have no clue how this has been passing before 🤷 @adinauer you said once, it was failing for you because you introduced this change, but I think the test is supposed to actually test that only `OkHttp` instrumentation is applied, without the listener

romtsn: Ok, figured, it was the other test for jvm which was overwriting the build files, so the test was failing depending on the order of execution. I've removed it altogether, since we anyway test that in `SentryPluginNonAndroidAutoInstallTest`

markushi (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -273,8 +284,8 @@ private fun Variant.configureProguardMappingsTasks(
     if (isAGP74) {
         val variant = AndroidVariant74(this)
         val sentryProps = getPropertiesFilePath(project, variant)
-        val guardsquareEnabled = extension.dexguardEnabled.get()
-        val isMinifyEnabled = isMinificationEnabled(project, variant, guardsquareEnabled)
+        val dexguardEnabled = extension.dexguardEnabled.get()
```
romtsn, line 287: This slipped thru in the previous PR, so I just changed it here

PR #602, opened 2023-11-29
Do not consider user-defined sentry integration versions when auto-installing and print a warning

### LINKED ISSUE #324: AbstractInstallStrategy should warn the user about a new sentry-something version installed
### Description

This strategy could issue a warning if it counts more than one occurence of the sentry-android-core/sentry-android-okhttp/etc module in the dependency tree, which could happen in case if it comes as a transitive dependency AND our auto-installation kicks in.

We could warn a user that we've upgraded his sentry-android-something version to the latest and advice him to explicitly specify it via configuration `sentryVersion`. We only do this for the upgrade (so if they have a higher transitive version, that's probably what they want anyway and we don't warn)

romtsn:
## :scroll: Description

* Do not consider user-defined sentry integrations when auto-installing anymore. This is necessary because we want to align integrations versions to the same one as `sentry-android-core` or `sentry` or `sentry-android` is. Only if the user has directly defined one of the core versions, we will use that to install integrations, otherwise `sentryVersion` is used.

This means if the user has defined something like that:
```groovy
// direct deps
dependencies {
  implementation 'io.sentry:sentry-android-core:7.0.0'
  implementation 'io.sentry:sentry-android-okhttp:6.34.0'
}

// or with the gradle plugin
sentry {
  autoInstallation.sentryVersion = '7.0.0'
}

dependencies {
  implementation 'io.sentry:sentry-android-okhttp:6.34.0'
}
```

Then in both cases it will use `7.0.0` for the `-okhttp` integration

* We also print a warning if we have overridden user-defined version to a newer one
* Change mocked backend as the old one was not available
* Do not report errors to Sentry when running tests (telemetry)

Closes #324 

Manually + automated

## :pencil: Checklist

- [] No breaking changes

### DIFF
--- a/.github/workflows/test-matrix-agp-gradle.yaml
+++ b/.github/workflows/test-matrix-agp-gradle.yaml
@@ -103,8 +103,8 @@ jobs:
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/main' }}
           gradle-version: ${{ matrix.gradle }}
-          gradle-home-cache-excludes: |
-            caches/transforms-3
+          gradle-home-cache-includes: |
+            **/.gradle-test-kit/caches
           arguments: integrationTest
 
       - name: Upload Test Results
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,31 @@
 
 ## Unreleased
 
+### Features
+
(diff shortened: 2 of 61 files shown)


adinauer (approved):
LGTM

Issue #609, opened 2023-12-13
error: unexpected argument '--url' found


matusfedorko:
### Gradle Version

8.0

### AGP Version

8.1.0

### Code Minifier/Optimizer

None

### Version

4.0.0

### Sentry SDK Version

0.5.2

### Steps to Reproduce

We use native Sentry SDK (not the Android Sentry Sdk) and a self hosted Sentry instance. I wanted to setup debug symbol uploading for which I added Sentry Android Gradle Plugin to our project as described here: https://docs.sentry.io/platforms/android/configuration/gradle/
In the example setup I changed the following:
```
orig
projectName
url
uploadNativeSymbols
includeNativeSources
includeSourceContext
tracingInstrumentation.enabled
autoInstallation.enabled
telemetry
ignoredBuildTypes
```

### Expected Result

I expected the plugin to upload debug symbols and sources to our self hosted sentry instance.

### Actual Result

I got this error:

```
error: unexpected argument '--url' found
note: to pass '--url' as a value, use '-- --url'
```

Then I switched to Sentry Gradle plugin version 3.14.0 and everything works.

(Note I had the same error with Android Gradle plugin version 7.4.2 and Gradle 7.5.1)

PR #610, opened 2023-12-15
Fix url parameter position

markushi:
## :scroll: Description

For some reason our cli is very picky about parameter position
```shell
sentry-cli --url https://myserver.invalid/ login # works

sentry-cli login --url https://myserver.invalid/ # fails
```

Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/609

Added a unit test. In the long run it could make sense to have a base class ~`BaseSentryCliTask`, as we have similar logic spread out a few times already.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -46,7 +46,7 @@ jobs:
         if: always()
 
       - name: Build the Debug variant
-        run: ./gradlew assembleDebug
+        run: ./gradlew assembleDebug --no-daemon
 
       - name: Build the Release variant
-        run: ./gradlew assembleRelease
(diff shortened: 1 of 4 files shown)


adinauer (approved):
LGTM

adinauer:
> For some reason our cli is very picky about parameter position

That's because there's args for the main `sentry-cli` command and there's separate args for subcommands like `login` or `debug-files` (which in turn may have even more subcommands with their own args). So args have to be after the command they belong to but before the next subcommand.

The `--url` arg is part of the `sentry-cli` command itself, see https://github.com/getsentry/sentry-cli/blob/4e4a261dd998c655d3c97db285f4b214178fb666/src/commands/mod.rs#L145

PR #619, opened 2023-12-28
Fix VerifyError when optimized code is instrumented

markushi:
## :scroll: Description
App startup instrumentation fails for some pre-bundled libraries due to bytecode optimizations which produce non spec-compliant JVM bytecode.

As an example the stack frame variable of member functions at index 0 is not a `this` reference anymore.
This mainly affects `ContentProvider` instrumentation as those (opposed to `Application` classes) are usually shipped within libraries.

Affected Examples:
* `com.google.mlkit.common.internal.MlKitInitProvider`
* `androidx.startup.InitializationProvider`

Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/618

Manual testing

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/appstart/ContentProviderMethodVisitor.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/appstart/ContentProviderMethodVisitor.kt
@@ -1,9 +1,25 @@
 package io.sentry.android.gradle.instrumentation.appstart
 
 import io.sentry.android.gradle.instrumentation.MethodContext
+import io.sentry.android.gradle.instrumentation.util.Types
 import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Type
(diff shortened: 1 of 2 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/appstart/ContentProviderMethodVisitor.kt
```diff
@@ -16,10 +32,23 @@ class ContentProviderMethodVisitor(
     instrumentableContext.descriptor
 ) {
 
+    private var thisIdx = 0
+
     override fun onMethodEnter() {
-        super.onMethodEnter()
+        // add local variables, this seems to confuse the optimizer enough
```
stefanosiano, line 38: 😅

stefanosiano (approved):
let's release a hotfix and check if the user issue is gone 🤞

PR #620, opened 2023-12-28
Add VisitorTest fixture for MlKitContentProvider bug

romtsn:
_#skip-changelog_

Tested it locally with #619 reverted and it was failing with the actual reason that we've fixed:
```
Instrumented class verification failed with the following exception:
org.objectweb.asm.tree.analysis.AnalyzerException: Error at instruction 21: Argument 1: expected Landroid/content/ContentProvider;, but found Landroid/content/Context;
	at org.objectweb.asm.tree.analysis.Analyzer.analyze(Analyzer.java:292)
	at org.objectweb.asm.util.CheckClassAdapter.verify(CheckClassAdapter.java:1079)
	at io.sentry.android.gradle.instrumentation.VisitorTest.instrumented class passes Java verifier(VisitorTest.kt:82)
```


stefanosiano (approved):

markushi (approved):
Awesome, thanks for adding!

Issue #585, opened 2023-10-24
Sentry Room integration breaks Room Kotlin CodeGen


G00fY2:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.1.2

### Proguard

Enabled

### Version

6.32.0

### Steps to Reproduce

- set up a project with `androidx.room:room-*:2.6.0`
- use KSP and set [`room.generateKotlin` to true](https://developer.android.com/jetpack/androidx/releases/room#2.6.0)
- build project with R8 enabled

### Expected Result

Compilation should succeed

### Actual Result

R8 fails in Room Dao classes:
`R8: java.lang.NullPointerException: Cannot read field "d" because "<parameter1>" is null`

First I did not have Sentry in mind when we faced this issue. Therefore also check out the issue in the Android issue tracker created by me: https://issuetracker.google.com/issues/306360975

romtsn:
Hi, indeed sounds like an incompatible change have been made to Room's generated code, which breaks our auto-instrumentation. We'll look into this, thanks for reporting!

stefanosiano:
I put the generated classes and the updated config in a new branch, so we can test it: https://github.com/getsentry/sentry-android-gradle-plugin/tree/fix/room-generate-kotlin

markushi:
Let's look at the generated bytecode of the room classes. We're suspecting the lack of the return state in Kotlin (as opposed to Java) to be the culprit.

ninniuz:
Hi guys, do you have any update on this? It is currently blocking us from adopting Room Kotlin CodeGen :(

G00fY2:
@ninniuz but you know that you can [disable Sentries room integration](https://docs.sentry.io/platforms/android/integrations/file-io/#configure), right?

This is how we workaround this (ofc you loos track of your Room metrics).

ninniuz:
@G00fY2 yes I know :) But I would like to keep the Room integration on and was wondering if there was any progress on this issue first.

romtsn:
we're taking a look into this tomorrow :)

stefanosiano:
hey @ninniuz 
we released the 4.2.0 version of our plugin that should fix this issue.
Please, let us know if anything doesn't work as expected

PR #630, opened 2024-01-10
Support Room kotlin gencode

stefanosiano:
_#skip-changelog_

## :scroll: Description
removed try catch from room instrumentation, as TWP (Tracing Without Performance) makes it useless
added new Room classes (generated through ksp) to VisitorTest
In order to use new Room version and ksp, i had to update agp, gradle, kotlin, room, sdk and kotlin compiler versions

Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/585
Also, removing all try/catch management from Room instrumentation makes it simpler and more robust

Added classes generated with ksp to VisitorTest

## :pencil: Checklist

- [x] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/AndroidXRoomDaoVisitor.kt
```diff
@@ -0,0 +1,75 @@
...
+    override fun visitMethodInsn(
+        opcode: Int,
+        owner: String?,
+        name: String?,
+        descriptor: String?,
+        isInterface: Boolean
+    ) {
+        super.visitMethodInsn(opcode, owner, name, descriptor, isInterface)
+        // The backend binds the exception to the current span, via tracing without performance, so we don't need any special try/catch management.
+        if (opcode == Opcodes.INVOKEVIRTUAL) {
```
romtsn, line 45: hrm, does `INVOKEVIRTUAL` work with the `CLOSE` method? It used to be `INVOKEINTERFACE` before

romtsn (approved):
one point to clarify, but otherwise LGTM!

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/room/visitor/AndroidXRoomDaoVisitor.kt
stefanosiano: nice catch!
I updated the instrumentation to take the method type into consideration and updated the condition for the close method

PR #634, opened 2024-01-16
use source files as input for generating uuids and source bundle

### LINKED ISSUE #285: Re-running the same build gives two different io.sentry.ProguardUuids
_Platform:_
- [x]  Android -> compile 31, min 23, target 28

_Build system:_
- [x]  Gradle -> 7.4

_Android Gradle Plugin:_
- [x]  Yes -> 7.1.1
- [ ]  No

_Sentry Android Gradle Plugin:_
- [x]  Yes -> 2.1.5
- [ ]  No

_Proguard/R8:_
- [x]  Enabled
- [ ]  Disabled

_Platform installed with:_
- [ ] Maven Central
- [x] Manually

The version of the SDK:
**5.6.1**

---

**Steps to reproduce:**
Run a release build with r8 and check the `assets/sentry-debug-meta.properties` file in the APK.
Re-run the build without making any changes.

**Actual result:**
`io.sentry.ProguardUuids` inside of `sentry-debug-meta.properties` changes even though the build is otherwise identical. This lack of idempotent builds is causing issues for internal infrastructure checks we have when re-running CI builds. R8 was definitely up to date and did not re-run in between builds.

**Expected result:**
If the inputs to sentry don't change, don't generate a new uuid.

### LINKED ISSUE #616: Debug Meta Properties File Generation Breaks Build Caching
### Gradle Version

8.5.0

### AGP Version

0

### Code Minifier/Optimizer

None

### Version

4.1.0

### Sentry SDK Version

7.1.0

### Steps to Reproduce

Can reproduce with a simple Gradle Java project (Android not relevant) that has one Java source file.

1. execute `./gradlew classes`
2. execute `./gradlew classes` repeatedly, and note that the task is always "up to date", because no source files have changed.

```
BUILD SUCCESSFUL in 383ms
1 actionable task: 1 up-to-date
```

3. introduce Sentry plugin and configure it to include source context

```kotlin
plugins {
    id("java")
    id("io.sentry.jvm.gradle") version "4.1.0"
}

group = "org.example"

sentry {
    includeSourceContext = true
}

repositories {
    mavenCentral()
}
```

4. execute `./gradlew classes` repeatedly, and note that Sentry tasks are run repeatedly, despite not having changed any of their inputs.

```
BUILD SUCCESSFUL in 1s
5 actionable tasks: 3 executed, 2 up-to-date
```

### Expected Result

The meta properties file should not be regenerated if none of its inputs have changed. Regenerating this file causes other Gradle tasks (such as the `application` plugin's `distTar` task or a Docker image building task) to re-run needlessly thus negatively impacting build times.

### Actual Result

Sentry tasks that generate the meta properties file explicitly disable Gradle's build caching (see [SentryGenerateDebugMetaPropertiesTask](https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateDebugMetaPropertiesTask.kt#L21)).

### LINKED ISSUE #628: Sentry-gradle + Skaffold causes infinite restarts with Spring Boot
### Gradle Version

8.5

### AGP Version

Kotlin 1.9.22

### Code Minifier/Optimizer

None

### Version

4.1.1

### Sentry SDK Version

Spring Boot 7.1.0

### Steps to Reproduce

Kotlin + Spring Boot + Gradle

Adding the Sentry Gradle plugin, then running my local dev environment with Skaffold sync, causes Skaffold to infinitely loop its sync. Spring barely finishes starting up before it is killed and auto-restarted because of changes in disk caused by the Gradle plugin. Using Sentry Spring Boot Starter lib directly does not exhibit this behavior.

### Expected Result

Spring starts up correctly and continues running

### Actual Result

Spring is constantly restarted by Skaffold because of file changes on disk

lbloder:
## :scroll: Description
Reduce the amount of tasks that run on every build by adding input dependencies on the proguard mapping file and the source files. Also change some `upToDateWhen` rules to allow tasks to be considered `UP-TO-DATE`

Fixes #616
Fixes #628
Fixes #285 

Added automated tests to verify the behaviour.

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
All tasks that use the `sentry-cli` are still run on every build. Most likely due to `SentryCliValueSource.obtain` returning a different value at each run, when the `sentry-cli` is unpacked from resources to a temp folder. This is something we should look into to improve build times and avoid unnecessary uploads to sentry.


examples/spring-boot-sample/build.gradle.kts
```diff
@@ -53,4 +53,10 @@ sentry {
     debug.set(true)
     includeSourceContext.set(true)
     additionalSourceDirsForSourceContext.set(setOf("testsrc"))
+
+    org.set("lukas-bloder")
```
markushi, line 57: @lbloder you probably want to remove the auth-token here

plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/UploadSourceBundleTask.kt
```diff
@@ -33,6 +33,9 @@ abstract class UploadSourceBundleTask : Exec() {
             includeSourceContext.getOrElse(false) &&
                 !sourceBundleDir.asFileTree.isEmpty
         }
+
+        // Allows gradle to consider this task up-to-date if the inputs haven't changed
```
ryandens, line 37: 👋 just a fly on the wall. I'm a sentry user and a stakeholder in issue #616,  but I'm reasonably certain this is incorrect. See the JavaDoc for the `upToDateWhen` method: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskOutputs.html#upToDateWhen-groovy.lang.Closure-

>  Adds a predicate to determine whether previous outputs of this task can be reused. The given closure is executed at task execution time. The closure is passed the task as a parameter. If the closure returns false, previous outputs of this task cannot be reused and the task will be executed. That means the task is out-of-date and no outputs will be loaded from the build cache.

> You can add multiple such predicates. The task outputs cannot be reused when any predicate returns false. 


See this discussion: https://discuss.gradle.org/t/outputs-uptodatewhen-confusion/17973/3


> If you read “false” as “not”, then…
`outputs.upToDateWhen { true }` means “outputs are up to date”.
`outputs.upToDateWhen { false }` means “outputs are not up to date”.

> If any upToDateWhen spec returns false, the task is considered out of date. If they return true, Gradle does the normal behavior of checking input/output files.

This should be removed to avoid confusion, but I don't think it will negatively impact anything

markushi (approved):
Nice one, this is great!

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginSourceContextTest.kt
```diff
@@ -55,6 +58,92 @@ class SentryPluginSourceContextTest :
         assertTrue(result.output) { "BUILD SUCCESSFUL" in result.output }
     }
 
+    @Test
```
adinauer, line 61: `l` should we add more tests to check whether changing an input file actually reruns the task?

adinauer (approved):
Mostly LGTM, ideally we link the `outputs.upToDateWhen { true }` docs explanation from the comment as this will likely come up again in the future. Maybe add some negative up-to-date check tests (i.e. file changed -> rerun task).

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginSourceContextTest.kt
lbloder: This would only make sense for taks `SentryUploadProguardMappingsTask` and `UploadSourceBundleTask` which are, at the moment always executed, due to the `cliExecutable` changing at every run (see NextSteps of the PR description). Once that issue is resolved, it would absolutely make sense to create such a test, especially to make sure the behaviour hasn't changed in future gradle versions.

PR #662, opened 2024-02-26
Use mapping hash as input for generate uuid task

romtsn:
## :scroll: Description

* When `isIncludeAndroidResources` is enabled, it also runs android resources/assets generation tasks so they are available in robolectric tests. Since our `generateUuid` task was depending on the mapping file (https://github.com/getsentry/sentry-android-gradle-plugin/pull/634), `minifyReleaseWithR8` task was also triggered when running robolectric tests.
* Instead we now set file hash as task property:
  * This still ensures the task is only running when the mapping file contents changed
  * But this does not trigger the minify task for robolectric tests (which is what we want!)
  * Additional optimization is that we also read the top lines of the mapping file, which usually contains a SHA-256 hash generated by r8 itself, and use it. This helps us with big files especially, because we don't need to generate hash of their contents (which can take multiple seconds)

Internal report from a customer

manually + automated

## :pencil: Checklist

- [x] No breaking changes


adinauer (approved):
LGTM

PR #664, opened 2024-02-27
Make generating tasks cacheable

### LINKED ISSUE #653: `generateSentryDebugMetaProperties` cache busts by adding a timestamp to the `sentry-meta-debug.properties`
### Gradle Version

8.5.0

### AGP Version

8.2.2

### Code Minifier/Optimizer

None

### Version

4.2.0

### Sentry SDK Version

7.1.0

### Steps to Reproduce

1. Assemble android app with sentry android gradle plugin + build cache enabled
2. Assemble again, noticing that the assemble step does not cache

### Expected Result

Assembling the application a second time results in a cache hit

### Actual Result

Assembling the application, and open the `sentry-debug-meta.properties` file in the generated assets and observe that a timestamp is commented in the second line of the file like the following
```
#Generated by sentry-android-gradle-plugin
#Fri Feb 09 12:09:39 MST 2024
io.sentry.build-tool=gradle
```

This appears to be from using `java.util.Properties` [props.store](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#store-java.io.Writer-java.lang.String-) inside the [task action](https://github.com/getsentry/sentry-android-gradle-plugin/blob/ef9e3598a3ac9396be4b96383fc8ba3e64e8efff/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateDebugMetaPropertiesTask.kt#L42).

romtsn:
## :scroll: Description

* Add @CacheableTask annotation to tasks that generate something based on inputs from other tasks/files
* Rework the SentryGenerateIntegrationList task a bit so it receives the integrations set as input instead of the entire SentryModulesService which allowed make it cacheable as well

Closes #653 

Manually + autoamted

## :pencil: Checklist

- [x] No breaking changes


adinauer (approved):
LGTM

Issue #651, opened 2024-02-05
sentry-cli path never considered up-to-date


lbloder:
### Description

Follow-Up on: https://github.com/getsentry/sentry-android-gradle-plugin/pull/634

Currently, tasks that depend on  the `sentry-cli` are never considered `up-to-date` and thus will be executed at every run, if the `sentry-cli` path is generated by extracting the packaged `sentry-cli` executable from the `jar` and placing it into a temp folder. This causes the `sentry-cli` path to change at every run and marks all dependent tasks `out-of-date`.

Affected tasks:
- `UploadSourceBundleTask`
- `BundleSourcesTask`
- `SentryUploadProguardMappingsTask`

romtsn:
@lbloder just one note - if the build is run with `--configuration-cache` then the sentry-cli path is preserved as part of the config cache between builds, so maybe that's alright as it is now?

adinauer:
Should we document this somewhere?

romtsn:
yeah we could, but also - it's always been like that :P

lbloder:
@markushi depends on whether we actually want to implement this. But not immediately, because I'm out of office for the next two weeks.

I guess if it works with `--configuration-cache` we should be fine. 

What do you all think?

romtsn:
I think it's alright, we could mention configuration cache somewhere in the docs perhaps, and if this issue doesn't get any traction from the community in a while, we just close it.

kahest:
Let's make sure we verify that any solution to this issue also fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/670#issuecomment-1994601135

kozaxinan:
> @markushi depends on whether we actually want to implement this. But not immediately, because I'm out of office for the next two weeks.
> 
> I guess if it works with `--configuration-cache` we should be fine.
> 
> What do you all think?

Not that wouldnt be fine.
1. Not every one uses configuration cache
2. After gradle 8.6, sharing configuration cache on CI becomes much simpler. We started sharing CC cache and now Sentry invalidates CC on every run.
3.

PhilGlass:
@kozaxinan see #670.

Our workaround for now is to install the Sentry CLI to a known location in our Docker image and set `cli.executable=/opt/bin/sentry-cli` in `sentry.properties`.

kozaxinan:
I tried disabling telemetry and now it is working.
[...]

PR #675, opened 2024-03-19
make cli cache path predictable

markushi:
## :scroll: Description

Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/651

Going from
```shell
./gradlew :app:assembleRelease --configuration-cache
Configuration cache is an incubating feature.
Calculating task graph as configuration cache cannot be reused because a build logic input of type 'SentryCliValueSource' has changed.
```

to 

```shell
./gradlew :app:assembleRelease --configuration-cache
Configuration cache is an incubating feature.
Reusing configuration cache.
```

Our Gradle plugin should be cache-friendly.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
@@ -66,7 +66,7 @@ internal object SentryCliProvider {
             // otherwise we need to unpack into a file
             logger.info { "Trying to load cli from $resourcePath in a temp file..." }
 
-            loadCliFromResources(projectBuildFile, resourcePath)?.let {
+            extractCliFromResources(projectBuildFile, resourcePath)?.let {
                 logger.info { "cli extracted from resources into: $it" }
                 memoizedCliPath = it
                 return@getSentryCliPath it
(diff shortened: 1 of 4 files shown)


plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginConfigurationCacheTest.kt
```diff
@@ -121,59 +122,39 @@ class SentryPluginConfigurationCacheTest :
...
-            .lines()
-            .find {
-                it.startsWith("[sentry] cli extracted from resources into:") ||
-                    it.startsWith("[sentry] Using memoized cli path:")
-            }
-            ?.substringAfter("[sentry] cli extracted from resources into:")
-            ?.substringAfter("[sentry] Using memoized cli path:")
-            ?.trim()
-
-        val cli = File(cliPath!!).also { it.delete() }
```
romtsn, line 173: can we still have a test that deletes sentry-cli and then run the build again to verify it does not fail, but extracts the cli again?

romtsn (approved):
LGTM just left one comment about a test.

And also, would be great if we test that after updating the cli/gradleplugin version it uses the new version instead of an old one. idk if we can easily create a test for that, but verifying it locally would be enough already

javadjafari1:
it just wasted 2 days of mine.

Issue #661, opened 2024-02-24
Enable source context for non-release buildtype


Shadowsx3:
### Problem Statement

Currently by setting the includeSourceContext option to true, only the build type of release type automatically reports the source context during the assemble task. 
It would be great if other custom build types could have the ability to report it automatically by enabling the option. 
The [documentation ](https://docs.sentry.io/platforms/android/configuration/gradle/)did not specify this detail and this change would be a great help in a variety of projects.

### Solution Brainstorm

There are places in the code that even if you enable the option to do so, check that specifically the name of the build type is release. You could change that as in the following example and depend on the user configuration.

[Example](https://github.com/getsentry/sentry-android-gradle-plugin/blob/263cb93d4c01eadf9349f0c5c9cd04d06592e554/plugin-build/src/main/kotlin/io/sentry/android/gradle/AppConfig.kt#L230)

romtsn:
hi @Shadowsx3 that's a good feature request, thank you! The reason we limited this to only release builds is to not regress build speed for debug builds, but having this configurable with a list of supported build types makes a lot of sense, we'll look into implementing this!

markushi:
Let's make this configurable:
* by default do it only for release builds
* if configured, you can define all required buildTypes (`~sentry { sourceContext { enabled: true|false, buildTypes: []}}`)

markushi:
Hey @RationalRank, thanks for the additional feedback!

You could always run the upload task (e.g. `./gradlew :app:sentryUploadSourceBundleRelease`) manually to upload sources.

PR #680, opened 2024-03-26
Add autoUploadSourceContextBuildTypes property

markushi:
## :scroll: Description
Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/661

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SentryPluginExtension.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SentryPluginExtension.kt
@@ -157,6 +157,13 @@ abstract class SentryPluginExtension @Inject constructor(project: Project) {
         .property(Boolean::class.java).convention(true)
 
     /**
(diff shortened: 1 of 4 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SentryPluginExtension.kt
```diff
@@ -156,6 +156,13 @@ abstract class SentryPluginExtension @Inject constructor(project: Project) {
     val autoUploadSourceContext: Property<Boolean> = objects
         .property(Boolean::class.java).convention(true)
 
+    /**
+     * Which build types should be considered for automatic source context upload.
+     * Default is "release" only.
+     */
+    val autoUploadSourceContextBuildTypes: SetProperty<String> = objects
```
markushi, line 163: I'm not really sure if we should limit ourselves to build types, should add support for flavors too?

Issue #686, opened 2024-03-28
Release build fails on Linux CI


tananaev:
### Gradle Version

8.4

### AGP Version

8.3.1

### Code Minifier/Optimizer

R8

### Version

4.4.0

### Sentry SDK Version

7.6.0

### Steps to Reproduce

Sentry configuration:

```kt
val isCi = System.getenv("CI") == "true"
sentry {
    org.set("...")
    projectName.set("...")
    authToken.set(System.getenv("SENTRY_AUTH_TOKEN"))
    includeProguardMapping.set(isCi)
    autoUploadProguardMapping.set(isCi)
    uploadNativeSymbols.set(isCi)
    autoUploadNativeSymbols.set(isCi)
    additionalSourceDirsForSourceContext.addAll("src/googlePlay/java", "src/internal/java")

    autoInstallation {
        enabled.set(false)
    }
}
```

Run release build:
```
./gradlew clean :app:bundleFlavorRelease --no-configuration-cache
```

### Expected Result

Build should not fail.

### Actual Result

Build fails with:
```
* What went wrong:
Execution failed for task ':app:***SentryNativeSymbolsForInternalRelease'.
> A problem occurred starting process 'command '/home/runner/work/.../app/build/tmp/sentry-cli-2.28.6.exe''
```

It's unclear why it's trying to run an exe file on Linux.

romtsn:
hi @tananaev could you please post the full log (and also enable debug logging via `sentry { debug.set(true) }`? I believe the reason is something different, `.exe` is just a file extension to run the cli on Windows, but it shouldn't matter really for Linux. E.g. we run the cli on our CI just fine, see here: https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/8435337336/job/23100506620#step:4:674

tananaev:
@romtsn, here's the last part of the build log with debug enabled:
```
...
> Task :app:kaptGenerateStubsInternalReleaseKotlin
> Task :app:kaptInternalReleaseKotlin
> Task :app:compileInternalReleaseKotlin
> Task :app:compileInternalReleaseJavaWithJavac
> Task :app:processInternalReleaseJavaRes
> Task :app:transformInternalReleaseClassesWithAsm
> Task :app:mergeInternalReleaseGeneratedProguardFiles
> Task :app:mergeInternalReleaseJavaResource
> Task :app:expandInternalReleaseArtProfileWildcards
[...]

PR #691, opened 2024-04-11
Fix sentry-cli not being extracted prior to task execution

markushi:
Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/686

Created a base task for all our sentry-cli based tasks. 
During task execution the cli is then extracted if the computed path does not exist.

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -93,43 +102,37 @@ internal object SentryCliProvider {
         }
     }
 
-    internal fun searchCliInResources(resourcePath: String): String? {
-        val resourceURL = javaClass.getResource(resourcePath)
-        val resourceFile = resourceURL?.let { File(it.file) }
```
markushi, line 98: I'm really not sure why this was done this way 😅 Shouldn't it be fine to just null check the `resourceURL`?

romtsn: yeah I also don't know, null check should be fine I guess 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -495,6 +507,9 @@ abstract class SentryCliInfoValueSource : ValueSource<String, InfoParams> {
 
 abstract class SentryCliVersionValueSource : ValueSource<String, VersionParams> {
     interface VersionParams : ValueSourceParameters {
+        @get:Input
+        val projectBuildFolder: Property<File>
```
romtsn, line 511:
```suggestion
        val buildDirectory: Property<File>
```
should we align naming with the one above?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryCliExecTask.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    @get:Input
+    @get:Optional
+    abstract val debug: Property<Boolean>
+
+    @get:Input
+    abstract val cliExecutable: Property<String>
+
+    @get:Input
+    abstract val buildDirectory: Property<File>
```
romtsn, line 26: `m`: I think you don't even have to pass it as a task input, you can actually access it within the task itself: `project.layout.buildDirectory`

romtsn (approved):
LGTM, great work!

PR #708, opened 2024-05-14
Fix misleading auth-token error message in case "sentry-cli info" fails

markushi:
## :scroll: Description

Telemetry uses `sentry-cli info` to gather some env setup information. However, if no auth token is present the stderr is printed and shown as an error in IDEs like Android Studio.

As it can be fine for `sentry-cli info` to fail, simply swallow and re-print the std err and fallback to a default config.

## :pencil: Checklist

- [ ] No breaking changes


romtsn (approved):
LGTM, not sure if we need a test for this, but it's also fine without

Issue #486, opened 2023-05-10
Use toml file to share library versions


markushi:
### Problem Statement

Since we're using composite builds sharing buildSrc code between projects [caused some troubles in the past](https://github.com/getsentry/sentry-android-gradle-plugin/pull/482/files#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R48).
 

### Solution Brainstorm

Utilize sharing dependency versions using toml files, details for this can be found [in the following docs](https://docs.gradle.org/current/userguide/platforms.html).

PR #712, opened 2024-05-17
Migrate to version catalog

StefMa:
## :scroll: Description

Migrate most of the dependencies from the Dependencies.kt file into Gradles version catalog.
Basically fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/486

For include/composite builds I added a symlink to share literally anything in a single `libs.versions.toml` file.

Unfourtnaly, the I can't move anything. Thre is still stuff left in `Dependencies.kt`.
Most noteworthy the `BuildPluginsVersion.AGP` (which is fine IMHO), but also the `Libs.PROGUARD`.
I can't remove them and still need sometimes the `buildscript` block because the gradle files are in groovy and it seems there is no possibility to use the BuildPluginsVersion.AGP from a groovy script.. 🤔 Didn't worked for me.

As mentioned, this fixed https://github.com/getsentry/sentry-android-gradle-plugin/issues/486
Just wanted to tackle this 🤷 🙃 

Since this is about dependency mangament basically any task would reveal if everything is correct as dependency stuff happen on configuration time. 
However, I also create my own fork and created a PR to check stuff. 
See https://github.com/StefMa/sentry-android-gradle-plugin/pull/1

Only the matrix builds failing because of the lack of a `SENTRY_AUTH_TOKEN`.

To note also:
Gradle 7.0.4 does not work because we have to enable the `featurePreview("VERSION_CATALOGS")`.
I'm not sure if it worth it to add or just drop this test case.
This Gradle version is quite old, but this is not my decision 🙃 

## :pencil: Checklist

- [ ] No breaking changes


StefMa:
@markushi 

The tests failing with 
> Unrecognized auth token format!

secrets are not available for PRs.
See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#accessing-secrets
In your workflow file you're using `SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}`.

A potential workaorund would be to use `pull_request_target` in combination of checking out the PR branch.
In combination with required reviewers it might somewhat safe 😅 
See also https://github.com/orgs/community/discussions/50161#discussioncomment-5824158

romtsn (approved):
@StefMa sorry for holding this off for so long, we were busy with other things and haven't given much attention to the gradle plugin lately. I've updated your PR with the recent changes from main and made the necessary adjustments so it passes CI. Thanks for your contribution!

Issue #698, opened 2024-04-26
Support K2 compiler


devPalacio:
### Problem Statement

[K2 Compiler](https://kotlinlang.org/docs/k2-compiler-guide.html) will be stable with the Kotlin 2.0 release, making an issue to track when `io.sentry.SentryKotlinCompilerPlugin` is compatible with K2.

kahest:
Thanks for creating the issue 🙏

jamesarich:
Kotlin 2.0.0 compiler merge to kotlin repo announced here: https://twitter.com/kotlin/status/1785238489635684525
more info on migration here: https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-compiler.html

kahest:
Quick update - this is [in progress](https://github.com/getsentry/sentry-android-gradle-plugin/pull/716)

markushi:
Another quick update: You can expect a `4.7.0` release today which adds support for K2!

kahest:
Release is out: [4.7.0](https://github.com/getsentry/sentry-android-gradle-plugin/releases/tag/4.7.0)

dev-gvs:
I tried the new version and got the following error:

```
java.lang.NoClassDefFoundError: org/jetbrains/kotlin/ir/expressions/IrStatementOrigin$DEFAULT_VALUE
	at io.sentry.compose.JetpackComposeTracingIrExtension$generate$transformer$1.wrapExpression(JetpackComposeTracingIrExtension.kt:179)
	at io.sentry.compose.JetpackComposeTracingIrExtension$generate$transformer$1.visitCall(JetpackComposeTracingIrExtension.kt:157)
```

markushi:
@dev-gvs Which version of Kotlin are you using for your project? The mentioned missing class still exists in [2.0.0](https://github.com/JetBrains/kotlin/blob/2.0.0/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/expressions/IrStatementOrigin.kt#L97), and we've verified functionality across a few different versions (1.8.x, 1.9.x as well as 2.0.0).

LukStankovic:
> > I tried the new version and got the following error:
> > ```
> > java.lang.NoClassDefFoundError: org/jetbrains/kotlin/ir/expressions/IrStatementOrigin$DEFAULT_VALUE
[...]

PR #716, opened 2024-05-28
Add support for Kotlin K2 compiler plugin

markushi:
## :scroll: Description

As of now:
 - [x] ~tests are broken, seems to be some configuration/version issue with the compiletesting library~
```
DEFAULT
java.lang.NoSuchFieldError: DEFAULT
	at com.tschuchort.compiletesting.KotlinCompilation.<init>(KotlinCompilation.kt:143)
	at io.sentry.compose.JetpackComposeInstrumentationTest$Fixture.compileFile(JetpackComposeInstrumentationTest.kt:83)
```
 - [x] ~when plugin is applied, it no-ops with the following error. This seems to stem from some compiler side API changes, it's probably best to peek into the JPC plugin source and align with their way of doing it.~
```
w: No definition of androidx.compose.ui.Modifier found, Sentry Kotlin Compiler plugin won't run. Please ensure you're applying to plugin to a compose-enabled project.
```

Support K2, fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/698

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add support for Kotlin K2 compiler plugin ([#716](https://github.com/getsentry/sentry-android-gradle-plugin/pull/716))
+
 ## 4.6.0
 
 ### Fixes
--- a/sentry-kotlin-compiler-plugin/build.gradle.kts
+++ b/sentry-kotlin-compiler-plugin/build.gradle.kts
@@ -1,6 +1,6 @@
 plugins {
-    kotlin("jvm") version "2.0.0"
-    kotlin("kapt") version "2.0.0"
+    kotlin("jvm") version "1.9.24"
+    kotlin("kapt") version "1.9.24"
     id("distribution")
     id("com.vanniktech.maven.publish") version "0.17.0"
     id("org.jlleitschuh.gradle.ktlint") version "10.2.1"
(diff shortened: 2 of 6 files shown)


romtsn:
there's a better maintained fork of kotlin-comple-testing, maybe we should try that one? https://github.com/ZacSweers/kotlin-compile-testing

markushi:
> there's a better maintained fork of kotlin-comple-testing, maybe we should try that one? [ZacSweers/kotlin-compile-testing](https://github.com/ZacSweers/kotlin-compile-testing)

Yeah I tried, but unfortunately that fork suffers from the same limitation:
> Because the internal APIs of the Kotlin compiler often change between versions, we can only support one kotlin-compiler-embeddable version at a time.
([link](https://github.com/ZacSweers/kotlin-compile-testing#compatibility))

romtsn (approved):
LGTM, good stuff! I assume it's still backward-compatible with older kotlin compiler versions, right?

romtsn:
looks like the test is failing because we're still running against an older kotlin version. Should we introduce another dimension to the matrix for kotlin versions?

markushi:
> looks like the test is failing because we're still running against an older kotlin version. Should we introduce another dimension to the matrix for kotlin versions?

Yeah I tried to do that, but upgrading to Kotlin 2.0 breaks in several areas:
* kapt breaks in the `android-instrumentation-sample` example
* the `-Xjvm-default=enable` flag is not supported in 2.0 anymore

Hate to say it, but I'll create a follow up ticket for this. In the meantime I tested this change with different Kotlin versions locally.

PR #725, opened 2024-06-20
Fix source context with configuration cache on AGP 8+

### LINKED ISSUE #683: Configuration Cache Incompatibility
### Gradle Version

8.7

### AGP Version

8.4.0-beta01

### Code Minifier/Optimizer

None

### Version

4.4.0

### Sentry SDK Version

7.6.0

### Steps to Reproduce

Enable gradle configuration cache

### Expected Result

Build succeed instead of failing

### Actual Result

When trying to build my app module via terminal the build fails with error:

```
Configuration cache state could not be cached: field `provider` of `org.gradle.configurationcache.serialization.codecs.ProviderBackedFileCollectionSpec` bean found in field `__sourceDirs__` of task `:myapp:generateSentryBundleIdMinApi29Debug` of type `io.sentry.android.gradle.sourcecontext.GenerateBundleIdTask`: error writing value of type 'org.gradle.api.internal.provider.BiProvider'                                                                                                                                                                                                                                    
> Querying the mapped value of property(org.gradle.api.file.Directory, property(org.gradle.api.file.Directory, property(org.gradle.api.file.Directory, map(org.gradle.api.file.Directory flatmap(provider(task 'generateMinApi29DebugBuildConfig', class com.android.build.gradle.tasks.GenerateBuildConfig)) check-type())))) before task ':myapp:generateMinApi29DebugBuildConfig' has completed is not supported     
```

Using --no-configuration-cache works.

See issue https://github.com/getsentry/sentry-android-gradle-plugin/issues/554

romtsn:
## :scroll: Description

Looks like Gradle has a bug where it queries values of providers using `.zip` or `.flatMap` too early, but using `.map` works, so we just use that + `.get()` for other providers (which will be still evaluated lazily  since it's inside `.map` I believe)

Closes #683 

See updated integration test in the diff 

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix source bundles with configuration cache on AGP 8+ ([#725](https://github.com/getsentry/sentry-android-gradle-plugin/pull/725)) 
+
 ## 4.8.0
 
 ### Fixes
--- a/plugin-build/common/src/main/kotlin/io/sentry/gradle/common/JavaVariant.kt
+++ b/plugin-build/common/src/main/kotlin/io/sentry/gradle/common/JavaVariant.kt
@@ -38,7 +38,7 @@ data class JavaVariant(
                     projectDir.dir(javaDir.absolutePath)
                 }
             }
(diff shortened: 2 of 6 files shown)


markushi (approved):
Nice!

PR #850, opened 2025-03-12
Create `.git-blame-ignore-revs` and add #849 to it

lcian:
#skip-changelog

#849 touched many lines and likely impacts the usefulness of `git blame`. With this change, that particular commit will be ignored by `git blame` (feature enabled by default in GitHub in the web view, can be set with `git config blame.ignoreRevsFile .git-blame-ignore-revs` locally).

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view


romtsn (approved):
TIL!

PR #858, opened 2025-03-25
feat(multi-module): Introduce SentryRootPlugin with App and Lib subplugins

### LINKED ISSUE #800: Introduce a new `AppModuleSubplugin`
### Description

The new `AppModuleSubplugin` should be applied exclusively to the `com.android.application` modules, and register the following tasks (most parts of this is done already):

- [ ] Register `UploadProguardMappingTask`
- [ ] Register `GenereateDependencyReportTask`
- [ ] Register instrumentation transforms
- [ ] Register `UploadSourceContextTask`

romtsn:
_#skip-changelog_ 

## :scroll: Description

- Introduce a new `io.sentry.gradle` plugin that has to be applied on the root project level. The plugin then:
	- Goes over `subprojects` and applies the respective subplugins depending on what plugins are currently applied to the gradle project
	- AppSubplugin implementation remained untouched and just works, however I will refactor and extract some things later on
	- LibSubplugin implementation to follow
	- Checks if the `io.sentry.android.gradle` plugin is in place and throws as they are incompatible and can't be used together
- Deprecate `io.sentry.android.gradle` and print a message to switch over to the new root plugin
	- It also delegates to the same AppSubplugin that is used by the new root plugin
- Factor out common things into a separate file that we share among the plugins

Closes #800 

So far just manually, will add/convert existing tests later

## :pencil: Checklist

- [] I added tests to verify the changes
- [ ] No breaking changes

## :crystal_ball: Next steps
- Make LibSubplugin do the actual work
  - also investigate if we can reuse the same one for all kinds of libs, not only android 
- Make RootPlugin collect sources from all modules into a single bundle and upload it
- Introduce new SentryRootExtension and SentryProjectExtension to allow configuration per-project or globally on a root project level
- Convert tests to use the root plugin and add a single test that tests the deprecated one

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
--- a/plugin-build/src/main/kotlin/io/sentry/gradle/subplugin/RootPlugin.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/gradle/subplugin/RootPlugin.kt
@@ -1,0 +1,50 @@
+package io.sentry.gradle.subplugin
+
+import io.sentry.android.gradle.extensions.SentryPluginExtension
+import io.sentry.gradle.SENTRY_ORG_PARAMETER
+import io.sentry.gradle.SENTRY_PROJECT_PARAMETER
+import org.gradle.api.Project
+import org.gradle.api.plugins.ExtraPropertiesExtension
+import org.gradle.api.tasks.StopExecutionException
+import org.gradle.internal.build.event.BuildEventListenerRegistryInternal
+
+internal class RootPlugin(private val project: Project) {
+  init {
+    check(project == project.rootProject) {
+      "This plugin must only be applied to the root project. Was ${project.path}."
+    }
+  }
+
+  fun apply(buildEvents: BuildEventListenerRegistryInternal) =
+    project.run {
+      subprojects { p ->
+        val extraProperties = p.extensions.getByName("ext") as ExtraPropertiesExtension
+        val sentryOrgParameter =
+          runCatching { extraProperties.get(SENTRY_ORG_PARAMETER).toString() }.getOrNull()
+        val sentryProjectParameter =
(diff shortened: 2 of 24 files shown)


.editorconfig
```diff
@@ -2,7 +2,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
romtsn, line 5: This was a leftover from ktlint -> ktfmt migration

settings.gradle.kts
```diff
@@ -28,7 +31,7 @@ include(":examples:android-gradle")
 
 include(":examples:android-gradle-kts")
 
-include(":examples:android-guardsquare-proguard")
+// include(":examples:android-guardsquare-proguard")
```
romtsn, line 34: For some reason this sample doesn't work but I will figure it out later (or maybe drop the sample altogether, as it's problematic to support it given there's [no plans](https://github.com/Guardsquare/proguard/issues/280) to support AGP 8+ from the proguard side)

plugin-build/src/main/kotlin/io/sentry/gradle/subplugin/RootPlugin.kt
```diff
@@ -0,0 +1,50 @@
...
+internal class RootPlugin(private val project: Project) {
+  init {
+    check(project == project.rootProject) {
+      "This plugin must only be applied to the root project. Was ${project.path}."
+    }
+  }
+
+  fun apply(buildEvents: BuildEventListenerRegistryInternal) =
+    project.run {
+      subprojects { p ->
```
runningcode, line 20: do we know if this is isolated projects compatible?
see build logic constraints here: https://docs.gradle.org/current/userguide/isolated_projects.html#build_logic_constraints

runningcode: Gradle 8.8+ introduces the isolated project callback where you can call `gradle.lifecycle.beforeProject {` from settings.gradle to access projects in a limited way.
https://docs.gradle.org/current/userguide/isolated_projects.html#gradle_8_8

settings.gradle.kts
runningcode: I think we should drop this since sample as they don’t support AGP 8+

romtsn:
closing in favour of #1178

PR #873, opened 2025-04-29
ci(agp-matrix): Generate compatibility matrix dynamically

romtsn:
## :scroll: Description

- Introduces a new script to fetch AGP and Gradle versions dynamically
  - Uses google [maven index](https://dl.google.com/dl/android/maven2/com/android/tools/build/group-index.xml) to fetch AGP versions
  - Uses AGP-Gradle [html compat table](https://developer.android.com/build/releases/gradle-plugin#updating-gradle) to infer the supported Gradle version
  - Uses Kotlin-Gradle [html compat table](https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin) to infer the min supported Gradle version for the current Kotlin version (will parameterize it in the following PRs too)
  - We now test against two pre-release AGP versions, latest stable and a previous major stable
- The output of the script is a json payload which is then being fed into github actions and consumed by the job that runs integration tests 
- Also adds a schedule for the test matrix job to run nightly so we get notified early even if there's no activity on the repo

To proactively test against newest AGP and Gradle versions

## :pencil: Checklist

- [ ] No breaking changes


.github/workflows/test-matrix-agp-gradle.yaml
```diff
@@ -11,42 +13,42 @@ concurrency:
...
 
 jobs:
+  generate-matrix:
+    name: Generate Compat Matrix
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
```
runningcode:
```suggestion
        uses: actions/checkout@v4
```
v4 is the latest

scripts/generate-compat-matrix.main.kts
```diff
@@ -0,0 +1,305 @@
...
+  @OptIn(ExperimentalStdlibApi::class)
+  override fun run() {
+    val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+
+    /**
+     * we test against AGP:
+     * - Latest stable
+     * - Pre-release alpha
+     * - Pre-release beta/rc
+     * - Previous latest major
```
runningcode: Do you mean `Latest version from latest major release`?

scripts/generate-compat-matrix.main.kts
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  /**
+   * Fetches Kotlin minimum supported Gradle version mapping from the official Kotlin documentation
+   * @return Map where key is Kotlin version range and value is Gradle version range
+   */
+  private fun fetchKotlinGradleCompatibility(): Map<VersionRange, VersionRange> {
+    return try {
+      // Parse the official Kotlin documentation page for compatibility info
+      val html = URL("https://kotlinlang.org/docs/gradle-configure-project.html").readText()
```
runningcode, line 145: yes! ive been looking for this documentation justification for raising out minimum kotlin version compatibility. will ping you separately about this

scripts/generate-compat-matrix.main.kts
```diff
@@ -0,0 +1,305 @@
...
+      val doc = Jsoup.parse(html)
+      
+      // Look for the compatibility table
```
runningcode, line 148: this makes us dependent on the structure of the page. hopefully it is stable🤞

gradle.properties
```diff
@@ -6,3 +6,5 @@ android.useAndroidX=true
 
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
+# For AGP matrix tests, we can't infer the correct Gradle version for pre-releases
+android.overrideVersionCheck=true
```
runningcode, line 10: What exactly does this do? It seems to me that it checks that the running Gradle version is compatible with AGP version?

Why do we need this in the parent build? Is it just while we are running the entire build with different versions of Gradle?

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ kotlin = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
 kotlinAndroid = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
 kotlinSpring = { id = "org.jetbrains.kotlin.plugin.spring", version.ref = "kotlin" }
 kapt = { id = "org.jetbrains.kotlin.kapt", version.ref = "kotlin" }
```
runningcode, line 19: do we still need kapt?

runningcode (approved):
looks good! i hope those pages remain stable!

gradle/libs.versions.toml
romtsn: yep, it's used in the compiler plugin for [AutoService](https://github.com/getsentry/sentry-android-gradle-plugin/blob/dee95533666423681d1cbe0fc4877107a1362516/sentry-kotlin-compiler-plugin/build.gradle.kts#L45)

scripts/generate-compat-matrix.main.kts
romtsn: yeah that's not ideal but I think it will stay like that for a while. Otherwise we'd have to find an alternative (like looking at the kotlin github repo perhaps)

scripts/generate-compat-matrix.main.kts
romtsn: Right! will update

romtsn:
```suggestion
     * - Latest version from previous major release
```

gradle.properties
romtsn: > Is it just while we are running the entire build with different versions of Gradle?

yep that's the reason! Was failing here without that flag: https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/14733865730/job/41354917196?pr=873

PR #881, opened 2025-05-14
Correct line endings

runningcode:
## :scroll: Description

This changes the line endings for some files to align based on the
`.gitattributes` file.

This also runs the `gradlew wrapper` task to update the gradle wrapper
scripts

This also bumps the outdated development plugin version which seems to
be blocking a Gradle upgrade.

## :pencil: Checklist

- [ ] No breaking changes


examples/android-instrumentation-sample/src/main/assets/tracks.db
romtsn, line 1: We still use this file [here](https://github.com/getsentry/sentry-android-gradle-plugin/blob/b30a8b1f3b1933cfd6c002719b60ec1e11a1b12a/examples/android-instrumentation-sample/src/main/java/io/sentry/samples/instrumentation/SampleApp.kt#L22), so I guess we should keep it

settings.gradle.kts
```diff
@@ -44,10 +44,6 @@ include(":examples:multi-module-sample:spring-boot-in-multi-module-sample")
 
 include(":examples:multi-module-sample:spring-boot-in-multi-module-sample2")
 
-includeBuild("plugin-build") {
-  dependencySubstitution {
```
romtsn, line 48: hm, is this going to work for the projects under the `examples` dir? I was under the impression if we remove substitution then they'd pull the gradle plugin from remote and not use the local one?

runningcode: This substitution should work automatically. Here is the how it figures out what to substitute : https://docs.gradle.org/current/userguide/composite_builds.html#included_build_declaring_substitutions

examples/android-instrumentation-sample/src/main/assets/tracks.db
romtsn: yeah it's a db that we preload into room

settings.gradle.kts
romtsn: TIL 👍

romtsn (approved):
LGTM!

examples/android-instrumentation-sample/src/main/assets/tracks.db
runningcode: I restored the line endings to the original state.

PR #891, opened 2025-05-15
[DNM] Just testing setting default Gradle version to 8

runningcode:
## :scroll: Description

[dnm] just testing

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/gradle/libs.versions.toml
runningcode, line 1: I removed this and other symlinks because they aren’t compatible with windows and the Gradle build itself as well as sample builds like https://github.com/jjohannes/idiomatic-gradle don’t use them. Seeing as we go to lengths to use OS-independent file separators in a lot of the Gradle code, I think it makes sense to do this to keep us windows compatible.

If this breaks anyone’s workflow, let me know and we can keep them.

plugin-build/build.gradle.kts
```diff
@@ -192,6 +192,45 @@ tasks.named("distZip") {
   onlyIf { inputs.sourceFiles.isEmpty.not().also { require(it) { "No distribution to zip." } } }
 }
 
+tasks.named("distTar").configure {
```
runningcode, line 195: This fixes all those deprecation warnings in the command line of using the output of another task without declaring a dependency on that task.
This makes our build have reproducible results since the output doesn’t depend on the task order execution.

plugin-build/buildSrc/settings.gradle.kts
```diff
@@ -1,7 +1,7 @@
 dependencyResolutionManagement {
     versionCatalogs {
         create("libs") {
-            from(files("../gradle/libs.versions.toml"))
+            from(files("../../gradle/libs.versions.toml"))
```
runningcode: this was made to fix the reference to the symlink

plugin-build/src/test/kotlin/io/sentry/android/gradle/testutil/ProjectTestUtil.kt
```diff
@@ -15,6 +15,7 @@ fun createTestAndroidProject(
   val appExtension =
     project.extensions.getByType(AppExtension::class.java).apply {
       compileSdkVersion(30)
+      namespace = "com.example.app"
```
runningcode, line 18: this must be declared in AGP 8.0+

plugin-build/src/test/kotlin/io/sentry/android/gradle/util/SentryPluginUtilsTest.kt
```diff
@@ -36,36 +34,6 @@ class SentryPluginUtilsTest(private val agpVersion: SemVer) {
     assertEquals("", "".capitalizeUS())
   }
 
-  @Test
```
runningcode, line 39: I had to remove all these tests as they are now running with AGP 8 which is not compatible with proguard.
IMO, we should drop all these proguard tests completely since it isn’t really supported with AGP 8+.

gradlew
runningcode, line 1: Changes here are auto-generated when updating the wrapper.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -1,8 +1,8 @@
-import org.gradle.util.VersionNumber
+import org.gradle.util.internal.VersionNumber
 
 object BuildPluginsVersion {
-    val AGP = System.getenv("VERSION_AGP") ?: "7.4.0"
-    val GROOVY_REDISTRIBUTED = System.getenv("VERSION_GROOVY") ?: "1.2"
+    val AGP = System.getenv("VERSION_AGP") ?: "8.10.1"
```
rbro112, line 4: Is there a way to try to fallback to the version from the toml file?

rbro112 (approved):

romtsn (approved):
LGTM!

PR #894, opened 2025-05-21
Add stricter plugin validation and fix issues.

runningcode:
## :scroll: Description

This adds stricter plugin validation to check that all tasks are
declared cacheable or not and all properties are correctly marked.

This also fixes issues that the task discovered and adds it as a
dependency to the check task.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/BundleSourcesTask.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/BundleSourcesTask.kt
@@ -17,6 +17,7 @@ import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.InputFile
@@ -25,6 +26,7 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskProvider
 
+@CacheableTask
 abstract class BundleSourcesTask : SentryCliExecTask() {
 
   init {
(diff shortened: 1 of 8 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/BundleSourcesTask.kt
```diff
@@ -41,7 +43,9 @@ abstract class BundleSourcesTask : SentryCliExecTask() {
   @get:InputDirectory
   abstract val sourceDir: DirectoryProperty
 
-  @get:InputFile abstract val bundleIdFile: RegularFileProperty
+  @get:InputFile
+  @get:PathSensitive(PathSensitivity.RELATIVE)
```
romtsn, line 47: hm, I'm wondering if this should be `PathSensitivity.NONE` instead, because we care about file contents only (the bundle id), but don't care where the file is located?

plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/BundleSourcesTask.kt
```diff
@@ -25,6 +26,7 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskProvider
 
+@CacheableTask
```
romtsn, line 29: I was under the impression that Exec tasks can't be really cacheable, but looking at the inputs and outputs it makes sense! Thanks for fixing!

runningcode: Yeah, it looks like it makes sense to cache here but of course you might know better than me if there are untracked inputs/outputs here.

plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/BundleSourcesTask.kt
runningcode: thanks yes, I can adjust

runningcode: adjusted

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadProguardMappingsTask.kt
```diff
@@ -33,9 +37,13 @@ abstract class SentryUploadProguardMappingsTask : SentryCliExecTask() {
     outputs.upToDateWhen { true }
   }
 
-  @get:InputFile abstract val uuidFile: RegularFileProperty
+  @get:InputFile
+  @get:PathSensitive(PathSensitivity.RELATIVE)
```
romtsn, line 41:
```suggestion
  @get:PathSensitive(PathSensitivity.NONE)
```

PR #902, opened 2025-06-04
Update Maven Publish Plugin

runningcode:
## :scroll: Description
This updates the Maven Publish Plugin which should unblock enabling the
configuration cache and also help with updating to Gradle 8.x in this
repo.

The changelog for all the versions is here: https://github.com/vanniktech/gradle-maven-publish-plugin/blob/main/CHANGELOG.md#version-0180-2021-09-13

Note, we cannot update to a higher version until the build itself is
build with Gradle 8.x or higher.

The api to configure signing was changed in 0.20.0 this is why the
gradle property was adjusted.

#skip-changelog

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Verify artifact contents
         shell: bash
         run: |
-          zipfile=$(ls -1 plugin-build/build/distributions | grep -v "PluginMarker")
+          zipfile=$(ls -1 plugin-build/build/distributions | grep -v "PluginMarker" | grep -v ".tar")
           filename=${zipfile/\.zip/}
           unzip plugin-build/build/distributions/$filename.zip -d /tmp
           find /tmp/$filename | grep "pom-default.xml"
--- a/plugin-build/build.gradle.kts
+++ b/plugin-build/build.gradle.kts
@@ -13,7 +13,7 @@ plugins {
   id("distribution")
   alias(libs.plugins.dokka)
   id("java-gradle-plugin")
-  alias(libs.plugins.mavenPublish)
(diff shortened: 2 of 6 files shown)


romtsn:
hm, the artifacts check is failing - could you verify why? I can't grasp from the build job logs what's missing https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/15455714004/job/43507512990?pr=902

.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Verify artifact contents
         shell: bash
         run: |
-          zipfile=$(ls -1 plugin-build/build/distributions | grep -v "PluginMarker")
+          zipfile=$(ls -1 plugin-build/build/distributions | grep -v "PluginMarker" | grep -v ".tar")
```
runningcode, line 52: @romtsn seems there is now an additional `.tar` artifact which is built since we call `assemble` which assembles everything. i just filtered it out here.

plugin-build/build.gradle.kts
```diff
@@ -226,7 +225,7 @@ distributions {
   main {
     contents {
       from("build${sep}libs")
-      from("build${sep}publications${sep}maven")
+      from("build${sep}publications${sep}pluginMaven")
```
runningcode, line 228: @romtsn figured it out. we were grabbing the maven publication from the wrong folder and that wrong folder was removed in the update.

romtsn (approved):
awesome, I guess we should also test how it works releasing with [craft](https://github.com/getsentry/craft), but we'll find out when doing the next release :)

PR #905, opened 2025-06-05
Remove some buildscript blocks

runningcode:
## :scroll: Description

#skip-changelog

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/examples/android-gradle/build.gradle
+++ b/examples/android-gradle/build.gradle
@@ -1,17 +1,7 @@
-buildscript {
-    repositories {
-        google()
-        mavenCentral()
-    }
-
-    dependencies {
-        classpath Libs.AGP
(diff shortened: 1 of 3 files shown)


examples/android-gradle/build.gradle
```diff
@@ -1,18 +1,8 @@
-buildscript {
```
romtsn, line 1: i think this was here intentional to test the plugin works with the deprecated buildscript setup. But I guess it's not that relevant anymore so we should be good

runningcode: Oh, I can’t foresee a scenario where the new plugin setup would work but this would not work, but happy to add it back if you think it is needed or that happened in the past.

romtsn: nope it's all good 👍

PR #908, opened 2025-06-06
Build more in parallel

runningcode:
## :scroll: Description
This favors calling all gradle tasks in one gradle invocation in order to allow more to build in parallel.

This also favors calling gradle directly instead of using makefiles in order to remove the opaqueness of the makefiles.

#skip-changelog

## :pencil: Checklist

- [ ] No breaking changes


.github/workflows/pre-merge.yaml
```diff
@@ -44,8 +44,5 @@ jobs:
           path: plugin-build/build/reports/tests/
         if: always()
 
-      - name: Build the Debug variant
-        run: ./gradlew assembleDebug
-
-      - name: Build the Release variant
-        run: ./gradlew assembleRelease
+      - name: Build the Debug and Release variants
+        run: ./gradlew assembleDebug assembleRelease
```
romtsn: could be just `./gradlew assemble` now I guess?

plugin-build/Makefile
romtsn, line 1: I think the idea behind makefiles is that if someone not familiar with gradle (or any other build tool) checks out the repo they could just run `make ...` and it'd hide the underlying complexity. A lot of our repos have that (e.g. [python](https://github.com/getsentry/sentry-python/blob/master/Makefile) or [cocoa](https://github.com/getsentry/sentry-cocoa/blob/main/Makefile)).

I'm fine with changing github workflows, but I guess it'd be good to keep the makefiles around

romtsn (approved):

plugin-build/Makefile
runningcode: ah ok, i can add it back

.github/workflows/pre-merge.yaml
runningcode: done

PR #910, opened 2025-06-06
Reduce number of cores/threads.

runningcode:
On a modern mac machine with 10+ cores this demands too many workers.
Gradle integration tests are disk bound no I/O bound so this doesn't
speed up the build but significantly increases memory consumption.
CI has only 4 cores anyways so this should have the same effect on CI as before.
#skip-changelog

## :pencil: Checklist

- [ ] No breaking changes


markushi (approved):
LGTM, thanks for adding some rationale to the PR description!

PR #911, opened 2025-06-10
Drop support for AGP 7.0-7.3, Gradle 7.0-7.5

runningcode:
## :scroll: Description
Drop support for Android Gradle Plugin 7.0.0-7.3.X.
These versions were not being properly tested on CI and dropping support for them greatly simplifies the codebase making it easier to add new features.

Customers using AGP 7.0.0-7.3.X should continue to use Sentry Gradle Plugin 5.7.0
This also drops support for Gradle 7.0-7.5. AGP 7.4 is only compatible with Gradle 7.5 or higher so dropping both makes sense.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,12 @@
 
 - Enable caching for BundleSourcesTask ([#894](https://github.com/getsentry/sentry-android-gradle-plugin/pull/894)
 
-### Dependencies
+### Breaking Changes
 
 - Drop support for Android Gradle Plugin 7.3.X and below. To continue to use 7.3.X and below use Sentry Android Gradle Plugin 5.7.0. ([#911](https://github.com/getsentry/sentry-android-gradle-plugin/pull/911))
+
+### Dependencies
+
 - Bump Android SDK from v8.12.0 to v8.13.3 ([#896](https://github.com/getsentry/sentry-android-gradle-plugin/pull/896), [#914](https://github.com/getsentry/sentry-android-gradle-plugin/pull/914))
   - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#8133)
   - [diff](https://github.com/getsentry/sentry-java/compare/8.12.0...8.13.3)


plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryTaskProviderTest.kt
```diff
@@ -125,102 +119,6 @@ class SentryTaskProviderTest {
     assertEquals(task, getPackageBundleTask(project, "debug")?.get())
   }
 
-  @Test
```
runningcode, line 128: Deleted these tests that I think are irrelevant for AGP 7.4+

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskTest.kt
```diff
@@ -19,13 +19,27 @@ class SentryExternalDependenciesReportTaskTest {
     val project = createRegularProject()
     val output = tempDir.newFolder("dependencies")
 
-    val task: TaskProvider<SentryExternalDependenciesReportTask> =
+    val task: TaskProvider<SentryExternalDependenciesReportTaskV2> =
```
runningcode, line 22: This might be the change I am the least sure of.

I couldn’t find a test for this `SentryExternalDependenciesReportTaskV2` and it seems that the output of `V2` is different than the regular `SentryExternalDependenciesReportTask` because the task doesn’t have any information about the version numbers. Otherwise the logic is the same.

rbro112: Pre-stamped, but would be good to ensure this doesn't affect anything before merge. cc @romtsn

markushi: Seems like it was only a workaround for making [our plugin more configuration cache friendly](https://github.com/getsentry/sentry-android-gradle-plugin/blob/8818afdfc6a705a8ab7df50dd8c988d5692239c7/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskFactory.kt#L26-L27). If the V2 implementation passes the V1 test, we should be good to go

markushi: Hmm maybe we should replace `it.id.displayName` [here](https://github.com/getsentry/sentry-android-gradle-plugin/blob/8818afdfc6a705a8ab7df50dd8c988d5692239c7/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskV2.kt#L70C31-L70C38) with something like this:
```Kotlin
.map {
    val component = it.id as ModuleComponentIdentifier
    "${component.group}:${component.module}:${component.version}"
}
```

markushi (approved):
Awesome, great effort, LGTM!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Dependencies
 
+- Drop support for Android Gradle Plugin 7.3.X and below. To continue to use 7.3.X and below use Sentry Android Gradle Plugin 5.7.0. ([#911](https://github.com/getsentry/sentry-android-gradle-plugin/pull/911))
```
romtsn, line 11: So I'm thinking this is a breaking change and we should probably make a new major for this. We can re-target size analysis to also be a part of the new major and the gradually roll out (through alphas, betas, RCs). I think it would be best for both dropping support and introducing a major new feature. What do you think?

runningcode: Agreed!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskFactory.kt
romtsn, line 1: in theory could drop this factory altogether and just use the task directly

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskTest.kt
runningcode: Sounds good. I will do this in a follow up. This will be a breaking change.

plugin-build/src/main/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransform.kt
romtsn, line 1: I think we can get rid of this file altogether as it was supposed to fix an issue that was present between AGP 7.0.0 and AGP 7.1.2

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginMRJarTest.kt
romtsn, line 1: Also probably fine to delete this test after getting rid of MetaInfStripTransform

plugin-build/src/test/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransformTest.kt
romtsn, line 1: same, this test won't be needed if we delete  the transform

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskTest.kt
romtsn: hmm, something seems off. I just executed the task on the `main` branch and here's the scan: https://scans.gradle.com/s/en46g26szhnc4/timeline?details=dfdg4vqyhz6a2. You can see it uses the V2 task and if you switch over to console it outputs the deps with their respective versions.

Does it have to do with how we set up the test (i.e. not providing the versions in `artifactIds`)?

runningcode: Ah yes you were absolutely right here. I’ve added the version numbers and now there is no change in how we verify the outputs of this test.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/dependencies/SentryExternalDependenciesReportTaskFactory.kt
runningcode: Agreed, i’ll do this in a follow up.

plugin-build/src/main/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransform.kt
runningcode: nice, deleted.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPluginMRJarTest.kt
runningcode: deleted.

plugin-build/src/test/kotlin/io/sentry/android/gradle/transforms/MetaInfStripTransformTest.kt
runningcode: deleted.

CHANGELOG.md
```suggestion
### Breaking Changes

- Drop support for Android Gradle Plugin 7.3.X and below. To continue to use 7.3.X and below use Sentry Android Gradle Plugin 5.7.0. ([#911](https://github.com/getsentry/sentry-android-gradle-plugin/pull/911))

### Dependencies

```
can we also move this to a breaking changes sub-section?

PR #915, opened 2025-06-12
Add upload functionality for size analysis

runningcode:
## :scroll: Description
This adds a new task `SentryUploadAppArtifactTask` which calls to
sentry-cli mobile-app upload (apk or abb).

Most sentry functionality is enabled by the variant filter in the sentry extension. However, the size upload functionality is enabled by default when we are on CI.
This is because some teams might want to enable debug builds to upload or have some different logic than the regular variant filter.

## :pencil: Checklist

- [ ] No breaking changes


gradle/libs.versions.toml
```diff
@@ -42,6 +42,7 @@ autoServiceAnnotatons = { group = "com.google.auto.service", name = "auto-servic
 kotlinJunit = { group = "org.jetbrains.kotlin", name = "kotlin-test-junit", version.ref = "kotlin" }
 kotlinCompileTesting = { group = "dev.zacsweers.kctfork", name = "core", version = "0.7.0" }
 composeDesktop = { group = "org.jetbrains.compose.desktop", name = "desktop", version = "1.6.10" }
+truth = { module = "com.google.truth:truth", version = "1.4.4" }
```
runningcode: perhaps controversial, but this saved me a ton of time in debugging those tests.

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -243,6 +246,18 @@ fun AndroidComponentsExtension<*, *, *>.configure(
         }
       }
     }
+    if (extension.sizeAnalysis.enabled.get() == true) {
+      val sentryTelemetryProvider =
```
runningcode: not sure if it is okay to re-initialize this here. it is also initialized on line 70 per variant.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTask.kt
```diff
@@ -0,0 +1,124 @@
...
+
+  override fun getArguments(args: MutableList<String>) {
+    args.add("mobile-app")
+    args.add("upload")
+
+    val bundleFile = bundle.orNull?.asFile
+    if (bundleFile != null) {
+      if (bundleFile.exists()) {
+        args.add(bundleFile.path)
+        return
```
runningcode, line 45: actually, if we need to add more arguments to the upload (like proguard or other files), this is perhaps not the best design since it relies on early returns.

plugin-build/src/agp74/kotlin/io/sentry/android/gradle/AGP74Compat.kt
```diff
@@ -42,6 +43,11 @@ data class AndroidVariant74(private val variant: Variant) : SentryVariant {
   override val installProvider: TaskProvider<out Task>?
     get() = (variant as? VariantImpl<*>)?.taskContainer?.installTask
 
+  val bundle: Provider<RegularFile> = variant.artifacts.get(SingleArtifact.BUNDLE)
```
runningcode: these pieces of code are not compatible with AGP 7.3 and below. if we are dropping them soon then i don’t have to worry about finding a backward compatible solution.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SizeAnalysisExtension.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.gradle.api.provider.ProviderFactory
+
+open class SizeAnalysisExtension
+@Inject
+constructor(objects: ObjectFactory, providerFactory: ProviderFactory) {
+
+  val enabled: Property<Boolean> =
+    objects
+      .property(Boolean::class.java)
+      .convention(providerFactory.isCi() && false) // set to false for now otherwise upload fails CI
```
runningcode, line 16: This is hardcoded to false for the time being so that we don’t auto-upload while it isn’t working.

gradle/libs.versions.toml
rbro112: Might be good to move out into a separate change to not muddy this one, but I am pro-truth

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTask.kt
rbro112: Yeah I think we might need to add mappings for apks, so I think a when statement would be better here, like when AAB, args are just aab, when APK we could append proguard mappings (in future when added to CLI), etc.

gradle/libs.versions.toml
runningcode: Agreed, but i did some complicated truth matching to match a string in a list ending with a path so I’ll make another PR to add truth and merge that before this. See the `StringListSubject` class at the bottom of this PR.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add upload functionality for size analysis ([#915](https://github.com/getsentry/sentry-android-gradle-plugin/pull/915))
```
romtsn, line 7: I guess we'd need to expand on this, like mention the new sizeAnalysis extension and when the builds are uploaded by default, but this can be done later when we GA this

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTask.kt
```diff
@@ -0,0 +1,124 @@
...
+import org.gradle.api.tasks.TaskProvider
+import org.gradle.work.DisableCachingByDefault
+
+@DisableCachingByDefault(because = "Upload task shouldn't be cached")
+abstract class SentryUploadAppArtifactTask @Inject constructor(objectFactory: ObjectFactory) :
+  SentryCliExecTask() {
+
+  @get:InputFile
+  @get:Optional
+  @get:PathSensitive(PathSensitivity.NAME_ONLY)
```
romtsn, line 29: should this be RELATIVE? In case e.g. the variant changes (e.g. `build/outputs/${variant}/aab`)?  Same for the `apk`

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTask.kt
```diff
@@ -0,0 +1,124 @@
...
+      sentryProperties: String?,
+      sentryOrg: Provider<String>,
+      sentryProject: Provider<String>,
+      sentryAuthToken: Property<String>,
+      sentryUrl: Property<String>,
+      taskSuffix: String = "",
+    ): Pair<TaskProvider<SentryUploadAppArtifactTask>, TaskProvider<SentryUploadAppArtifactTask>> {
+      val uploadMobileBundleTask =
+        project.tasks.register(
+          "uploadSentryBundle$taskSuffix",
```
romtsn, line 87:
```suggestion
          "uploadSentryBundle${taskSuffix}ForSizeAnalysis",
```
don't know if we have to be specific here, but just a suggestion

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTaskTest.kt
romtsn, line 1: Would be cool to add an integration test - you can take a look at https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginIntegrationTest.kt which tests uploads against a simple python [server](https://github.com/getsentry/sentry-android-gradle-plugin/tree/main/test) spawned locally

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTask.kt
runningcode: This is a good point, but we might do more than just size analysis with the upload.

gradle/libs.versions.toml
runningcode: This was done in a separate PR already.

plugin-build/src/agp74/kotlin/io/sentry/android/gradle/AGP74Compat.kt
runningcode: We dropped 7.3 and below support.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTask.kt
runningcode: If the variant changes then the content of the file will change as well. I don't see why the location of file is important here though. It won't affect the directory in the uploaded zip.

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/SentryUploadAppArtifactTaskTest.kt
runningcode: i've added this! good suggestion!

plugin-build/sentry-cli.properties
```diff
@@ -1,2 +1,2 @@
-version = 2.52.0
+version = 2.53.0-alpha
```
runningcode: This is so that the integration tests use 2.53.0-alpha and pass.

CHANGELOG.md
runningcode: Yes, we'll add a lot more documentation with GA.

romtsn (approved):
Re-checked, all looks good!

PR #924, opened 2025-06-18
Faster/correct spotless configuration

runningcode:
## :scroll: Description
And fix all the files that weren’t being picked up by spotless

#skip-changelog

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,12 +15,12 @@ allprojects {
       if (name != rootProject.name) {
         kotlin {
           ktfmt(libs.versions.ktfmt.get()).googleStyle()
-          targetExclude("**/generated/**")
+          target("**/*.kt")
         }
       }
       kotlinGradle {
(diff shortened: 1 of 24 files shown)


build.gradle.kts
```diff
@@ -15,12 +15,12 @@ allprojects {
       if (name != rootProject.name) {
         kotlin {
           ktfmt(libs.versions.ktfmt.get()).googleStyle()
-          targetExclude("**/generated/**")
+          target("**/*.kt")
```
romtsn, line 18: that's so weird that you have to specify this explicitly given it's a kotlin formatter 🤷

runningcode: Agreed. Usually the way it works is that Spotless automatically infers the source sets but for Android builds it doesn’t do that. There’s a warning about it here: https://github.com/diffplug/spotless/tree/main/plugin-gradle#kotlin

PR #930, opened 2025-06-25
fix(ci): Drop unlock transforms workaround and disable maxParallelForks for integration tests

romtsn:
_#skip-changelog_

## :pencil: Checklist

- [ ] No breaking changes


plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/BaseSentryPluginTest.kt
```diff
@@ -127,12 +121,35 @@ abstract class BaseSentryPluginTest(
...
+    } catch (e: Exception) {
+      e.printStackTrace()
+    }
   }
 
   @After
   fun teardown() {
     try {
-      runner.appendArguments("app:cleanupAutoInstallState").build()
+      runner.withArguments("app:cleanupAutoInstallState").build()
```
romtsn, line 152: this was running all the previous commands as well (like assembleDebug) that we invoked during the test run. Changing this to only call the cleanup task

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/BaseSentryPluginTest.kt
```diff
@@ -108,12 +108,6 @@ abstract class BaseSentryPluginTest(
                 }
               }
             }
-
-                // unlock transforms because we're running tests in parallel therefore they may conflict
```
romtsn, line 112: This was causing configuration cache to be invalidated sometimes, like [here](https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/15877905474/job/44770117956), so our tests that test the config cache reuse were failing. So we now just run this outside the tests gradle build

runningcode (approved):
looks good! nice simplification!

PR #959, opened 2025-08-12
chore(changelog): Add 5.9.0

romtsn:
_#skip-changelog_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,17 @@
   - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#2511)
   - [diff](https://github.com/getsentry/sentry-cli/compare/2.46.0...2.51.1)
 
+## 5.9.0
+
+### Dependencies
+
+- Bump Android SDK from v8.14.0 to v8.19.0 ([#936](https://github.com/getsentry/sentry-android-gradle-plugin/pull/936), [#943](https://github.com/getsentry/sentry-android-gradle-plugin/pull/943), [#958](https://github.com/getsentry/sentry-android-gradle-plugin/pull/958))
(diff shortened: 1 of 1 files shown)


stefanosiano:
do we also need to update the version in [gradle.properties](https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/plugin-build/gradle.properties#L10)?

CHANGELOG.md
```diff
@@ -20,6 +20,17 @@
 ### Dependencies
 
 - Set build default Gradle Version to 8.14.2 ([#891](https://github.com/getsentry/sentry-android-gradle-plugin/pull/891))
+- Bump Android SDK from v8.14.0 to v8.19.0 ([#936](https://github.com/getsentry/sentry-android-gradle-plugin/pull/936), [#943](https://github.com/getsentry/sentry-android-gradle-plugin/pull/943), [#958](https://github.com/getsentry/sentry-android-gradle-plugin/pull/958))
+  - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#8190)
+  - [diff](https://github.com/getsentry/sentry-java/compare/8.14.0...8.19.0)
+- Bump CLI from v2.46.0 to v2.51.1 ([#948](https://github.com/getsentry/sentry-android-gradle-plugin/pull/948), [#957](https://github.com/getsentry/sentry-android-gradle-plugin/pull/957))
+  - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#2511)
+  - [diff](https://github.com/getsentry/sentry-cli/compare/2.46.0...2.51.1)
```
stefanosiano, line 28: actually these version updates are duplicate

romtsn: you mean they are on main already? that's right, but this PR is just to update the changelog with a version that we backported (current main is set to be shipped as v6), so it's fine to have duplicates there.

PR #974, opened 2025-09-05
fix(compat): Fix build compatibility with AGP 9.+

### LINKED ISSUE #776: Lazy variant details aren't supported for ProGuard mappings upload
### Gradle Version

8.10.2

### AGP Version

8.7.0

### Code Minifier/Optimizer

R8

### Version

4.11.0

### Sentry SDK Version

7.15.0

### Steps to Reproduce

1. Configure your app's version code and name via the property-based APIs AGP offers for that these days.
    1. A recipe doing so in one way (manifest transformation) can be seen [here](https://github.com/android/gradle-recipes/tree/agp-8.5/transformManifest).
    2. Setting the variant outputs' properties can also be used. See a usage example from Slack [here](https://github.com/slackhq/foundry/blob/8228e72ef7f76a283df544793a8b9f68adceda46/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/ApkVersioningPlugin.kt#L148-L149).
3. Use this plugin to upload ProGuard mapping files for your release build.
4. Check out the uploaded mappings at https://example.sentry.io/settings/projects/myproject/proguard/.
5. Notice that all of them are linked to the non-existing release `com.example.myproject@undefined`.
    1. This means it's quite difficult to find the mappings used in a release.
        1. Why do I care? Because I log (to Sentry) a bunch of things that aren't exceptions, Sentry's automatic deobfuscation doesn't apply. To deobfuscate logged details manually, I need the correct mappings file of the release that raised an issue. While it's rare for me to apply this manual process, it's still crucial to allow understanding all Sentry errors. Downloading the mappings file from Sentry allows there to be a single source of truth when it comes to all things related to a release.
    3. Note: This is _not_ breaking Sentry's automatic deobfuscation of stack traces. That still works fine thanks to the UUID-based lookup.
    4. As seen [here](https://github.com/getsentry/sentry-android-gradle-plugin/blob/71333dcf60b438263221342a4c5214973fc37c5b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt#L477), that seems to be expected at this time since the plugin doesn't support all the various ways the version details can be set in newer AGP versions.

### Expected Result

The lazily configured version details should be used when uploading ProGuard mappings.

### Actual Result

The statically set version values are used when uploading ProGuard mappings.

romtsn:
Closes #776

### DIFF
--- a/plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/BaseSentryPluginTest.kt
+++ b/plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/BaseSentryPluginTest.kt
@@ -1,7 +1,6 @@
 package io.sentry.android.gradle.integration
 
 import io.sentry.android.gradle.util.PrintBuildOutputOnFailureRule
-import io.sentry.android.gradle.util.SemVer
 import java.io.ByteArrayOutputStream
 import java.io.File
 import java.io.OutputStreamWriter
@@ -55,12 +54,19 @@ abstract class BaseSentryPluginTest(
         """
             import io.sentry.android.gradle.autoinstall.AutoInstallState
             import io.sentry.android.gradle.util.GradleVersions
+            import org.gradle.util.internal.VersionNumber
 
             buildscript {
               repositories {
                 google()
                 gradlePluginPortal()
                 mavenCentral()
+                maven {
+                  url 'https://storage.googleapis.com/r8-releases/raw'
+                  content {
+                    includeGroup 'com.android.tools'
+                  }
+                }
               }
               dependencies {
                 classpath 'com.android.tools.build:gradle:$androidGradlePluginVersion'
@@ -68,6 +74,11 @@ abstract class BaseSentryPluginTest(
                 // withPluginClasspath on the Gradle Runner.
                 $additionalBuildClasspath
                 classpath files($pluginClasspath)
+                if (VersionNumber.parse("$androidGradlePluginVersion").major < 8) {
+                  // AGP 7.x has troubles with compileSdk 34 due to some R8 shenanigans, so we have to use a newer
+                  // version of R* here
+                  classpath 'com.android.tools:r8:8.11.18'
+                }
               }
             }
 


examples/multi-module-sample/build.gradle.kts
```diff
@@ -1,14 +1,15 @@
...
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
   alias(libs.plugins.kotlin)
   id("io.sentry.jvm.gradle")
 }
 
-tasks.withType<KotlinCompile>().configureEach {
-  kotlinOptions {
+tasks.named<KotlinCompile>("compileKotlin") {
```
runningcode: isn’t the new way to do this something like this?
```suggestion
kotlin {
```

runningcode (approved):
Thank you for doing this!

examples/android-room-lib/build.gradle.kts
```diff
@@ -5,15 +5,18 @@ plugins {
...
-    minSdk = LibsVersion.MIN_SDK_VERSION
-    targetSdk = LibsVersion.SDK_VERSION
-  }
+  defaultConfig { minSdk = LibsVersion.MIN_SDK_VERSION }
 
-  kotlinOptions { jvmTarget = JavaVersion.VERSION_1_8.toString() }
   namespace = "io.sentry.android.instrumentation.lib"
+
+  compileOptions {
+    sourceCompatibility = JavaVersion.VERSION_11
```
runningcode: Should we just go straight to 17?

settings.gradle.kts
```diff
@@ -3,6 +3,7 @@ pluginManagement {
     gradlePluginPortal()
     mavenCentral()
     google()
+    maven { url = uri("https://storage.googleapis.com/r8-releases/raw") }
```
runningcode: Can we use repository filter here? https://docs.gradle.org/current/userguide/filtering_repository_content.html

It speeds up performance scanning through the repositories.

romtsn: sure! good idea

examples/android-room-lib/build.gradle.kts
romtsn: let me try that 👍

PR #976, opened 2025-09-08
feat(vcs): Add buildConfiguration to VCSExtension

runningcode:
## Summary
- Add `buildConfiguration` field to `VcsInfoExtension` that maps to the `build_configuration` CLI argument
- Field defaults to the build variant (combination of flavor and build type) if not explicitly set
- Enables users to customize the build configuration sent to sentry-cli for build uploads

## Changes
- **VcsInfoExtension.kt**: Added `buildConfiguration` property with documentation
- **SentryUploadAppArtifactTask.kt**: 
  - Added `buildConfiguration` input property
  - Added `--build-configuration` CLI argument handling
  - Updated task registration to set default value from build variant
- **AndroidComponentsConfig.kt**: Pass build variant as default value to task registration

🤖 Generated with [Claude Code](https://claude.ai/code)


plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/VcsInfoExtension.kt
```diff
@@ -59,4 +59,11 @@ open class VcsInfoExtension @Inject constructor(objects: ObjectFactory) {
    * will be used.
    */
   val prNumber: Property<Int> = objects.property(Int::class.java).convention(null as Int?)
+
+  /**
+   * The build configuration to use for the upload. If not provided, the build variant
+   * will be used.
+   */
+  val buildConfiguration: Property<String> =
```
runningcode: I think this is the best place for this because it will be shared between snapshots and size analysis.
However, maybe to make it fit better, We should rename this file to `cliInfo`? Thoughts? I will do that in a separate PR however we decide.

romtsn: hmm, I'm thinking - is it not a common case where I'd have a different variant for snapshots than for size analysis (from a single build)? Just thinking if it'd make sense to extract this into its own sub-extension which size analysis and snapshots would embed.

runningcode: Yes that’s a good point about this field. It is definitely a case where you would want a different variant for snapshots than size analysis. For example snapshots with debug builds and size analysis on release builds.

By default this is kind of like a “tag” to the build. So what this means is that I cannot compare a “release” build with a “debug” build. That is all that it is used for. If you did want to compare release builds with debug builds then you would override this field to say “debug” for both (a contrived example).

I’ve updated the description to be more accurate here. What do you think?

romtsn: ah I see, thanks for the clarification! Alright, I'm fine with keeping it in this extension too

PR #980, opened 2025-09-09
Auto install Spring Boot 4 and Spring 7 modules

adinauer:
## :scroll: Description

Do not auto install Spring Boot 3 / Spring 6 dependencies for Spring Boot 4 / Spring 7 and instead install the correct dependencies.

We've recently added support for Spring Boot 4 (pre releases) but the Gradle plugins still tries to add our dependency for Spring Boot 3 (sentry-spring-boot-jakarta) instead of (sentry-spring-boot-4), which cause the application to fail to compile/start.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@
 bin/
 gen/
 out/
-release/
 
 # Gradle files
 .gradle/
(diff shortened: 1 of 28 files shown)


lcian (approved):
Looks good to me.
I think the failures on AGP 9-alpha4 are expected as I see it that build also fails on main currently.

PR #986, opened 2025-09-17
Add Distribution API to Sentry Android Gradle Plugin

runningcode:
## Summary
- Add `DistributionExtension` with `enabledVariants` property
- Add `distribution` configuration DSL to `SentryPluginExtension`
- Mark API as `@Experimental`

## Details
This adds the API to specify which build variants should have distribution enabled. **No functionality is wired yet** - this is just the API surface.

This API is **opt in** (unlike the rest of Sentry features) because users need to explicitly choose which variants get build distribution. For example: a typical set up will disable distribution for release builds but enable them for internal testing builds like a nightly release or alpha testing. Users will also typically want this behavior disabled for debug builds.

Usage:
```gradle
sentry {
  distribution {
    enabledVariants.set(["nightlyRelease", "alphaRelease"])
  }
}
```

#skip-changelog (Until this feature is wired up)

Closes EME-277

🤖 Generated with [Claude Code](https://claude.ai/code)


plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SentryPluginExtension.kt
```diff
@@ -111,6 +111,13 @@ abstract class SentryPluginExtension @Inject constructor(project: Project) {
     sizeAnalysisAction.execute(sizeAnalysis)
   }
 
+  val distribution: DistributionExtension = objects.newInstance(DistributionExtension::class.java)
+
+  @org.jetbrains.annotations.ApiStatus.Experimental
```
chromy: It this normal to use for the Sentry APIs? I don't see it anywhere else in sentry-android-gradle-plugin / sentry-java - but maybe this is the only Experimental API at present.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import javax.inject.Inject
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.SetProperty
+import org.jetbrains.annotations.ApiStatus.Experimental
+
+@Experimental
+open class DistributionExtension @Inject constructor(objects: ObjectFactory) {
+
+  /** Set of Android build variants that should have build distribution enabled. */
```
chromy: Lets call out explicitly the interaction between this and `ignored{BuildTypes,Flavors,Variants}` - if something is both disabled globally and enabled here what happens? (I think there is an argument for either option so we could do whatever is easiest but would be good to be clear either way)

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
```diff
@@ -0,0 +1,14 @@
...
+  val enabledFor: SetProperty<String> =
```
chromy: I think this could be `enabledVariants` just to make it 100% clear what you need to put there (and to leave room if we ever wanted to add other `enabledFoos` APIs) but up to you.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SentryPluginExtension.kt
runningcode: I discussed this with @romtsn.  This is the only one here in this codebase but it is used a lot in the `sentry-java` codebase.
In the `sentry-java` codebase new features either get that `ApiStatus.Experimental` annotation or the options are put in the `ExperimentalOptions`.
I like this annotation because it alerts users in AndroidStudio/Intellij.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
runningcode: Good point. i wrote that it isn’t related to those and updated the PR. I’m open to making it related but I think that get’s confusing and also limits options of consumers of the API.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
runningcode: enabledVariants sounds good! updated.

romtsn (approved):
considering the comments from @chromy all lgtm 👍

chromy (approved):
lgtm, thanks!

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
chromy: Thanks!

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
chromy: Thanks!

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SentryPluginExtension.kt
chromy: sg!

PR #999, opened 2025-10-06
Add GenerateDistributionPropertiesTask for distribution options

runningcode:
## Summary

Implements EME-397 by adding support for setting `SentryOptions.DistributionOptions` from Gradle using the existing properties file mechanism. There are probably many ways of passing these options to the client but using properties files doesn't require the client to recompile resources or code and there is already an existing mechanism in the codebase to do this so I just piggy-backed on that mechanism.

This PR adds a `GenerateDistributionPropertiesTask` that generates distribution-related properties which are bundled into `sentry-debug-meta.properties` and read by the SDK at runtime.

## Example Usage

```gradle
sentry {
  org.set("my-org")
  projectName.set("my-project")
  authToken.set(System.getenv("SENTRY_AUTH_TOKEN"))
  
  distribution {
    enabledVariants.addAll("debug", "staging")
  }
}
```

#skip-changelog since this is for an unreleased feature that isn’t complete yet

## Sentry-java

The sentry-java PR that implements this is here: https://github.com/getsentry/sentry-java/pull/4784


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -0,0 +1,76 @@
...
+      orgAuthToken.orNull?.let { writer.appendLine("$ORG_AUTH_TOKEN_PROPERTY=$it") }
+      writer.appendLine("$BUILD_CONFIGURATION_PROPERTY=${buildConfiguration.get()}")
+    }
+  }
+
+  companion object {
+    internal const val SENTRY_DISTRIBUTION_OUTPUT = "sentry-distribution.properties"
+    const val ORG_SLUG_PROPERTY = "io.sentry.distribution.org-slug"
+    const val PROJECT_SLUG_PROPERTY = "io.sentry.distribution.project-slug"
+    const val ORG_AUTH_TOKEN_PROPERTY = "io.sentry.distribution.org-auth-token"
```
chromy: This probably wants to be:
```
const val DISTRIBUTION_AUTH_TOKEN_PROPERTY = "io.sentry.distribution.auth-token"
```
or something.

I think it's not given that this will always be a Sentry 'organization token'. One idea from security was to use a 'custom integration' which lets you generate a custom token with arbitrary scopes (and we would add a 'disto' scope) rather than trying to reuse org tokens which have some strange permissions (write only access to releases etc). See https://www.notion.so/sentry/Security-Emerge-Distro-2738b10e4b5d807ca777e70c3232cccd#2738b10e4b5d807ca777e70c3232cccd.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -0,0 +1,76 @@
...
+      taskSuffix: String,
+      buildConfiguration: String,
+    ): TaskProvider<GenerateDistributionPropertiesTask> {
+      return project.tasks.register(
+        "generateSentryDistributionProperties$taskSuffix",
+        GenerateDistributionPropertiesTask::class.java,
+      ) { task ->
+        task.output.set(output)
+        task.withSentryTelemetry(extension, sentryTelemetryProvider)
+        // TODO we should check if the org and project are available in sentry.properties
```
chromy, line 67: Yeah would be ideal if we could default these automatically given you have to have both the DSN and the main auth token it seems like it ought to be possible somehow but I guess Slack will know.

runningcode: Good point. I created EME-398 to fix this (out of scope for this PR).

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
runningcode: That sounds good. I'll rename!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -0,0 +1,76 @@
...
```
romtsn, line 67: oh yeah, this is a good point! We also check `ext` for org and project names, but I'm not sure if anyone is really using extras to set that. Maybe we should get rid of it in the upcoming major altogether

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -0,0 +1,76 @@
...
+        task.orgSlug.set(extension.org)
+        task.projectSlug.set(extension.projectName)
+        // TODO we should have a separate authToken for Build Distribution
+        task.orgAuthToken.set(extension.authToken)
```
romtsn, line 71: I guess it's more of a security topic, but idk if it's fine to inject and expose the auth token at runtime, given that we recommend hiding it in an env variable even on CI, for example.

runningcode: This is a good point. We will make these custom grained token in the future: See Hector’s comment here: https://github.com/getsentry/sentry-android-gradle-plugin/pull/999#discussion_r2410185138

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
runningcode: Ah good point. I have a ticket to follow up on this. I added a note about ext. https://linear.app/getsentry/issue/EME-398/support-environment-variables-and-sentryproperties-for

PR #1000, opened 2025-10-08
Add fallback pattern for org and project in GenerateDistributionPropertiesTask

runningcode:
## Summary

Update `GenerateDistributionPropertiesTask` to check multiple sources for `org` and `project` configuration, following the established pattern used by other tasks like `SentryUploadProguardMappingsTask`.

### Fallback Pattern Implementation

**For org (organization slug):**
1. Extra properties (`ext.sentryOrg`)
2. Extension property (`extension.org`)
3. Environment variable (`SENTRY_ORG`)
4. sentry.properties file (`defaults.org` or `org`)

**For projectName (project slug):**
1. Extra properties (`ext.sentryProject`)
2. Extension property (`extension.projectName`)
3. Environment variable (`SENTRY_PROJECT`)
4. sentry.properties file (`defaults.project` or `project`)

**For authToken:**
- Only from extension property (`extension.authToken`) - no fallback added

### Code Changes

- Updated `GenerateDistributionPropertiesTask.register()` to implement fallback logic
- Updated `configureDistributionPropertiesTask()` to accept and pass `sentryOrg`/`sentryProject` parameters
- Added 5 comprehensive test cases covering all fallback scenarios

## Testing

All tests pass:
- ✅ Extension properties (baseline)
- ✅ Extra properties override extension properties
- ✅ Fallback to sentry.properties when extension not set
- ✅ Extension properties override sentry.properties
- ✅ Handles missing values gracefully

## Notes

- This follows the same pattern as other tasks in the codebase
- Uses the same environment variable names as sentry-cli (`SENTRY_ORG`, `SENTRY_PROJECT`)
- Auth token intentionally kept simple (extension-only) per discussion

Resolves EME-397


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -57,18 +61,44 @@ abstract class GenerateDistributionPropertiesTask : PropertiesFileOutputTask() {
...
-        // TODO we should check if the org and project are available in sentry.properties
-        task.orgSlug.set(extension.org)
-        task.projectSlug.set(extension.projectName)
-        // TODO we should have a separate authToken for Build Distribution
+
+        // Load sentry.properties if available
+        val sentryPropertiesFile = variant?.let { SentryPropertiesFileProvider.getPropertiesFilePath(project, it) }
+        val sentryProperties = sentryPropertiesFile?.let { PropertiesUtil.loadMaybe(File(it)) }
+
+        // Resolve org slug with fallback chain: ext -> extension -> env -> sentry.properties
```
runningcode, line 92: Happy to discuss a different priority here.
Note: we're not getting `.sentryclirc` files yet. That can be added later.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -57,18 +61,47 @@ abstract class GenerateDistributionPropertiesTask : PropertiesFileOutputTask() {
...
+          project.provider {
+            sentryProject
+              ?: extension.projectName.orNull
+              ?: System.getenv("SENTRY_PROJECT")
+              ?: sentryProperties?.getProperty("defaults.project")
+              ?: sentryProperties?.getProperty("project")
+          }
+        task.projectSlug.set(projectProvider)
+
+        // Auth token only from extension (no fallback)
```
romtsn, line 120: so this is deliberate? 'cause I think you can also set it via env variable (SENTRY_AUTH_TOKEN) and sentry.properties too. But I guess it's not gonna be the one you need for distribution, right?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -57,18 +61,47 @@ abstract class GenerateDistributionPropertiesTask : PropertiesFileOutputTask() {
...
         task.withSentryTelemetry(extension, sentryTelemetryProvider)
-        // TODO we should check if the org and project are available in sentry.properties
-        task.orgSlug.set(extension.org)
-        task.projectSlug.set(extension.projectName)
-        // TODO we should have a separate authToken for Build Distribution
+
+        // Load sentry.properties if available
+        val sentryPropertiesFile =
+          variant?.let { SentryPropertiesFileProvider.getPropertiesFilePath(project, it) }
+        val sentryProperties = sentryPropertiesFile?.let { PropertiesUtil.loadMaybe(File(it)) }
```
romtsn: wondering if that's fine to load it at configuration phase here, or should we move it to the task action?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
runningcode: yes that’s correct, we’re going to add a separate env variable and extension property for this one in the future which is why it isn’t added here.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
runningcode: Ah good point. This should probably be marked as an input as well.

PR #1001, opened 2025-10-09
feat: Auto-install distribution dependency when enabled for variant

runningcode:
## Summary

When build distribution is enabled for a variant (via `distribution.enabledVariants`) and `autoInstallation.enabled = true`, the plugin now automatically adds the `sentry-android-distribution` dependency to that specific variant.

Note: the library is only released as an alpha 8.24.0-alpha.1. You'll need to set `sentry.autoInstallation.sentryVersion="8.24.0-alpha.1"` for this to work.

## Changes

- Added `SENTRY_ANDROID_DISTRIBUTION` module to `SentryModules`
- Implemented logic in `configureDistributionPropertiesTask` to add the distribution dependency per variant when both distribution and autoInstallation are enabled
- Added 3 integration tests to verify the behavior

Refs: EME-278

#skip-changelog as this is part of the unreleased build distribution feature

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt
@@ -31,7 +31,8 @@ internal const val SENTRY_GROUP = "io.sentry"
 // Note: sentry-android-distribution is not included here because it requires variant-specific
 // installation logic. Unlike other integrations that are installed globally when their
 // dependencies are detected, distribution must be installed per-variant based on
-// extension.distribution.enabledVariants. See AndroidComponentsConfig.configureDistributionPropertiesTask.
+// extension.distribution.enabledVariants. See
+// AndroidComponentsConfig.configureDistributionPropertiesTask.
 private val strategies =
   listOf(
     AndroidOkHttpInstallStrategy.Registrar,


plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -390,6 +391,19 @@ private fun ApplicationVariant.configureDistributionPropertiesTask(
 ): TaskProvider<GenerateDistributionPropertiesTask>? {
   val variantName = name
   if (extension.distribution.enabledVariants.get().contains(variantName)) {
+    // Add the build distribution dependency if autoInstallation is enabled
```
chromy: I see all the machinery in `plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt` & `plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/*/*Strategy.kt` 

I assume we don't want to use that because we want to condition on `extension.distribution.enabledVariants`? It might be worth a comment either here and/or in `plugin-build/src/main/kotlin/io/sentry/android/gradle/autoinstall/AutoInstall.kt` explaining why distribution has a different implementation to the others

settings.gradle.kts
```diff
@@ -5,9 +5,7 @@ pluginManagement {
     google()
     maven {
       url = uri("https://storage.googleapis.com/r8-releases/raw")
-      content {
-        includeGroup("com.android.tools")
-      }
+      content { includeGroup("com.android.tools") }
```
chromy, line 8: You ran some linter/formatter over this and/or `sentry-java` right as a big one off change before right? We should try get whatever you used added to the CI.

runningcode: this is weird. i just ran spotlessApply and it did this.

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
runningcode: Yes that was my thought, it’s just easier to include it in here although it does make the task do two things instead of one. I’ll add a comment.

settings.gradle.kts
runningcode: I think have a theory: We run spotless on all the modules but perhaps we don’t call spotlessCheck on the root project. I will open a separate PR once I investigate further.

PR #1007, opened 2025-10-13
feat(distribution): Add dedicated distribution auth token property

runningcode:
## Summary

This PR adds a dedicated `authToken` property to `DistributionExtension` to decouple distribution authentication from the main org auth token. This provides better security separation by allowing distribution operations to use a separate token with potentially different permissions.

## Changes

- Add `authToken` property to `DistributionExtension` with `SENTRY_DISTRIBUTION_AUTH_TOKEN` environment variable as the default convention
- Update `GenerateDistributionPropertiesTask` to use `authToken` from the distribution extension
- Update all existing tests to use the new property
- Add new tests for `authToken` property behavior and environment variable convention

## Implementation Details

The implementation uses Gradle's convention mechanism to provide a clean fallback to the environment variable. The `authToken` property has `System.getenv("SENTRY_DISTRIBUTION_AUTH_TOKEN")` as its default convention, eliminating the need for explicit fallback logic in the task.

## Usage

Users can configure the distribution auth token in two ways:

### Via Gradle extension (overrides environment variable):
```kotlin
sentry {
  distribution {
    authToken.set("your-distribution-token")
  }
}
```

### Via environment variable (used by default):
```bash
export SENTRY_DISTRIBUTION_AUTH_TOKEN=your-distribution-token
```

#skip-changelog

### DIFF
--- a/plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTaskTest.kt
+++ b/plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTaskTest.kt
@@ -184,6 +184,65 @@ class GenerateDistributionPropertiesTaskTest {
   }
 
   @Test
+  fun `distribution authToken takes precedence over main authToken`() {
+    val project = createProject()
+    val extension = project.extensions.findByName("sentry") as SentryPluginExtension
+    extension.authToken.set("main-auth-token")
+    extension.distribution.authToken.set("distribution-auth-token")
+
+    val task: TaskProvider<GenerateDistributionPropertiesTask> =
+      GenerateDistributionPropertiesTask.register(
+        project,
+        extension,
+        null,
+        project.layout.buildDirectory.dir("dummy/folder/"),
+        "test",
+        "debug",
+      )
+
+    val outputDir = File(project.buildDir, "dummy/folder/")
+    outputDir.mkdirs()
+
+    task.get().generateProperties()
+
+    val expectedFile = File(project.buildDir, "dummy/folder/sentry-distribution.properties")
+    val props = PropertiesUtil.load(expectedFile)
+
+    // Distribution auth token should be used, not main auth token
+    assertEquals("distribution-auth-token", props.getProperty(DISTRIBUTION_AUTH_TOKEN_PROPERTY))
+  }
+
+  @Test
+  fun `does not fall back to main authToken when distribution authToken is not set`() {
+    val project = createProject()
+    val extension = project.extensions.findByName("sentry") as SentryPluginExtension
+    extension.authToken.set("main-auth-token")
+    // Explicitly set distribution auth token to null to override the env var convention
+    extension.distribution.authToken.set(null as String?)
+
+    val task: TaskProvider<GenerateDistributionPropertiesTask> =
+      GenerateDistributionPropertiesTask.register(
+        project,
+        extension,
+        null,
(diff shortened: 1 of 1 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
```diff
@@ -16,4 +17,8 @@ open class DistributionExtension @Inject constructor(objects: ObjectFactory) {
    */
   val enabledVariants: SetProperty<String> =
     objects.setProperty(String::class.java).convention(emptySet())
+
+  /** Auth token used for distribution operations. */
+  val distributionAuthToken: Property<String> =
```
runningcode: This is already inside the distribution extension so it is a bit redundant to say `distributionAuthToken`.
Should we just call it `authToken`? I wanted to make sure it is distinct from the `authToken` in the root of the extension..

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/DistributionExtension.kt
```diff
@@ -16,4 +17,8 @@ open class DistributionExtension @Inject constructor(objects: ObjectFactory) {
    */
   val enabledVariants: SetProperty<String> =
     objects.setProperty(String::class.java).convention(emptySet())
+
+  /** Auth token used for distribution operations. */
+  val authToken: Property<String> =
```
runningcode, line 22: There’s also `sentry.authToken` (this is `sentry.distribution.authToken`). This could be confusing since the names may shadow each other. Open to naming suggestions.

plugin-build/src/test/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTaskTest.kt
```diff
@@ -132,16 +132,17 @@ class GenerateDistributionPropertiesTaskTest {
...
+    assertEquals(envToken, props.getProperty(DISTRIBUTION_AUTH_TOKEN_PROPERTY))
   }
 
   @Test
   fun `extension properties override sentry properties file`() {
     val project = createProject()
     val extension = project.extensions.findByName("sentry") as SentryPluginExtension
     extension.org.set("extension-org")
-    extension.authToken.set("extension-token")
+    extension.distribution.authToken.set("extension-token")
```
chromy, line 145: Should we set both:
```
extension.authToken.set("extension-token")
extension.distribution.authToken.set("distribution-token")
```

And check `distribution-token` gets set?
Similarly may be worth having a test that if `extension.distribution.authToken` and `SENTRY_DISTRIBUTION_AUTH_TOKEN` aren't set we don't fall back to `extension.authToken` or SENTRY_AUTH_TOKEN

chromy: (assuming we don't have one already, I didn't check carefully)

runningcode: Good point. Added two tests.
1. use the `extension.distribution.authToken` if `extension.authToken` is present
2. don’t use `extension.authToken` when `extension.distribution.authToken` is not present.

chromy: Thanks!

PR #1008, opened 2025-10-15
feat(distribution): Include distribution properties for all enabled variants

runningcode:
## Summary
Theoretically, a customer might have a variant disabled for Sentry but they might want to use it for Distribution.

Example “debug” is “ignoredVariants” but they still want to use that variant for distribution.

The question to ask is:

> Why would someone want to distribute a build that has sentry disabled?

**If we think this situation will never happen** then please say no to this PR since I think it makes this code more complicated!!!

This PR ensures that distribution properties are included in the APK/bundle for all variants configured in `extension.distribution.enabledVariants`, regardless of whether they are "allowed" variants (filtered by `includeProguardMapping`, variant filters, etc.).

Previously, distribution properties were only included for variants that passed the `isVariantAllowed` check because the properties were added through the `InjectSentryMetaPropertiesIntoAssetsTask`, which only runs for allowed variants.

In addition, it writes the distribution properties to a separate file `sentry-distribution.properties` instead of `sentry-debug-meta.properties`.

See code for more comments.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -61,8 +61,17 @@ fun ApplicationAndroidComponentsExtension.configure(
   tmpDir.mkdirs()
 
   onVariants { variant ->
+    val paths = OutputPaths(project, variant.name)
+      variant.configureDistributionPropertiesTask(
+        project,
+        extension,
+        null,
+        paths,
+        sentryOrg,
+        sentryProject,
+      )
+
     if (isVariantAllowed(extension, variant.name, variant.flavorName, variant.buildType)) {
-      val paths = OutputPaths(project, variant.name)
       val sentryTelemetryProvider =
         variant.configureTelemetry(project, extension, cliExecutable, sentryOrg, buildEvents)
 
@@ -104,17 +113,6 @@ fun ApplicationAndroidComponentsExtension.configure(
           sentryProject,
         )
       generateProguardUuidTask?.let { tasksGeneratingProperties.add(it) }
-
-      val generateDistributionPropertiesTask =
-        variant.configureDistributionPropertiesTask(
-          project,
-          extension,
-          sentryTelemetryProvider,
-          paths,
-          sentryOrg,
-          sentryProject,
-        )
-      generateDistributionPropertiesTask?.let { tasksGeneratingProperties.add(it) }
 
       sentryVariant.configureNativeSymbolsTask(
         project,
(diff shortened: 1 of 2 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateDistributionPropertiesTask.kt
```diff
@@ -43,7 +43,7 @@ abstract class GenerateDistributionPropertiesTask : PropertiesFileOutputTask() {
 
   @get:Input abstract val projectSlug: Property<String>
 
-  @get:Input abstract val orgAuthToken: Property<String>
+  @get:Input abstract val authToken: Property<String>
```
runningcode, line 46: not exactly related to this PR but I went and changed this to authToken since in the future this will not be an `orgAuthToken` but a new purpose built token.

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -61,8 +61,17 @@ fun ApplicationAndroidComponentsExtension.configure(
   tmpDir.mkdirs()
 
   onVariants { variant ->
+    val paths = OutputPaths(project, variant.name)
+      variant.configureDistributionPropertiesTask(
+        project,
+        extension,
+        null,
```
runningcode, line 68: the `null` here is the `sentryTelemetryProvider`
Is that expensive to create even for variants that might not use sentry?

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -61,8 +61,17 @@ fun ApplicationAndroidComponentsExtension.configure(
   tmpDir.mkdirs()
 
   onVariants { variant ->
+    val paths = OutputPaths(project, variant.name)
+      variant.configureDistributionPropertiesTask(
```
runningcode, line 65: the main change here is moving this function outside of the `isVariantAllowed`

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -412,17 +410,25 @@ private fun ApplicationVariant.configureDistributionPropertiesTask(
...
+        sentryTelemetryProvider = sentryTelemetryProvider,
+        output = paths.distributionPropertiesDir,
+        taskSuffix = name.capitalized,
+        buildConfiguration = name,
+        sentryOrg = sentryOrg,
+        sentryProject = sentryProject,
+        variant = variant,
+      )
+
+    requireNotNull(sources.assets) { "sources.assets must not be null" }.addGeneratedSourceDirectory(
```
runningcode, line 426: This is now generating the properties files in to AGP’s `addGeneratedSourceDirectory` mechanism so the properties files are added to the build.

This is separate from the existing mechanism that uses transforms since we might not be running that if we don’t need to (for example: debug builds)

chromy (commented):
> If we think this situation will never happen then please say no to this PR since I think it makes this code more complicated!!!

In that case should we at least table this until someone asks for it and we can consider then?

PR #1047, opened 2026-01-07
fix(proguard): Do not generate empty debug-meta.properties file

### LINKED ISSUE #876: Possible to avoid changes to assets? (sentry-debug-meta.properties)
### Description

I'm using a tool to patch my app to avoid needing to create a new release for every change (Shorebird). However patches can not be made when assets have changed.

When using Sentry Android Gradle Plugin, the file `sentry-debug-meta.properties` is changed every time a new build is created which means I can't use shorebird to patch my app.

Is there a way around this?

### LINKED ISSUE #1045: No way to disable Sentry Gradle Plugin to generate sentry-debug-meta.properties from the apk (inside assets folder)
### Gradle Version

8.14.3

### AGP Version

8.13.2

### Code Minifier/Optimizer

R8

### Version

5.12.2

### Sentry SDK Version

8.29.0

### Steps to Reproduce

When I build app (release) with this config:
```kotlin
sentry {
...
      includeProguardMapping.set(false)
      autoUploadProguardMapping.set(false)
      uploadNativeSymbols.set(false)
      includeDependenciesReport.set(false)
      includeSourceContext.set(false)
      includeNativeSources.set(false)
      telemetry.set(false)
...
}
```
Still generate an empty sentry-debug-meta.properties file (with timestamp) inside apk (assets folder) and make my app can not reproduce build by F-droid

Image

### Expected Result

No sentry-debug-meta.properties generated or no timestamp inside sentry-debug-meta.properties

### Actual Result

Still generate an empty sentry-debug-meta.properties file (with timestamp) inside apk (assets folder)

romtsn:
## :scroll: Description

- Do not generate empty `sentry-debug-meta.properties` file if e.g. `includeProguardMapping` was set to false. Previously, we'd still generate it containing a single `io.sentry.build-tool=gradle` entry without any ids, but this wasn't very useful, so now we just don't include this file to assets at all (#1045)  
- Do not write timestamp to the properties file by using a custom `store` method. This timestamp has no use for us, and just invalidates build cache for some tools like shorebird for no reason (#876)

Closes #1045 
Closes #876 

Manually + automated

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix reproducible builds by writing `sentry-debug-meta.properties` without timestamps ([#876](https://github.com/getsentry/sentry-android-gradle-plugin/pull/876))
+- Skip generating `sentry-debug-meta.properties` when `includeProguardMapping` and `includeSourceContext` are disabled ([#1047](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1047))
+
 ## 6.0.0-rc.1
 
 ### Fixes


plugin-build/src/main/kotlin/io/sentry/android/gradle/util/PropertiesUtil.kt
```diff
@@ -18,5 +19,19 @@ class PropertiesUtil {
...
+
+    /**
+     * Stores properties to a writer without timestamps, ensuring reproducible builds.
+     *
+     * @param props The properties to store
+     * @param writer The writer to write to
+     * @param comment Optional comment to include at the top of the file (without the # prefix)
+     */
+    fun store(props: Properties, writer: Writer, comment: String? = null) {
+      comment?.let { writer.write("# $it\n") }
```
markushi, line 31: I'm not sure - should this be `\r\n` to properly work on windows? E.g. like this?
```suggestion
      comment?.let { writer.write("# $it\r\n") }
```

markushi (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/PropertiesUtil.kt
romtsn: As discussed, the behavior seems to match what's been there before: https://github.com/JetBrains/kotlin/blob/5ff3aafb5bd6b1664f95727cc54193435d9cb664/libraries/stdlib/src/kotlin/text/Appendable.kt#L75

PR #1050, opened 2026-01-08
fix(proguard): Fix task ordering for proguard UUID generation with configuration cache

romtsn:
## :scroll: Description

  Fix task ordering for proguard UUID generation with configuration cache.

  - Fix provider pattern in `AGP74Compat.kt` to preserve task dependencies from AGP's artifact API
  - Change `@Internal` to `@InputFiles` so Gradle properly orders the task after minification
  - Generate random UUID as fallback when mapping file doesn't exist
  - Only consider first existing mapping file (aligns with upload task behavior)

##  :bulb: Motivation and Context

Multiple customer reports internally. I wasn't able to reproduce this consistently, sadly.

  Race condition where SentryGenerateProguardUuidTask could run before the mapping file is created (observed with configuration cache + AGP 8.13.2 + Kotlin 2.3.0), resulting in a stale/mismatched UUID.

  Root cause:
  1. `@Internal` on proguardMappingFiles prevented Gradle from inferring task dependencies and properly serializing this as part of configuration cache
  2. Provider wrapping pattern (`project.provider { x.get() }`) broke AGP's artifact task-dependency chain

  - Existing unit and integration tests pass
  - `regenerates UUID if mapping file changes` and `does not regenerate UUID every build` tests cover the behavior

 ## :pencil: Checklist

  - [x] No breaking changes

##  :crystal_ball: Next steps

  N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fix reproducible builds by writing `sentry-debug-meta.properties` without timestamps ([#876](https://github.com/getsentry/sentry-android-gradle-plugin/pull/876))
 - Skip generating `sentry-debug-meta.properties` when `includeProguardMapping` and `includeSourceContext` are disabled ([#1047](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1047))
+- Fix task ordering for Proguard/R8 UUID generation with configuration cache ([#1050](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1050))
 
 ## 6.0.0-rc.1
 


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateProguardUuidTask.kt
runningcode, line 31: I wonder why this is internal.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateProguardUuidTask.kt
```diff
@@ -29,16 +32,22 @@ abstract class SentryGenerateProguardUuidTask : PropertiesFileOutputTask() {
...
 
   @TaskAction
   fun generateProperties() {
     val outputDir = output.get().asFile
     outputDir.mkdirs()
 
-    val proguardMappingFileHash =
-      proguardMappingFiles.files.joinToString { if (it.isFile) it.contentHash() else STATIC_HASH }
-    val uuid = UUID.nameUUIDFromBytes(proguardMappingFileHash.toByteArray())
+    // SentryUploadProguardMappingsTask also picks up the first existing mapping file, so there's
```
runningcode, line 44: Is the `proguardMappingFiles` in the same order for this task and that task?

runningcode (approved):
Just had some questions but otherwise looks good! Nice digging here!

PR #1054, opened 2026-01-21
fix(proguard): Fix compatibility with Kotlin/Compose 2.3.0

romtsn:
## :scroll: Description

Kotlin/Compose 2.3.0 introduced a new [task](https://github.com/JetBrains/kotlin/blob/b73fc4e8afb382976646fac728e717fd0b1d1c9c/libraries/tools/kotlin-compose-compiler/src/common/kotlin/org/jetbrains/kotlin/compose/compiler/gradle/internal/ComposeAgpMappingFile.kt#L84-L87) that modifies the mapping file (and also recalculates its SHA-hash) and this breaks our existing hooks. We have been hooking into the minify task before, but now this does not work anymore because the Kotlin task can run after/in-parallel with our task and modify the mapping file, so we'd be essentially looking at an empty mapping file location, or an outdated mapping file.

This PR introduces a new method to hook into the mapping file generation:
- On AGP 8.3+ it uses the new [toListenTo](https://github.com/android/gradle-recipes/tree/agp-8.3/listenToArtifacts) API so the task is automatically wired by AGP with the mapping file artifact
- On AGP < 8.3 it still uses the old way of hooking (`minifyTask.finalizedBy`) but in addition we also add a `mustRunAfter` to our task so it runs after the compose one.

Fixes https://github.com/getsentry/sentry-java/issues/5045

Manually + a new test + existing tests should cover too

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP74Compat.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP74Compat.kt
@@ -97,7 +97,7 @@ data class AndroidVariant74(private val variant: Variant) : SentryVariant {
       // as well to ensure the mapping file exists when our UUID task runs.
       val composeMappingMergeTask = getComposeMappingMergeTask(project, variantName)
       composeMappingMergeTask?.let { composeTask ->
-        task.configure { it.mustRunAfter(composeTask) }
+        task.configure { generateUuidTask -> generateUuidTask.mustRunAfter(composeTask) }
       }
     }
   }


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateProguardUuidTask.kt
```diff
@@ -29,16 +30,29 @@ abstract class SentryGenerateProguardUuidTask : PropertiesFileOutputTask() {
   override val outputFile: Provider<RegularFile>
     get() = output.file(SENTRY_UUID_OUTPUT)
 
-  @get:Internal abstract val proguardMappingFiles: ConfigurableFileCollection
+  // Used by AGP < 8.3 with conventional file paths
+  @get:Internal abstract val fallbackMappingFiles: ConfigurableFileCollection
```
runningcode, line 34: is it still correct that these are internal? i guess better to do this in a separate PR if we need to change this.

plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP83Compat.kt
```diff
@@ -0,0 +1,51 @@
...
+    dexguardEnabled: Boolean,
+  ) {
+    if (!dexguardEnabled) {
+      variant.artifacts
+        .use(task)
+        .wiredWith(SentryGenerateProguardUuidTask::mappingFile)
+        .toListenTo(SingleArtifact.OBFUSCATION_MAPPING_FILE)
+    } else {
+      // when dexguard is enabled we still want to go the old way, because AGP API does not apply
+      // there
```
runningcode:
```suggestion
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP74Compat.kt
```diff
@@ -77,6 +78,30 @@ data class AndroidVariant74(private val variant: Variant) : SentryVariant {
...
+
+      // we just hack ourselves into the Proguard/R8/DexGuard task's doLast.
+      minifyTask?.configure { it.finalizedBy(task) }
+
+      // When Kotlin Compose compiler is enabled, the final mapping file is written by
+      // mergeStagingReleaseComposeMapping task, not by R8 directly. We need to hook into that task
+      // as well to ensure the mapping file exists when our UUID task runs.
+      val composeMappingMergeTask = getComposeMappingMergeTask(project, variantName)
+      composeMappingMergeTask?.let { composeTask ->
+        task.configure { it.mustRunAfter(composeTask) }
```
runningcode: this was hard to follow but i think it makes sense and I don't see a better way to do it. nice finding!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateProguardUuidTask.kt
romtsn: Yeah, still needs to be internal for [this test](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1054/files#diff-4d537f61265c992b45593c58e615e79b289b26659f5b49a349d3a9156ef621edL968) to pass, otherwise it'll trigger minify task in robolectric tests unfortunately

plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP74Compat.kt
romtsn: yeah, maybe I can give better names to this tasks so it's not that confusing.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryGenerateProguardUuidTask.kt
runningcode: strange! is that true for 8.3+ as well?

romtsn: I actually haven't tried, but will give it a shot!

PR #1091, opened 2026-03-03
feat(snapshots): Add snapshot upload task

runningcode:
Context this is just the first version of the feature for BYO snapshots where customers already generate their own snapshots through another framework. This allows them to upload the snapshot folder directory to us.

## Summary
- Adds a new `sentryUploadSnapshots` task that invokes `sentry-cli build snapshots --app-id <APP_ID> <path>` to upload snapshots to Sentry
- Adds a `sentry.snapshots` extension block with `appId` and `path` properties
- Task is registered unconditionally using lazy property wiring

## Usage

In your app module's `build.gradle.kts`:

```kotlin
sentry {
  snapshots {
    appId.set("com.example.myapp")
    path.set(file("path/to/snapshots"))
  }
}
```

Then run:

```bash
./gradlew sentryUploadSnapshots
```

### DIFF
--- a/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPropertiesFileProviderTest.kt
+++ b/plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryPropertiesFileProviderTest.kt
@@ -178,6 +178,28 @@ class SentryPropertiesFileProviderTest(private val agpVersion: SemVer) {
   }
 
   @Test
+  fun `getPropertiesFilePath with null variant finds file inside project folder`() {
+    val (project, _) = createTestAndroidProject(forceEvaluate = !AgpVersions.isAGP74(agpVersion))
+    createTestFile(project.projectDir, "sentry.properties")
+
+    assertEquals("42", File(getPropertiesFilePath(project)!!).readText())
+  }
+
+  @Test
+  fun `getPropertiesFilePath with null variant skips buildType paths`() {
+    val rootProject = ProjectBuilder.builder().build()
+    val (project, _) =
+      createTestAndroidProject(
+        parent = rootProject,
+        forceEvaluate = !AgpVersions.isAGP74(agpVersion),
+      )
+    // Only place the file under src/debug/ — with null variant this should not be found
+    createTestFile(project.projectDir, "src${sep}debug${sep}sentry.properties")
+
+    assertEquals(null, getPropertiesFilePath(project))
+  }
+
+  @Test
   fun `getPropertiesFilePath finds file inside root buildType flavor folder`() {
     val rootProject = ProjectBuilder.builder().build()
     val (project, _) =


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -65,6 +66,24 @@ constructor(private val buildEvents: BuildEventListenerRegistryInternal) : Plugi
         sentryProjectParameter,
       )
 
+      project.afterEvaluate {
+        if (extension.snapshots.path.isPresent) {
+          SentryUploadSnapshotsTask.register(
+            project = project,
+            extension = extension,
+            sentryTelemetryProvider = null,
```
runningcode: neeed to fix this

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SnapshotsExtension.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.jetbrains.annotations.ApiStatus.Experimental
+
+@Experimental
+open class SnapshotsExtension @Inject constructor(objects: ObjectFactory) {
+
+  /** The application identifier used to associate snapshots with an app. */
+  val appId: Property<String> = objects.property(String::class.java)
```
runningcode, line 13: For BYO snapshots, we need to explicitly set the app id (otherwise we don't know what variant it is coming from).
For future versions where we automatically find the snapshots and previews, we can determine this automatically.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SnapshotsExtension.kt
```diff
@@ -0,0 +1,17 @@
...
+
+  /** The path to the folder containing snapshots to upload. */
+  val path: DirectoryProperty = objects.directoryProperty()
```
runningcode, line 16: thoughts on `path` as a name?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  companion object {
+    fun register(
+      project: Project,
+      extension: SentryPluginExtension,
+      sentryOrgOverride: String?,
+      sentryProjectOverride: String?,
+    ): TaskProvider<SentryUploadSnapshotsTask> {
+      val sentryTelemetryProvider = SentryTelemetryService.register(project)
```
runningcode: so this register the sentryTelemetryProvider unconditionally.
in the rest of the codebase it is only register onVariants when the feature is enabled but here we want to always register this task so that build authors can wire tasks to this task.

thoughts?

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SnapshotsExtension.kt
rbro112: Fine with me, I think it's pretty commonplace

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
rbro112: I personally have no idea what `SentryTelemetryService` so I'll defer to @romtsn or another plugin expert here

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SnapshotsExtension.kt
rbro112: So if I'm understand, we're not going to infer at this time? I think it would be nice to try to infer based on the module we're running this from, but that can be an add on

runningcode: The app id is variant dependent because you can add the application id suffix. For example `com.example` and `com.example.debug`.

So yes we could do this by registering one task per variant but then we would also need a way of setting this in the extension per variant. Both of which make it more confusing to use. Given that users will have to specify the folder in the extension, I think that specifying the app id isn't too big of an ask but open to discuss this!

romtsn: I guess we could start simple and then change it if we get feedback? It feels like a one-time setup, so probably not that tedious to maintain

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
romtsn: That's used to ingest telemetry into sentry about the task, e.g. this is for artifact uploads (size analysis, etc.): https://sentry.sentry.io/explore/traces/?aggregateField=%7B%22groupBy%22%3A%22%22%7D&aggregateField=%7B%22yAxes%22%3A%5B%22p95%28span.duration%29%22%5D%7D&mode=samples&project=4506241308295168&query=span.name%3A%EF%80%8DContains%EF%80%8DSentryUploadAppArtifactTask&statsPeriod=30d

if you're not interested in this, we can simply skip it for snapshots. If you want to have it, then we'd also need to call `start()` and just pass null for the SentryVariant param. I'd vote for having it in.

romtsn: as a general thought: we should probably revamp this at some point and add logs + metrics and track more things

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -65,6 +66,13 @@ constructor(private val buildEvents: BuildEventListenerRegistryInternal) : Plugi
         sentryProjectParameter,
       )
 
+      SentryUploadSnapshotsTask.register(
```
romtsn, line 69: I guess we could pass `cliExecutable` here as a param, we already retrieve it on line 50 above

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
```diff
@@ -0,0 +1,72 @@
...
+      return project.tasks.register(
+        "sentryUploadSnapshots",
+        SentryUploadSnapshotsTask::class.java,
+      ) { task ->
+        task.workingDir(project.rootDir)
+        task.debug.set(extension.debug)
+        task.cliExecutable.set(project.cliExecutableProvider())
```
romtsn: should we also set `task.sentryProperties` here, similar to what we do for the other cli tasks?

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SnapshotsExtension.kt
runningcode: yeah, that was my idea

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
runningcode: Yes, that method requires knowing the variant. but I'll add a new version that doesn't require the variant.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
runningcode: Good point

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
runningcode: ah i see. ok i did the dance to start the telemetry service just like it does for the other tasks but without a variant. let me know if that looks good.

runningcode: Oops, this seems like a real issue: https://github.com/getsentry/sentry-android-gradle-plugin/pull/1091#discussion_r2890197360

I don't see a way around that without reworking the telemtry service.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -65,6 +66,15 @@ constructor(private val buildEvents: BuildEventListenerRegistryInternal) : Plugi
...
+        project,
+        extension,
+        buildEvents,
+        cliExecutable,
+        sentryOrgParameter,
+        sentryProjectParameter,
+      )
```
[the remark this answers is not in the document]
runningcode: Yup that's the intentional design.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
runningcode: Chatted with Roman offline. we will remove the telemetry for this task and see if we can rework this later.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadSnapshotsTask.kt
runningcode: resolved!

PR #1107, opened 2026-03-13
ci: Publish snapshot builds to Maven Central

runningcode:
## Summary

- Adds a `mavenCentralSnapshots` Maven repository to both `plugin-build` and `sentry-kotlin-compiler-plugin` build files
- Adds a new GitHub Actions workflow (`.github/workflows/publish-snapshot.yml`) that publishes `-SNAPSHOT` versions of both plugins on every push to `main`
- Snapshot flow is completely independent from the existing craft-based release process

## Pre-requisite

Snapshots must be enabled for the `io.sentry` namespace in the [Sonatype Central Portal](https://central.sonatype.com/) under Namespaces (one-time manual step).

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -13,6 +13,7 @@ jobs:
   publish-snapshot:
     name: Publish snapshot to Maven Central
     runs-on: ubuntu-latest
+    environment: production
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6


.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,51 @@
...
+          ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.OSSRH_USERNAME }}
+          ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.OSSRH_PASSWORD }}
+
+      - name: Publish Kotlin Compiler Plugin snapshot
+        working-directory: sentry-kotlin-compiler-plugin
+        run: >
+          ../gradlew publishAllPublicationsToMavenCentralSnapshotsRepository
+          -PVERSION_NAME=${{ steps.version.outputs.snapshot_version }}
+        env:
+          ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.OSSRH_USERNAME }}
```
romtsn, line 51: TIL you can do that with env variables 👀

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,51 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  publish-snapshot:
+    name: Publish snapshot to Maven Central
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
romtsn: probably needs to be pinned (and setup-java too), following #1116

.github/workflows/publish-snapshot.yml
runningcode: yes to answer your question from a gradle side.

from the gihtub secrets perspective, these are just placeholders. i think unfortunately these variables aren't set. i will ask around tomorrow.

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,51 @@
...
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  publish-snapshot:
+    name: Publish snapshot to Maven Central
+    runs-on: ubuntu-latest
```
BYK, line 15:
```suggestion
    runs-on: ubuntu-latest
    environment: production
```

You need to put the secrets into a protected environment (I typically use the name `production` but it can be anything) and then add environment protection rules so they can only be used in protected branches to prevent random PRs or branches accessing these secrets.

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@0723195856401067f7a2779048b490ace7a47d7c # pin@v4
+
+      - name: Set up Java
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
```
BYK, line 25: FYI there's no real added security for pinning GitHub Actions' own actions to SHAs

runningcode: Thanks, I'm just following the pattern in the rest of the repo.

PR #1154, opened 2026-04-17
feat(snapshots): Add @SentrySnapshot runtime annotation (EME-1055)

runningcode:
## Summary

- Adds a new Maven artifact `io.sentry:sentry-snapshots-runtime` publishing the `@SentrySnapshot` annotation. Consumers apply it alongside `@Preview` to set a per-snapshot `diffThreshold`.
- Published as a new included build at `sentry-snapshots-runtime/`, parallel to `sentry-kotlin-compiler-plugin`, and registered in `.craft.yml`.
- Scoped to only the annotation artifact. A follow-up will wire the Gradle plugin's preview scanner (`PreviewMethodScanner`) to read it and surface `diffThreshold` in the sidecar JSON written by `GenerateSnapshotTestsTask`.

#skip-changelog

Resolves part 1 of [EME-1055](https://linear.app/getsentry/issue/EME-1055).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1158, opened 2026-04-21
feat(snapshots): Emit diff_threshold in snapshot sidecar (EME-1055)

runningcode:
## Summary

Finishes EME-1055 (follow-up to #1154) by wiring the `@SentrySnapshot(diffThreshold = ...)` runtime annotation into the per-snapshot sidecar JSON emitted by the generated Paparazzi test.

- The generated test's `writeSidecarMetadata` now reflects on `preview.declaringClass` / `preview.methodName` to read `@SentrySnapshot.diffThreshold`. The value is written as `diff_threshold` only when the annotation is present **and** the threshold is non-default (`!= 0f`), matching the existing "omit when default" pattern used for `font_scale`, `api_level`, etc.
- The reflective lookup uses `Class.forName` for the annotation class and is wrapped in `runCatching`, so a missing runtime jar or a class-loading failure silently skips the field rather than failing the test.

## Scope

- Touches only the runtime-generated sidecar. `ExportPreviewMetadataTask`'s build-time `preview-metadata.json` is intentionally left alone.
- No changes to `PreviewScanner` ASM machinery; detection is fully at test runtime.

#skip-changelog

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTask.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTask.kt
@@ -376,6 +376,17 @@ class $CLASS_NAME(
         if (info.showSystemUi) metadata["show_system_ui"] = true
         if (info.showBackground) metadata["show_background"] = true
 
+        val diffThreshold: Float? = runCatching {
+            val declaring = Class.forName(preview.declaringClass)
+            val method = declaring.declaredMethods.firstOrNull { it.name == preview.methodName }
+                ?: return@runCatching null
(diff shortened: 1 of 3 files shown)


romtsn (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTask.kt
```diff
@@ -376,6 +376,17 @@ class $CLASS_NAME(
         if (info.showSystemUi) metadata["show_system_ui"] = true
         if (info.showBackground) metadata["show_background"] = true
 
+        val diffThreshold: Float? = runCatching {
```
runningcode, line 379: This catches all errors here in scanning since we are using reflection.

PR #1159, opened 2026-04-22
feat(instrumentation): Add binder IPC call tracing

markushi:
## Summary

- Adds simple registry-based bytecode instrumentation that wraps known APIs causing Binder IPC calls (ContentResolver, PackageManager, Settings.*, ConnectivityManager, various system service managers, etc.) with `SentryIpcTracer.onCallStart` / `onCallEnd` calls so the SDK can emit spans/metrics for cross-process hops.
- The matching adapter class `io.sentry.android.core.internal.binder.SentryBinderAdapter` is added in getsentry/sentry-java#5326 and will ship in sentry-java `TBD`. Instrumentation is planned to no-op when `sentry-android-core` is below that version.
- New `tracingInstrumentation.binder.enabled` extension property, defaults to `true`.

## Test plan

- [x] `./gradlew :plugin-build:test --tests "io.sentry.android.gradle.instrumentation.binder.*"` (9 tests, all pass)
  - `BinderMethodRegistryTest` — registry lookup (unknown owner, unknown method, instance, static, Context subtype coverage)
  - `BinderIpcMethodVisitorTest` — ASM-level visitor test: wraps known instance + static calls with tracer, emits correct component/method LDC constants, adds try/catch handler, leaves unknown calls and opcode-mismatched calls untouched
- [x] Full unit test suite passes
- [x] `spotlessApply` clean

## Requires

- getsentry/sentry-java#5326 merged & released as 8.40.0


plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/BinderExtension.kt
```diff
@@ -0,0 +1,15 @@
+package io.sentry.android.gradle.extensions
+
+import javax.inject.Inject
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+
+open class BinderExtension @Inject constructor(objects: ObjectFactory) {
+  /**
+   * Enables or disables Binder IPC call instrumentation. Defaults to true. This requires
+   * sentry-android-core version TODO or above, and needs to be enabled. See
```
0xadam-brown, line 10: h: TODO (same below)

(though I'm sure they're on your radar)

0xadam-brown, line 10: l: Should we remove the "and needs to be enabled" since we enable it by default, or is that phrase referring to something else?

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/binder/Binder.kt
```diff
@@ -0,0 +1,32 @@
...
+      apiVersion,
+      originalVisitor,
+      CLASSNAME,
+      listOf(BinderMethodInstrumentable()),
+      parameters,
+    )
+  }
+
+  override fun isInstrumentable(data: ClassContext) = !data.isSentryClass()
+}
```
0xadam-brown, line 32: m: Any concerns about impact on build time? 

Our hash map lookups are fast, and I see we also perform the same scan for logging, etc., so the effect will be incremental for anyone who hasn't disabled the defaults. Not a blocker, just making sure we considered it + curious to hear how we think about these sorts of scans (eg, okay to add indefinitely b/c cost is marginal vs something we weigh carefully each time out).

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/binder/BinderMethodRegistry.kt
```diff
@@ -0,0 +1,2099 @@
+package io.sentry.android.gradle.instrumentation.binder
+
+data class BinderMethodSpec(
+  val owner: String,
+  val name: String,
```
0xadam-brown, line 5: l: name -> method[Name]

(I found myself making this translation as I read through the diff. Up to you, and ignore if ASM always uses `name`, of course.)

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/binder/BinderMethodRegistry.kt
```diff
@@ -0,0 +1,2099 @@
...
+    return specs.find { it.name == name }
+  }
+
+  @Suppress("LongMethod")
+  private fun buildRegistry(): Map<String, List<BinderMethodSpec>> {
+    val specs = mutableListOf<BinderMethodSpec>()
+
+    // region content & package management
+
+    specs.addAll(
```
0xadam-brown, line 25: Amazing work gathering all of these! 

Do we have a maintenance strategy going forward, by chance? Could be worth considering a Warden job, etc. at some point focused on hard-to-cover maintenance gaps.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/binder/BinderMethodVisitor.kt
```diff
@@ -0,0 +1,157 @@
...
+        opcode == Opcodes.INVOKESTATIC
+      } else {
+        opcode == Opcodes.INVOKEVIRTUAL || opcode == Opcodes.INVOKEINTERFACE
+      }
+    if (!opcodeMatchesSpec) {
+      super.visitMethodInsn(opcode, owner, name, descriptor, isInterface)
+      return
+    }
+
+    // We rewrite `target(args...)` into the equivalent of:
```
0xadam-brown, line 73: Thanks for the helpful documentation 💯

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/binder/BinderMethodVisitor.kt
```diff
@@ -0,0 +1,157 @@
...
+    // (push/invokeStatic) on purpose: those helpers route through visitMethodInsn - this very
+    // override - and would recurse. `super.visitMethodInsn` is reserved for the original target
+    // call so it still passes through any downstream visitors.
+
+    // The arguments (and the receiver, for instance calls) are already on the stack. Spill them
+    // into locals so we can run onCallStart first and reload them inside the try block. The stack
+    // is LIFO, so arguments are popped in reverse order.
+    val argTypes = Method(name, descriptor).argumentTypes
+    val argLocals = IntArray(argTypes.size)
+    for (i in argLocals.size - 1 downTo 0) {
```
0xadam-brown, line 94: super nit if you want it: argsLocal.size - 1 -> argsLocals.lastIndex

plugin-build/src/main/kotlin/io/sentry/android/gradle/services/SentryModulesService.kt
```diff
@@ -116,6 +120,8 @@ abstract class SentryModulesService :
     sentryModules.isAtLeast(SentryModules.SENTRY_ANDROID_CORE, SentryVersions.VERSION_APP_START) &&
       parameters.appStartEnabled.get()
 
+  fun isBinderInstrEnabled(): Boolean = parameters.binderEnabled.get()
```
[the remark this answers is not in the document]
0xadam-brown: +1

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/binder/BinderMethodRegistryTest.kt
```diff
@@ -0,0 +1,43 @@
...
+  @Test
+  fun `lookup returns null for known owner but unknown method`() {
+    assertNull(BinderMethodRegistry.lookup("android/content/ContentResolver", "unknownMethod"))
+  }
+
+  @Test
+  fun `lookup returns spec for known instance binder method`() {
+    val spec = BinderMethodRegistry.lookup("android/content/ContentResolver", "query")
+    assertNotNull(spec)
+    assertEquals("ContentResolver", spec.component)
```
0xadam-brown, line 26: l: Worth asserting on entire spec content? ie, also name and owner

(same below, esp covering subtypes separately, which would benefit most)

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/binder/BinderMethodVisitorTest.kt
```diff
@@ -0,0 +1,152 @@
...
+          false,
+        )
+        visitInsn(Opcodes.RETURN)
+      }
+
+    val instrumented = instrument(bytes)
+    val text = disassemble(instrumented)
+
+    assertFalse(text.contains("SentryBinderAdapter"))
+  }
```
0xadam-brown, line 151: l/m: Could be worth adding a test that actually loads the instrumented class to trigger bytecode verification (eg, no corrupted stack) and to verify that we can recover the binder method return value.

eg, create minimal stubs (a recording SentryBinderAdapter + a fake BluetoothAdapter.isEnabled() or whatever) -> instrument a synthetic class that calls BluetoothAdapter.isEnabled() and returns true -> load and invoke via reflection -> verify true is actually returned.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add Binder IPC call instrumentation ([#1159](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1159))
```
0xadam-brown, line 7: l: Do we normally also tell folks how to enable/disable (+ inform them that it's auto-enabled)?

0xadam-brown (commented):
Thanks @markushi – looks excellent! 🥇 

Just the missing SDK version gate + TODOs that are essential. A few additional comments on my end, but no other blockers.

PR #1186, opened 2026-05-12
feat(snapshots): Add global diffThreshold setting

runningcode:
## Summary
- Adds a `diffThreshold` property to `SnapshotsExtension` (default `0.0`) so users can set a global pixel-difference tolerance for all snapshots via the DSL
- The generated Paparazzi test uses this value as the tolerance instead of a hardcoded `0.0`
- Passed as `--diff-threshold` to the CLI upload command

```kotlin
sentry {
  snapshots {
    enabled.set(true)
    diffThreshold.set(0.05) // 5% tolerance for all snapshots
  }
}
```

EME-1133

### DIFF
--- a/plugin-build/src/test/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTaskTest.kt
+++ b/plugin-build/src/test/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTaskTest.kt
@@ -141,7 +141,7 @@ class GenerateSnapshotTestsTaskTest {
 
     assertTrue(
       content.contains(
-        "if (diffThreshold != null && diffThreshold != 0f) metadata[\"diff_threshold\"] = diffThreshold"
+        "if (effectiveThreshold != 0f) metadata[\"diff_threshold\"] = effectiveThreshold"
       )
     )
   }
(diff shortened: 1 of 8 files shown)


plugin-build/src/test/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTaskTest.kt
```diff
@@ -225,12 +225,28 @@ class GenerateSnapshotTestsTaskTest {
...
+
+    assertTrue(content.contains("val tolerance = 0.0"))
+  }
+
+  @Test
+  fun `generated file uses configured diffThreshold as tolerance`() {
+    val content = generateAndRead(packageTrees = listOf("com.example"), diffThreshold = 0.05)
+
+    assertTrue(content.contains("val tolerance = 0.05"))
+    assertFalse(content.contains("val tolerance = 0.0\n"))
```
markushi: Not sure if this needs to be checked at all, as the assert above should be good enough
```suggestion
```

markushi (approved):
LGTM!

plugin-build/src/test/kotlin/io/sentry/android/gradle/snapshot/GenerateSnapshotTestsTaskTest.kt
runningcode: Thank you! removed

PR #1199, opened 2026-05-18
fix(snapshot): Use string values for all tags

runningcode:
## Summary
- Changed the tags map type from `Map<String, Any>` to `Map<String, String>` in the generated snapshot sidecar
- Numeric tags (`font_scale`, `api_level`, `width_dp`, `height_dp`) now use `.toString()`
- Boolean tags (`show_system_ui`, `show_background`) now use `"true"` instead of `true`

Fixes EME-1156

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


cameroncooke (commented):
Should we just remove tags as they aren't supported by Snapshots to align with iOS?

runningcode:
They are implemented and validated as strings on the backend. The issue is that we are sending up `boolean`, `int`, etc.

NicoHinderling (approved):

PR #1201, opened 2026-05-19
fix(cli): Normalize Linux ARM64 arch name for sentry-cli binary lookup

### LINKED ISSUE #1168: "arm64" not normalized to "aarch64" when searching for sentry-cli binary
### Gradle Version

8.5.0

### AGP Version

-

### Code Minifier/Optimizer

None

### Version

4.0.0

### Sentry SDK Version

7.0.0

### Steps to Reproduce

On an ARM macbook:

```
$ docker build --platform linux/arm64 -f Dockerfile .
...
 > [redacted] ./gradlew build:
0.213 Downloading https://services.gradle.org/distributions/gradle-8.5-bin.zip                                                              
1.187 ............10%.............20%............30%.............40%.............50%............60%.............70%.............80%............90%.............100%                                                                                                                     
16.36                                                                                                                                       
16.36 Welcome to Gradle 8.5!
16.36 
16.36 Here are the highlights of this release:
16.36  - Support for running on Java 21
16.36  - Faster first use with Kotlin DSL
16.36  - Improved error and warning messages
16.36 
16.36 For more details see https://docs.gradle.org/8.5/release-notes.html
16.36 
16.36 Starting a Gradle Daemon (subsequent builds will be faster)
53.66 
53.66 > Configure project :
53.66 [sentry] Falling back to invoking `sentry-cli` from shell
61.46 
61.46 FAILURE: Build failed with an exception.
61.46 
61.46 * What went wrong:
61.46 A problem occurred starting process 'command 'sentry-cli''
61.46 > Could not start 'sentry-cli'
61.46 
61.46 * Try:
61.46 > Run with --stacktrace option to get the stack trace.
61.46 > Run with --info or --debug option to get more log output.
61.46 > Run with --scan to get full insights.
61.46 > Get more help at https://help.gradle.org.
61.46 
61.46 Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.
61.46 
61.46 You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
61.46 
61.46 For more on this, please refer to https://docs.gradle.org/8.5/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
61.46 
61.46 BUILD FAILED in 1m 1s
```

but for amd64 it works

```
$ docker build --platform linux/amd64 --no-cache -f Dockerfile .
...
[success]
```

### Expected Result

plugin should build on both linux AArch64 and amd64

### Actual Result

plugin doesn't find sentry-cli on AArch64.

Codex says it's because the arch name isn't normalized

> Because the plugin’s bundled sentry-cli lookup is architecture-name sensitive, and the arm64 path is weaker than the amd64 one.
> 
> In plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt:131, Linux does this:
> 
> - amd64 -> Linux-x86_64
> - anything else -> Linux-$osArch
> 
> The bundled binaries are downloaded as plugin-build/download-sentry-cli.sh:12:
> 
> - Linux-x86_64
> - Linux-aarch64
> 
> So on amd64, the code explicitly normalizes the JVM arch name and finds the bundled binary. On ARM, it does not normalize. If your Docker/JVM reports os.arch=arm64, the plugin looks for /bin/sentry-cli-Linux-arm64, but the artifact only contains sentry-cli-Linux-aarch64. That misses the resource and falls back to shell sentry-cli in plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt:50, which is why the container then complains that sentry-cli is not installed.
> 
> That also matches the tests: there’s coverage for amd64, but not aarch64/arm64 normalization in plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt:156.

and suggests this patch

```patch
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
@@ -132,8 +132,15 @@ internal object SentryCliProvider {
     // TODO: change to .lowercase(Locale.ROOT) when using Kotlin 1.6
     val osName = System.getProperty("os.name").toLowerCase(Locale.ROOT)
     val osArch = System.getProperty("os.arch")
+
     return when {
       "mac" in osName -> "Darwin-universal"
-      "linux" in osName -> if (osArch == "amd64") "Linux-x86_64" else "Linux-$osArch"
+      "linux" in osName -> {
+        val normalizedArch =
+          when (osArch) {
+            "amd64", "x86_64" -> "x86_64"
+            "arm64", "aarch64" -> "aarch64"
+            else -> osArch
+          }
+        "Linux-$normalizedArch"
+      }
       "win" in osName -> "Windows-i686.exe"
       else -> null
     }
```

romtsn:
## Summary

- Normalize Linux architecture names in `getCliSuffix()` so that both `arm64` and `aarch64` JVM values resolve to the bundled `sentry-cli-Linux-aarch64` binary
- Also normalize `x86_64` alongside the existing `amd64` mapping for robustness
- Add tests for `arm64`, `aarch64`, and `x86_64` arch values

Fixes #1168

## Test plan

- [x] Existing `SentryCliProviderTest` tests still pass
- [x] New test: `os.arch=arm64` -> `Linux-aarch64`
- [x] New test: `os.arch=aarch64` -> `Linux-aarch64`
- [x] New test: `os.arch=x86_64` -> `Linux-x86_64`
- [ ] Manually verify in Docker: `docker build --platform linux/arm64` no longer fails to find sentry-cli


plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -134,7 +134,17 @@ internal object SentryCliProvider {
     val osArch = System.getProperty("os.arch")
     return when {
       "mac" in osName -> "Darwin-universal"
```
runningcode: Unrelated to this PR, but since the universal binary is twice the size of the others, wouldn't it be better for performance if we used the platform specific binary on mac?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -134,7 +134,17 @@ internal object SentryCliProvider {
...
+          when (osArch) {
+            "amd64",
+            "x86_64" -> "x86_64"
+            "arm64",
+            "aarch64" -> "aarch64"
+            else -> osArch
+          }
+        "Linux-$normalizedArch"
+      }
       "win" in osName -> "Windows-i686.exe"
```
runningcode, line 152: we also now have windows aarch64 releases of the cli. should we fix that?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -134,7 +134,17 @@ internal object SentryCliProvider {
...
       "mac" in osName -> "Darwin-universal"
-      "linux" in osName -> if (osArch == "amd64") "Linux-x86_64" else "Linux-$osArch"
+      "linux" in osName -> {
+        val normalizedArch =
+          when (osArch) {
+            "amd64",
+            "x86_64" -> "x86_64"
+            "arm64",
+            "aarch64" -> "aarch64"
+            else -> osArch
```
runningcode, line 148: what about i686? Does that fall under else?

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -134,7 +134,17 @@ internal object SentryCliProvider {
     val osArch = System.getProperty("os.arch")
     return when {
       "mac" in osName -> "Darwin-universal"
-      "linux" in osName -> if (osArch == "amd64") "Linux-x86_64" else "Linux-$osArch"
+      "linux" in osName -> {
+        val normalizedArch =
+          when (osArch) {
+            "amd64",
+            "x86_64" -> "x86_64"
+            "arm64",
```
runningcode, line 146: isn't arm64 and aarch64 two separate binaries?

runningcode (commented):
Should we take a look at https://github.com/getsentry/sentry-cli/releases/tag/3.4.2 and make sure we support all the different releases correctly here?

runningcode (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
romtsn: Done in this PR - switched from Darwin-universal to Darwin-arm64/Darwin-x86_64. JAR size stays the same (~26MB total), but the extracted binary at runtime is ~half the size.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
romtsn: Windows i686 works on all Windows architectures via WoW64 (64-bit) Windows runs 32-bit executables natively), which is why there have been no complaints. Adding Windows x86_64/aarch64 would increase JAR size by ~23MB for a minor optimization. Could be a follow-up if there's demand.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
romtsn: Yes, i686 passes through the else branch and matches the bundled Linux-i686 binary directly. The i386 JVM variant wouldn't match, but 32-bit Linux is extremely rare for Android development

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
romtsn: No, they're two JVM names for the same architecture (ARMv8 64-bit). sentry-cli only publishes Linux-aarch64 binary. We normalize both to aarch64

PR #1222, opened 2026-05-22
fix: Remove AGP 7.x R8 workaround

runningcode:
## Summary
- Remove the conditional R8 classpath override that was only needed for AGP 7.x compatibility with compileSdk 34
- AGP 7.x is no longer in the CI matrix and the default AGP version is 8.10.1, making this dead code
- Removes the workaround from both `build.gradle.kts` and the test template in `BaseSentryPluginTest.kt`

#skip-changelog

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,4 @@
-buildscript {
-  dependencies { classpath("org.apache.commons:commons-compress:1.28.0") }
-}
+buildscript { dependencies { classpath("org.apache.commons:commons-compress:1.28.0") } }
 
 plugins {
   alias(libs.plugins.kotlin) version BuildPluginsVersion.KOTLIN apply false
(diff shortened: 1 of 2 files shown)


0xadam-brown (approved):
⚡

romtsn (approved):
I guess it's fine to not test compat with 7.x.x -- checked the usage and there's overall around ~500 AGP 7.x.x users (out of those who opted in for telemetry). We should drop it in the next major and it's fine to accept the risk of breaking something for the small number of users for now, imo (it's like 1.5% of overall).

Image

runningcode:
I'm seeing this in sentry-java but probably not related to this:

```WARNING: R8: An error occurred when parsing kotlin metadata. This normally happens when using a newer version of kotlin than the kotlin version released when this version of R8 was created. To find compatible kotlin versions, please see: https://developer.android.com/studio/build/kotlin-d8-r8-versions```

https://github.com/getsentry/sentry-java/actions/runs/28527031334/job/84566138543?pr=5670#step:9:508

runningcode:
Actually we don't run the release builds against AGP 7.X so this wouldn't affect customers.

PR #1231, opened 2026-05-26
ref(gradle): Replace File with DirectoryProperty in SentryCliProvider

runningcode:
## Summary
- Replace deprecated `project.buildDir` (`File`) and `project.rootDir` (`File`) usages with Gradle's lazy `DirectoryProperty` API (`project.layout.buildDirectory`, `project.layout.projectDirectory`) in `SentryCliProvider`, `SentryCliExecTask`, `SentryTelemetryService`, and `AndroidComponentsConfig`
- Update `SentryCliValueSource`, `SentryCliInfoValueSource`, and `SentryCliVersionValueSource` parameter interfaces to use `DirectoryProperty` instead of `Property<File>`
- Add `Project.getIsolatedRootProjectDir()` which uses the Gradle 8.8+ `isolated.rootProject.projectDirectory` API for isolated projects compatibility, with a fallback for older versions
- Fix all affected unit and integration tests

Relates to [GRADLE-70](https://linear.app/getsentry/issue/GRADLE-70)

#skip-changelog

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -13,7 +13,6 @@ import com.android.build.api.variant.HostTestBuilder.Companion.UNIT_TEST_TYPE
 import com.android.build.api.variant.Variant
 import com.android.build.gradle.internal.utils.setDisallowChanges
 import io.sentry.BuildConfig
-import io.sentry.android.gradle.SentryPlugin.Companion.sep
 import io.sentry.android.gradle.SentryPropertiesFileProvider.getPropertiesFilePath
 import io.sentry.android.gradle.SentryTasksProvider.capitalized
 import io.sentry.android.gradle.SentryTasksProvider.getAssembleTaskProvider
@@ -43,7 +42,6 @@ import io.sentry.android.gradle.util.SentryPluginUtils.isMinificationEnabled
 import io.sentry.android.gradle.util.SentryPluginUtils.isVariantAllowed
 import io.sentry.android.gradle.util.collectModules
 import io.sentry.android.gradle.util.hookWithAssembleTasks
-import java.io.File
 import org.gradle.api.Project
 import org.gradle.api.file.Directory
 import org.gradle.api.provider.Provider
@@ -61,7 +59,8 @@ fun ApplicationAndroidComponentsExtension.configure(
   sentryProject: String?,
 ) {
   // temp folder for sentry-related stuff
-  val tmpDir = project.layout.buildDirectory.dir("sentry-tmp").map { it.file("sentry") }.get().asFile
+  val tmpDir =
+    project.layout.buildDirectory.dir("sentry-tmp").map { it.file("sentry") }.get().asFile
   tmpDir.mkdirs()
 
   onVariants { variant ->
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -59,8 +59,7 @@ fun ApplicationAndroidComponentsExtension.configure(
   sentryProject: String?,
 ) {
   // temp folder for sentry-related stuff
-  val tmpDir =
-    project.layout.buildDirectory.dir("tmp").map { it.file("sentry") }.get().asFile
+  val tmpDir = project.layout.buildDirectory.dir("tmp").map { it.file("sentry") }.get().asFile
   tmpDir.mkdirs()
 
   onVariants { variant ->
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -60,7 +60,7 @@ fun ApplicationAndroidComponentsExtension.configure(
 ) {
   // temp folder for sentry-related stuff
   val tmpDir =
-    project.layout.buildDirectory.dir("sentry-tmp").map { it.file("sentry") }.get().asFile
+    project.layout.buildDirectory.dir("tmp").map { it.file("sentry") }.get().asFile
   tmpDir.mkdirs()
 
   onVariants { variant ->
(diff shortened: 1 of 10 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -61,7 +61,7 @@ fun ApplicationAndroidComponentsExtension.configure(
   sentryProject: String?,
 ) {
   // temp folder for sentry-related stuff
-  val tmpDir = File("${project.buildDir}${sep}tmp${sep}sentry")
+  val tmpDir = project.layout.buildDirectory.dir("sentry-tmp").map { it.file("sentry") }.get().asFile
```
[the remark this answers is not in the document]
0xadam-brown: **[1]**: m: Suspect we want to keep the existing directory. But if not, a CHANGELOG entry would warn anyone who might have tooling that depends on the directory location. 

**[2]**: l: If you want to simplify, I think `dir()` accepts composite paths: `project.layout.buildDirectory.dir("sentry-tmp/sentry").get().asFile`.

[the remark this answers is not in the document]
runningcode: 1. Yeah, we can keep the original if needed. I just didn't like the idea of a "tmp" folder in the root of the build directory. If another plugin also decides to call their outputs "tmp" then we will have hard to debug caching issues. Maybe it is better to have this debate in a separate PR so I'll undo it here and do that seperately.

2. Yes that is true but `/` is not a path separator on windows.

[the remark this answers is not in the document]
runningcode: Do you know more about what tooling would be built on this tmp dir?

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -421,7 +421,7 @@ class SentryMinimalException(message: String) : RuntimeException(message) {
```
0xadam-brown, line 421: m: Not in the diff, but looks like we're still using `project.buildDir` two times above (ll. 291, 318): 

```
cliVS.parameters.buildDirectory.set(project.buildDir)
```

(Sorry for missing it the first time round.) 

Worth a find + replace to make sure others haven't slipped through 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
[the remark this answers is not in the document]
0xadam-brown: > Do you know more about what tooling would be built on this tmp dir?

Nothing specific in mind – just a Hyrum's law thought. 

> Maybe it is better to have this debate in a separate PR so I'll undo it here and do that seperately.

Fwiw, Gradle's Task.temporaryDir resolves to `build/tmp/<name>` + it's where built-in tasks (Jar, etc.) put scratch files, so the original `tmp/sentry` was at least consistent with that. But no strong opinions.

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
runningcode: good catch. I'll have to do another PR to fix those. this pr was already difficult as it was to get it to pass.

runningcode: Fixed in: https://github.com/getsentry/sentry-android-gradle-plugin/pull/1235

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
```diff
@@ -179,8 +200,16 @@ fun Project.cliExecutableProvider(): Provider<String> {
...
-    it.parameters.projectDir.set(project.projectDir)
-    it.parameters.projectBuildDir.set(project.layout.buildDirectory.asFile.get())
-    it.parameters.rootProjDir.set(project.rootDir)
+    it.parameters.projectDir.set(layout.projectDirectory)
+    it.parameters.projectBuildDir.set(layout.buildDirectory)
+    it.parameters.rootProjDir.set(getIsolatedRootProjectDir())
+  }
+}
+
+fun Project.getIsolatedRootProjectDir(): Directory {
```
0xadam-brown: m: internal?

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -61,7 +59,7 @@ fun ApplicationAndroidComponentsExtension.configure(
   sentryProject: String?,
 ) {
   // temp folder for sentry-related stuff
-  val tmpDir = File("${project.buildDir}${sep}tmp${sep}sentry")
+  val tmpDir = project.layout.buildDirectory.dir("tmp").map { it.file("sentry") }.get().asFile
```
0xadam-brown, line 62: l: `it.dir("sentry")` would be more self-descriptive.

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
```diff
@@ -23,6 +24,11 @@ class SentryCliProviderTest {
 
   @get:Rule val systemPropertyRule = SystemPropertyRule()
 
+  private fun dirProp(dir: File): DirectoryProperty {
```
0xadam-brown, line 27: l: dirProp() spins up a new ProjectBuilder project per call. Reusing one project in the test class would be cheaper.

0xadam-brown (commented):
A couple quick (final) comments. The "l" ones are optional; the "m" is worth addressing if my assumption is right 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
runningcode: Yeah, that's what AI says, but i wanted to keep it as similar as possible for this PR. I've changed it to a `DirectoryProperty` and refactored all the code in this subsequent PR so this is moot: https://github.com/getsentry/sentry-android-gradle-plugin/pull/1236/changes

plugin-build/src/test/kotlin/io/sentry/android/gradle/SentryCliProviderTest.kt
runningcode: I think the tests are designed to be isolated so this is intentional but what do you mean by cheaper? Maybe I don't understand the cost.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryCliProvider.kt
runningcode: updated to private.

romtsn (approved):

PR #1232, opened 2026-05-26
fix(ci): Use JVM 17 for test tasks

romtsn:
_#skip-changelog_

This should address the AGP 9.3.0 matrix dimension failure

### DIFF
--- a/plugin-build/build.gradle.kts
+++ b/plugin-build/build.gradle.kts
@@ -70,8 +70,12 @@ java {
   targetCompatibility = JavaVersion.VERSION_11
 }
 
+// Tests need JVM 17 because the CI matrix tests against AGP 9.3+ (compiled with JVM 17 bytecode),
+// but production stays on JVM 11 to support hybrid/gaming SDKs that lag behind native tooling.
+val jvm17TestCompileTasks = setOf("compileTestJava", "compileTestKotlin")
+
 tasks.withType<JavaCompile>().configureEach {
-  if (name.contains("Test")) {
+  if (name in jvm17TestCompileTasks) {
     targetCompatibility = JavaVersion.VERSION_17.toString()
   }
 }
@@ -96,7 +100,7 @@ tasks.withType<KotlinCompile>().configureEach {
   }
 
   compilerOptions {
-    jvmTarget.set(if (name.contains("Test")) JVM_17 else JVM_11)
+    jvmTarget.set(if (name in jvm17TestCompileTasks) JVM_17 else JVM_11)
     // Kotlin supports current + 3 previous language versions.
     // We want 1.8, but if the compiler no longer supports it, use the oldest it does support.
     // e.g. Kotlin 2.1 oldest=1.8, Kotlin 2.3 oldest=2.0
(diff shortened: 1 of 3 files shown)


plugin-build/build.gradle.kts
```diff
@@ -89,7 +96,7 @@ tasks.withType<KotlinCompile>().configureEach {
   }
 
   compilerOptions {
-    jvmTarget.set(JVM_11)
+    jvmTarget.set(if (name.contains("Test")) JVM_17 else JVM_11)
```
runningcode: what happens if we set them all to JVM_17?

I ask because if we compile against JDK 17 for tests but compile against JDK 11 for production it could lead to subtle bugs that aren't covered by the tests. Probably unlikely to happen against the JDK but we should try to protect against it anyways.

romtsn: I guess our plugin could not be run on JVM 11 anymore? Probably it's fine for native android, but I'm worried of the hybrid and gaming customers, which tend to be always a bit behind the native tooling.

runningcode: Do you know if AGP 9.3.0 needs to be compiled against JVM 17 in order to run? if so, we're just pushing the problem until later.

runningcode (approved):

plugin-build/build.gradle.kts
```diff
@@ -69,6 +70,12 @@ java {
   targetCompatibility = JavaVersion.VERSION_11
 }
 
+tasks.withType<JavaCompile>().configureEach {
```
0xadam-brown, line 77: l: a quick comment explaining why we use different targets for prod vs test could be good for future us (or our favorite LLM).

plugin-build/build.gradle.kts
0xadam-brown: l: We could consolidate a bit if we care to (plus make the check slightly more robust): 

```
// Special case tests because CI/matrix runs on Java 17 (AGP 9.3+); keep production
// pegged to java 11 to support slow-to-update platforms (gaming, hybrid SDKs, etc.).
private val jvm17TestCompileTasks = setOf("compileTestJava", "compileTestKotlin")

...

// [Replace name.contains("Test") checks with the following]
if (name in jvm17TestCompileTasks) {
  sourceCompatibility = JavaVersion.VERSION_17.toString()
  targetCompatibility = JavaVersion.VERSION_17.toString()
}

...
jvmTarget.set(if (name in jvm17TestCompileTasks) JVM_17 else JVM_11)
```

Up to you of course.

0xadam-brown (approved):
Lgtm – a couple optional comments.

plugin-build/build.gradle.kts
romtsn: likely yes. I guess while we are not compiling against AGP 9.3.0 we should be good, though. I filed https://github.com/getsentry/sentry-android-gradle-plugin/issues/1234 (for the next major) to drop support for AGP 7.x.x and also bump the jvm compat version to 17.

romtsn: addressed in https://github.com/getsentry/sentry-android-gradle-plugin/pull/1232/commits/53262fa33f770236d57de26a98506c54cc5713e0

PR #1236, opened 2026-05-27
fix(plugin): Make instrumentation log dir variant-specific

runningcode:
## Summary
- The debug instrumentation log directory (`build/tmp/sentry`) was shared across all variants, which could cause log file corruption when AGP runs variant transforms in parallel
- Each variant now writes to its own directory: `build/sentry-logs/instrumentation/<variantName>`
- Removes the eager `mkdirs()` at configuration time — `CommonClassVisitor` already creates the directory at execution time
- Migrates `tmpDir` from `Property<File>` to `DirectoryProperty` using the Gradle layout API
- Updates integration tests to reference the new variant-specific path

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -58,10 +58,6 @@ fun ApplicationAndroidComponentsExtension.configure(
   sentryOrg: String?,
   sentryProject: String?,
 ) {
-  // temp folder for sentry-related stuff
-  val tmpDir = project.layout.buildDirectory.dir("tmp").map { it.file("sentry") }.get().asFile
-  tmpDir.mkdirs()
-
   onVariants { variant ->
     // Validate distribution configuration for this variant
     val updateSdkVariants = extension.distribution.updateSdkVariants.get()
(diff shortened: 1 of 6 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -234,7 +228,9 @@ fun ApplicationAndroidComponentsExtension.configure(
           params.appStartEnabled.setDisallowChanges(
             extension.tracingInstrumentation.appStart.enabled
           )
-          params.tmpDir.set(tmpDir)
+          params.tmpDir.set(
+            project.layout.buildDirectory.dir("sentry-logs/instrumentation/${variant.name}")
+          )
```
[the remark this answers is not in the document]
runningcode: This is an internal property, not user configurable so I don't see why this is needed (it also wasn't set before).

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -234,7 +228,9 @@ fun ApplicationAndroidComponentsExtension.configure(
           params.appStartEnabled.setDisallowChanges(
             extension.tracingInstrumentation.appStart.enabled
           )
-          params.tmpDir.set(tmpDir)
+          params.tmpDir.set(
+            project.layout.buildDirectory.dir("sentry-logs/instrumentation/${variant.name}")
```
romtsn, line 232: any specific reason we're going away from the `/tmp/sentry` pattern?

romtsn (approved):
one Q, but LGTM otherwise!

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
runningcode: Yeah, I think it is a bad practice since it isn't specific to our plugin. If someone else creates a `tmp` folder and uses it as input or output then we'll have changed the other tasks inputs or have overlapping outputs which would affect caching.

PR #1245, opened 2026-06-01
ref(build): Consolidate sentry-java version into version catalog

runningcode:
Remove the duplicate `sdk_version` property from `plugin-build/gradle.properties` and consolidate the sentry SDK version into a single declaration in `gradle/libs.versions.toml`.

I have a feeling that there is a bug here with included builds and `libs.versions.toml` files, so let's keep the custom script but it will update a single source of truth now.


gradle/libs.versions.toml
```diff
@@ -5,7 +5,7 @@ agp = "8.10.1"
 asm = "9.4" # // compatibility matrix -> https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/instrumentation/InstrumentationContext#apiversion
 ktfmt = "0.51"
 sqlite = "2.1.0"
-sentry = "7.0.0"
```
runningcode, line 8: i'm wondering why dependabot never updated this.

.github/workflows/update-deps.yml
```diff
@@ -25,15 +25,6 @@ jobs:
           name: CLI
           ssh-key: ${{ secrets.CI_DEPLOY_KEY }}
 
-  android:
```
runningcode, line 28: technically we don't need this if dependabot updates the libs.versions.toml.

scripts/update-android.sh
```diff
@@ -2,10 +2,10 @@
 set -euo pipefail
 
 cd $(dirname "$0")/../
-file='plugin-build/gradle.properties'
+file='gradle/libs.versions.toml'
```
runningcode, line 5: I have a feeling that there is a bug here with dependabot and included builds and `libs.versions.toml` files since this was never updated. So let's keep the custom updater script but it will update a single source of truth now.

plugin-build/build.gradle.kts
```diff
@@ -59,7 +59,7 @@ dependencies {
   testImplementationAar(libs.sqliteFramework)
   testRuntimeOnly(files(androidSdkPath))
   testImplementationAar(libs.sentryAndroid)
-  testImplementationAar(libs.sentryAndroidOkhttp)
+  testImplementation(libs.sentryOkhttp)
   testImplementationAar(libs.sentryOkhttp)
```
0xadam-brown: l: Seems we should be able to get rid of the AAR config now that we're depending on the JAR directly.

0xadam-brown (approved):
One comment, but doesn't have to block you.

gradle/libs.versions.toml
```diff
@@ -5,7 +5,7 @@ agp = "8.10.1"
...
+sentry = "8.43.0"
```
romtsn, line 8: potentially bumping this all the time would require us adapting SentryTelemetryService, but that's probably a good thing

gradle/libs.versions.toml
```diff
@@ -51,7 +51,6 @@ sqlite = { group = "androidx.sqlite", name = "sqlite", version.ref = "sqlite" }
 sqliteFramework = { group = "androidx.sqlite", name = "sqlite-framework", version.ref = "sqlite" }
 sentry = { group = "io.sentry", name = "sentry", version.ref = "sentry" }
 sentryAndroid = { group = "io.sentry", name = "sentry-android", version.ref = "sentry" }
-sentryAndroidOkhttp = { group = "io.sentry", name = "sentry-android-okhttp", version.ref = "sentry" }
```
romtsn, line 54: could we still keep it at a fixed version and drop that codepath (and the dep) in the next major of the plugin altogether?

romtsn (approved):
1 comment, but LGTM otherwise

runningcode:
Ok, updated to keep the older version for the tests

PR #1256, opened 2026-06-03
build(plugin): Add dependency locking and verification to `plugin-build` (GRADLE-104)

runningcode:
## Summary

Hardens the published plugin build against supply-chain risk by pinning its transitive dependencies, scoped to `plugin-build/` (the build that produces the shipped artifact). Part of [GRADLE-104](https://linear.app/getsentry/issue/GRADLE-104/acknowledge-dependency-pinning-audit-for-sentry-android-gradle-plugin) / SDK-1316.

Two complementary controls, both enforced automatically by existing CI (no workflow changes):

- **Gradle dependency locking** — `dependencyLocking { lockAllConfigurations() }` plus a `gradle.lockfile`. STRICT mode (the default) fails the build on unexpected version drift.
- **Gradle dependency verification** — `gradle/verification-metadata.xml` with a SHA-256 checksum for every resolved artifact (`verify-metadata=true`, `verify-signatures=false` — no PGP keyring to maintain).

The verification metadata captures the formatter (spotless), docs (dokka), and publish tooling in addition to the project configurations.

## Why the version-matrix tests are exempt

Pinning protects the **canonical build** (the default versions that ship). But [`test-matrix-agp-gradle.yaml`](.github/workflows/test-matrix-agp-gradle.yaml) deliberately tests compatibility against a *matrix* of AGP / Kotlin / Gradle versions, set via the `VERSION_AGP` / `VERSION_KOTLIN` env vars (read in `buildSrc/.../Dependencies.kt`) and the `gradle-version` override. Those overrides intentionally diverge from the lockfile and the recorded checksums, so pinning and the matrix are fundamentally in conflict:

- **Locking:** with `VERSION_KOTLIN` overridden, `compileClasspath` resolves e.g. `kotlin-stdlib:2.3.21`, but the lockfile pins `{strictly 2.0.0}` → `Dependency version enforced by Dependency Locking` failure.
- **Verification:** the matrix (and the standalone `-p plugin-build` integration test) resolve artifacts that aren't in `verification-metadata.xml` (different SDK/AGP versions; the buildscript-classpath `.module` files) → `Dependency verification failed`.

So we keep pinning on the canonical build and exempt the version-varying jobs:

- Locking is guarded to only apply when no `VERSION_AGP` / `VERSION_KOTLIN` override is present:
  ```kotlin
  val isVersionOverrideBuild =
    System.getenv("VERSION_AGP") != null || System.getenv("VERSION_KOTLIN") != null
  if (!isVersionOverrideBuild) {
    dependencyLocking { lockAllConfigurations() }
  }
  ```
- Verification is turned off on the version-matrix and standalone integration-test steps via `--dependency-verification=off`.

Canonical jobs (`pre-merge`, `publish`, the distribution build) keep **strict** locking *and* verification, so the protection that matters for the shipped artifact is fully enforced.

## Notes for reviewers

- After changing a dependency, regenerate the lockfile and verification metadata in one go:
  `./gradlew -p plugin-build resolveAndLockAll --write-locks --write-verification-metadata sha256`.
  This is now documented in [CONTRIBUTING.md](CONTRIBUTING.md), and AGENTS.md links to it.
- Local gotcha: version overrides only take effect with `--no-daemon`, because `System.getenv()` returns the *daemon's* startup environment (which is why CI, with a fresh daemon per job, behaves correctly).


plugin-build/build.gradle.kts
```diff
@@ -67,6 +67,18 @@ dependencies {
...
+
+tasks.register("resolveAndLockAll") {
+  notCompatibleWithConfigurationCache("Filters configurations at execution time")
+  doFirst {
+    require(gradle.startParameter.isWriteDependencyLocks) {
+      "$path must be run from the command line with the `--write-locks` flag"
+    }
+  }
+  doLast { configurations.filter { it.isCanBeResolved }.forEach { it.resolve() } }
+}
```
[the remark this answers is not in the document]
runningcode: oh great.. this is gonna be fun

.github/workflows/integration-tests-sentry-cli.yml
```diff
@@ -35,4 +35,6 @@ jobs:
       - name: Start server and run integration test for sentry-cli commands
         run: |
           test/integration-test-server-start.sh &
-          ./gradlew -p plugin-build integrationTest --tests "*IntegrationTest"
+          # Verification is disabled here because the standalone `-p plugin-build` build
+          # resolves buildscript-classpath artifacts not captured in the metadata.
+          ./gradlew -p plugin-build integrationTest --tests "*IntegrationTest" --dependency-verification=off
```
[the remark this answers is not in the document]
runningcode: Weird that sentry wants to disable the dependency verification. This is wrong.

plugin-build/build.gradle.kts
```diff
@@ -67,6 +67,26 @@ dependencies {
...
```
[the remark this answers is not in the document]
runningcode: Quite an edge case but seems valid but not worth fixing.

0xadam-brown (commented):
Nice! Coordinate + content locks look great. 

**m:** Two high-level suggestions:

**[1]** Could be nice to include a pre-merge workflow that warns folks if their lock file needs updating, ie, regenerate the lock file, diff against what's in the PR, and fail if changed.  

**[2]** Have dependabot PRs self-heal w/r/t the lock file. 

Eg, i) extract lock file generation into a script, ii) trigger a small workflow on `pull_request` from dependabot/* when gradle/**, plugin-build/**, or buildSrc/** change, and iii) re-run the script from (i). (Or whatever makes sense...)

[1] is optional, but I think we should do [2], lest our dependabot life become even busier. Either here or in a fast-follow is fine by me. Thoughts?

runningcode:
Yeah those are good points. 
For 1, I will update and add this
For 2, Dependabot [supports Gradle lock files](https://github.com/orgs/community/discussions/164000) but [not dependency verification](https://github.com/dependabot/dependabot-core/issues/1996).
I wonder if we should consider switching to Renovate since it supports lock files and dependency verification out of the box and is free on open source repos: https://docs.renovatebot.com/modules/manager/gradle/
Or otherwise drop the dependency verification from this PR.

.github/workflows/update-dependency-locks.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: dependabot/fetch-metadata@25dd0e34f4fe68f24cc83900b1fe3fe149efef98 # v3.1.0
+
+      - name: Check out PR branch
+        if: steps.metadata.outputs.package-ecosystem == 'gradle'
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          ref: ${{ github.head_ref }}
+          # Pushing with the deploy key (rather than GITHUB_TOKEN) re-triggers CI
+          # so the regenerated files are validated.
+          ssh-key: ${{ secrets.CI_DEPLOY_KEY }}
```
runningcode: I think we need to add this secret.

.github/workflows/update-dependency-locks.yml
```diff
@@ -0,0 +1,65 @@
+name: Update Dependency Locks
```
runningcode: This is generated by claude and not tested. We'll just need to merge this and adjust as needed since there isn't really a way to testing the full integration and secrets.

0xadam-brown: 👍 on merging and iterating. 

l: Are you planning the pre-merge staleness check as a follow-up PR, or do we want to include it in this one? (Assuming you still think it's worthwhile.)

.github/workflows/update-dependency-locks.yml
```diff
@@ -0,0 +1,68 @@
...
+
+# Dependabot bumps versions in gradle/libs.versions.toml but cannot regenerate
+# the plugin-build lockfile or verification metadata (it has no support for the
+# latter, and its lockfile support breaks with version catalogs). Without this,
+# every Dependabot Gradle PR fails STRICT-mode locking and checksum verification.
+# This job regenerates both and pushes them back onto the Dependabot branch.
+
+on:
+  pull_request:
```
0xadam-brown: l: Worth adding a concurrency group? 

If dependabot rebases while regeneration is in-flight, two runs could race.

```
concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref }}
   cancel-in-progress: true
```

0xadam-brown (approved):
Thanks for the updates 💯 

A few non-blockers if you want them.

CONTRIBUTING.md
```diff
@@ -7,6 +7,26 @@ We suggest opening an issue to discuss bigger changes before investing on a big
...
+dependency graph for supply-chain hardening: resolved versions are recorded
+in `plugin-build/gradle.lockfile` and a SHA-256 checksum for every artifact
+in `plugin-build/gradle/verification-metadata.xml`. Locking runs in STRICT
+mode, so any drift fails the build.
+
+Whenever you add, remove, or bump a dependency in
+`plugin-build/build.gradle.kts`, regenerate both files and commit them:
+
+```bash
+./gradlew -p plugin-build resolveAndLockAll --write-locks --write-verification-metadata sha256
```
romtsn, line 22: shall we update `AGENTS.md` to mention this as well?

romtsn (approved):
LGTM, good stuff! If you want to take the renovate path, I'd already run it through legal a while ago, and it should be enabled for this repo. I don't think there's a requirement to use dependabot strictly.

CONTRIBUTING.md
runningcode: done!

.github/workflows/update-dependency-locks.yml
runningcode: this is a good point but moot now that we have renovate.

.github/workflows/update-dependency-locks.yml
runningcode: now we have renovate so I removed this file!

PR #1264, opened 2026-06-04
fix(config-cache): Resolve CLI path via ValueSource task input (GRADLE-70)

runningcode:
## Summary

The sentry-cli path was memoized in a `@Volatile` static field on `SentryCliProvider`. That field survives across builds within the Gradle daemon and is **not** invalidated when switching branches, so a stale path pointing at a sentry-cli temp file extracted under a previous checkout could be reused — producing `Could not start '…/.sentry-cliXXXX.exe' … No such file or directory` failures when the configuration cache is enabled.

This removes the static memoization and stops threading the cli path (`cliExecutable: Provider<String>`) through the `configure()` call chain into every task. Each `SentryCliExecTask` now resolves the path itself in `asSentryCliExec()` via the existing `cliExecutableProvider()` `ValueSource` (which Gradle correctly invalidates on branch switches), and wires its own `buildDirectory` task input.

This is one of two PRs splitting the config-cache work; the companion PR #1263 (GRADLE-82) makes the telemetry org lookup config-cache friendly. This PR keeps the `cliExecutable` parameter on `configure()` because telemetry still needs it on `main`; whichever PR lands second will drop that now-unused parameter in a small rebase.

Fixes GRADLE-70


0xadam-brown (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fixes
 
 - Compose tracing no longer adds the Sentry modifier multiple times for chained modifiers (e.g. `Modifier.fillMaxSize().padding()`) on Kotlin 2.2 and newer ([#1253](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1253))
+- Resolve the sentry-cli path as a task input instead of memoizing it in a static field, fixing stale-path build failures when switching branches with the configuration cache enabled ([#1264](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1264))
```
romtsn: need to rebase and move this to Unreleased again, since there has been a new version released in the meantime

```suggestion
- Resolve the sentry-cli path as a task input instead of memoizing it in a static field, fixing stale-path build failures when switching branches with the configuration cache enabled ([#1264](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1264))
  - This fixed the issue where sentry-cli could not be found (`A problem occurred starting process 'command  ../sentry-cliXXX.exe'`)
```
maybe add this for better SEO and discoverability?

romtsn (approved):
great impact!

CHANGELOG.md
runningcode: Good idea, added it

PR #1265, opened 2026-06-04
build(deps): Upgrade default Kotlin to 2.1.21

runningcode:
Bumps the default Kotlin version in `buildSrc/src/main/java/Dependencies.kt` from `1.8.20` to the latest `2.1.x` release (`2.1.21`).

Staying on the 2.1 line is deliberate: `plugin-build` derives its `languageVersion`/`apiVersion` from the oldest language version the compiler still supports (current − 3, floored at 1.8). That resolves to **1.8 for Kotlin 2.1**, but would jump to **2.0 for Kotlin 2.3** — so 2.1.x keeps the plugin's Kotlin 1.8 language-level support intact.

The KSP version is now derived from the resolved `KOTLIN` value instead of only the `VERSION_KOTLIN` env var. This keeps the new 2.x default consistent with KSP2, while a 1.x override (e.g. from the compat test matrix) still selects KSP1.

### DIFF
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -1,14 +1,17 @@
 object BuildPluginsVersion {
     val AGP = System.getenv("VERSION_AGP") ?: "8.10.1"
-    val KOTLIN = System.getenv("VERSION_KOTLIN") ?: "1.8.20"
+    // Bumping this may implicitly raise the language level: plugin-build derives its
+    // languageVersion/apiVersion from the oldest the compiler still supports (current minus
+    // three, floored at 1.8), so e.g. Kotlin 2.1 keeps 1.8 but Kotlin 2.3 forces 2.0.
+    val KOTLIN = System.getenv("VERSION_KOTLIN") ?: "2.1.21"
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/Dependencies.kt
```diff
@@ -1,14 +1,14 @@
 object BuildPluginsVersion {
     val AGP = System.getenv("VERSION_AGP") ?: "8.10.1"
-    val KOTLIN = System.getenv("VERSION_KOTLIN") ?: "1.8.20"
+    val KOTLIN = System.getenv("VERSION_KOTLIN") ?: "2.3.21"
```
0xadam-brown: l: Is it worth adding a compat-matrix row with VERSION_KOTLIN=1.x  to make sure CI keeps exercising the KSP1 fallback branch? Or worth removing 1.x support altogether?

One or the other seems coherent, but up to you. (The compat-matrix had dropped 1.x before this PR, meaning we were in a weird state in that we were dogfooding on a stale version.)

0xadam-brown (approved):

romtsn (approved):

PR #1285, opened 2026-06-09
feat(plugin): Auto-instrument SQLiteDriver for Room users (GRADLE-107)

### LINKED ISSUE #1281: Auto-instrument SQLiteDriver for Room users
### Problem Statement

Room 2.7 introduced `RoomDatabase.Builder.setDriver(SQLiteDriver)` as the new way to configure an underlying SQLite implementation ([link](https://developer.android.com/jetpack/androidx/releases/room#2.7.0)), with the intention of eventually replacing `SupportSQLiteOpenHelper`. (Room 3.0+ does just that.) 

The Sentry Android Gradle Plugin currently auto-instruments only the open helper path. Apps that adopt the driver-based API will lose automatic SQL span creation unless they manually wrap their driver with `SentrySQLiteDriver.create(...)` at every `setDriver(...)` call site.                        

#### Proposed approach

Define a new ASM ClassInstrumentable + MethodVisitor pair for wrapping the parameter passed to `Room.DatabaseBuilder.setDriver(SQLiteDriver)`. The wrapper will be the `SentrySQLiteDriver` introduced under [JAVA-275](https://linear.app/getsentry/issue/JAVA-275/support-for-androidxsqlites-sqlitedriver) / [#5001](https://github.com/getsentry/sentry-java/issues/5001).

In other words, this:

```kotlin
val database = Room.databaseBuilder(context, MyDatabase::class.java, "dbName")
     .setDriver(AndroidSQLiteDriver())
     .build()
```

becomes:

```kotlin
val database = Room.databaseBuilder(context, MyDatabase::class.java, "dbName")
     .setDriver(SentrySQLiteDriver.create(AndroidSQLiteDriver()))
     .build()
```

The `SentrySQLiteDriver` implementation will need to protect against duplicate wrapping, should the driver passed to setDriver() already be wrapped.

### Coverage

**tl;dr**: The proposed approach would cover all Room users; SQLDelight users don’t need to be covered (they still use SupportSQLiteOpenHelper); and the few developers who use SQLiteDriver directly would need to wrap manually.

#### 1. Persistence libraries

The vast majority of developers access SQLite via a persistence library, and that library is almost always Room or SQLDelight.

`Room.DatabaseBuilder.setDriver()` is the sole Room 2.7+ and 3.0+ entry point for SQLiteDriver – at least for now. We'll want to monitor the API as it evolves. (Room first supported SQLiteDriver in 2.7.0-alpha01.) 

SQLDelight doesn’t officially support SQLiteDriver; it uses SupportSQLiteOpenHelper instead ([link](https://github.com/sqldelight/sqldelight/issues/6072)). But there's a third-party bridge library ([link](https://github.com/eygraber/sqldelight-androidx-driver)). Use of the bridge library isn’t widespread (90 stars, 6 forks), so we don't plan to support its creation pattern ([link](https://github.com/eygraber/sqldelight-androidx-driver#create-the-driver-and-database)). We can revisit if there's sufficient community demand.

#### 2. Direct SQLiteDriver use

Under the proposed approach, the few developers whose use SQLiteDriver directly would need to wrap it manually. Eg, `AndroidSqliteDriver().open()` -> `SentrySQLiteDriver.create(AndroidSqliteDriver()).open()`.

0xadam-brown:
## :scroll: Description

Adds a new bytecode method visitor that lets us auto-wrap all occurrences of `SQLiteDriver` with `SentrySQLiteDriver` whenever the driver is passed to `Room.DatabaseBuilder.setDriver(SQLiteDriver)` (Room's sole driver access point – see the "Coverage" section from https://github.com/getsentry/sentry-android-gradle-plugin/issues/1281 for why we don't need to worry about anything else).

In other words, this:

```kotlin
val database = Room.databaseBuilder(context, MyDatabase::class.java, "dbName")
     .setDriver(AndroidSQLiteDriver())
     .build()
```

becomes:

```kotlin
val database = Room.databaseBuilder(context, MyDatabase::class.java, "dbName")
     .setDriver(SentrySQLiteDriver.create(AndroidSQLiteDriver()))
     .build()
```

The wrapping policy is naive in that every SQLiteDriver passed to Room.DatabaseBuilder.setDriver() is wrapped. That's deliberate because SentrySQLiteDriver protects against double-wrapping internally, which lets us keep our visitor implementation simple.

Room 2.7 introduced `RoomDatabase.Builder.setDriver(SQLiteDriver)` as the new way to configure an underlying SQLite implementation ([link](https://developer.android.com/jetpack/androidx/releases/room#2.7.0)), with the intention of eventually replacing `SupportSQLiteOpenHelper`. (Room 3.0+ [does just that](https://developer.android.com/jetpack/androidx/releases/room3#3.0.0-alpha01).)

The Sentry Android Gradle Plugin currently auto-instruments only the open helper path. Apps that adopt the driver-based API will lose automatic SQL performance instrumentation unless they manually wrap their driver with `SentrySQLiteDriver.create(...)` at every `setDriver(...)` call site.

This PR provides auto-wrapping for the driver path too.

Addresses [GRADLE-107](https://linear.app/getsentry/issue/GRADLE-107/auto-wrap-sqlitedriver-with-sentrysqlitedriver-for-room-users) / #1281 

## Preconditions

1. sentry-android-sqlite version in the SAGP build must include SentrySQLiteDriver
2. InstrumentationFeature.DATABASE must be enabled (⚠️ in other words, we piggyback on the existing DATABASE toggle used to auto-instrument the SupportSQLiteOpenHelper + Room DAO spans)

## Coverage

- Auto-wraps SQLiteDriver for all Room users.
- SQLDelight users don't need driver auto-wrapping (they still use SupportSQLiteOpenHelper, which we already auto-instrument).
- The few developers who use SQLiteDriver directly will need to wrap it manually.

## Impact on build times

Near zero. The new instrumentation gates on a two-entry allowlist (androidx.room.RoomDatabase$Builder / androidx.room3.RoomDatabase$Builder), which means 0-1 classes will be visited per build in virtually all scenarios. The rest of the per-class cost is a single Set.contains check.

### SupportSQLiteOpenHelper

Driver instrumentation operates alongside our existing instrumentation of SupportSQLiteOpenHelper. (It's fine to use both drivers and open helpers in the same project, so long as a given db file is associated with exactly one of them. Room's APIs help enforce that restriction.) 

The Sentry Android SDK – and not the SAGP – is responsible for protecting against double-wrapping and duplicate spans. In general that's easy (just check the type being passed to the wrapper). In the case of the SupportSQLiteDriver used with Room [2.7, 3.0) things get a bit trickier, as it's both a driver and consumes an open helper. But the SDK has that covered as well (see [#5514](https://github.com/getsentry/sentry-java/pull/5514)).

### Room DAO methods

Driver instrumentation also operates in tandem with our existing Room DAO spans – in theory, at least. Unfortunately DAO instrumentation broke under API changes in Room 2.7, which is the same version that introduced SQLiteDriver. So for now the two are never active at the same time. Fixing that is a task for another day...

**[1]** Added unit tests + integration tests in this PR

**[2]** Verified that all the (for now) `@Ignored` integration tests pass against a local sentry-java SDK artifact containing SentrySQLiteDriver.

**[3]** Built out a local version of the Sentry Android sample app that let me manually test SAGP + sentry-java compatibility end-to-end. Verified that auto-instrumentation of `SQLiteDriver` works as expected. 

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

- Merge updates to sentry-docs ([#18489](https://github.com/getsentry/sentry-docs/pull/18489))
- Consider a CI cron job that checks Room's latest API for changes that would impact our implementation here and in the Sentry Android SDK.


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/AndroidXSQLiteDriver.kt
```diff
@@ -0,0 +1,54 @@
...
+      className = currentClassName.substringAfterLast('.'),
+      methodInstrumentables = METHOD_INSTRUMENTABLES,
+      parameters = parameters,
+    )
+  }
+
+  // Broad activation: the setDriver call site can appear in any class. The MethodInstrumentable
+  // and MethodVisitor filter at the call-site level. Per-method overhead is one MethodVisitor
+  // construction; per-invoke overhead is one visitMethodInsn predicate evaluation.
+  override fun isInstrumentable(data: ClassContext): Boolean = !data.isSentryClass()
```
0xadam-brown: ⚠️ Checking every non-Sentry class for instrumentability is the most controversial implementation decision in this PR. 

Some existing visitors do likewise, including logging and remapping – and logging is enabled for users by default. That makes the cost here incremental for most (db instrumentation is also enabled by default). But users who've disabled logging and enabled db instrumentation will be especially impacted.

I've run benchmarks against the DuckDuckGo app (see the PR description for details). They suggest that the impact is within reason (~1.5-2.5s extra for warm builds). We can get smarter if we hear differently, but the lift is non-trivial, so I've opted for the simple approach for now.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Features
+
+- Auto-wrap SQLiteDriver with SentrySQLiteDriver for Room users ([#1281](https://github.com/getsentry/sentry-android-gradle-plugin/issues/1281))
+  - Gated on `sentry-android-sqlite` >= 8.44.0 and the existing `tracingInstrumentation` `DATABASE` feature
```
0xadam-brown: **TODO**: Update version number once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -529,6 +530,68 @@ class SentryPluginTest :
     }
   }
 
+  @Ignore("Placeholder version VERSION_SQLITE_DRIVER not yet on Maven")
```
0xadam-brown: **TODO**: Remove `@Ignore` once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -529,6 +530,68 @@ class SentryPluginTest :
...
+    assertTrue {
+      "[sentry] Instrumentable: ChainedInstrumentable(instrumentables=" +
+        "AndroidXSQLiteOpenHelper, AndroidXSQLiteDriver, AndroidXRoomDao)" in build.output
+    }
+  }
+
+  // The open helper wrapper (>= 6.21.0) and driver wrapper (>= VERSION_SQLITE_DRIVER) are
+  // independent gates and can be used simultaneously; both instrumentables must appear when the
+  // newer SDK is on the classpath.
+  @Ignore("Placeholder version VERSION_SQLITE_DRIVER not yet on Maven")
```
0xadam-brown: **TODO**: Remove `@Ignore` once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/Versions.kt
```diff
@@ -33,6 +33,7 @@ internal object SentryVersions {
   internal val VERSION_LOGCAT = SemVer(6, 17, 0)
   internal val VERSION_APP_START = SemVer(7, 1, 0)
   internal val VERSION_SQLITE = SemVer(6, 21, 0)
+  internal val VERSION_SQLITE_DRIVER = SemVer(8, 44, 0)
```
0xadam-brown: **TODO**: Update version number once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -529,6 +531,87 @@ class SentryPluginTest :
     }
   }
 
+  @Ignore(SQLITE_DRIVER_IGNORE_REASON)
```
0xadam-brown: **TODO**: Remove `@Ignore` once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -529,6 +531,87 @@ class SentryPluginTest :
...
+
+    assertInstrumentableChain(
+      build,
+      "AndroidXSQLiteOpenHelper",
+      "AndroidXSQLiteDriver",
+      "AndroidXRoomDao",
+    )
+  }
+
+  @Ignore(SQLITE_DRIVER_IGNORE_REASON)
```
0xadam-brown: **TODO**: Remove `@Ignore` once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -529,6 +531,87 @@ class SentryPluginTest :
...
+
+  @Test
+  fun `does not apply sqliteDriver instrumentable when sentry gate fails with room3`() {
+    val build =
+      buildDatabaseInstrumentation(SQLITE, ROOM3_AT_DRIVER_FLOOR, SENTRY_ANDROID_SQLITE_OPEN_HELPER)
+
+    assertInstrumentableChain(build, "AndroidXSQLiteOpenHelper", "AndroidXRoomDao")
+  }
+
+  @Ignore(SQLITE_DRIVER_IGNORE_REASON)
```
0xadam-brown: **TODO**: Remove `@Ignore` once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryPluginTest.kt
```diff
@@ -529,6 +531,87 @@ class SentryPluginTest :
...
+
+    assertInstrumentableChain(
+      build,
+      "AndroidXSQLiteOpenHelper",
+      "AndroidXSQLiteDriver",
+      "AndroidXRoomDao",
+    )
+  }
+
+  @Ignore(SQLITE_DRIVER_IGNORE_REASON)
```
0xadam-brown: **TODO**: Remove `@Ignore` once we ship the SentrySQLiteDriver in sentry-android.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/AndroidXSQLiteDriver.kt
0xadam-brown: No longer relevant. I've updated the approach so that we no longer check all non-Sentry classes. Instead of checking for all occurrences of `setDriver(SQLiteDriver)`, we now check for `Room.DatabaseBuilder.setDriver(SQLiteDriver)`, making incremental build time trivial. 

(I realized we could do the latter after thinking more carefully through the affected APIs. We don't lose anything in actual coverage under the new approach.)

gradle/libs.versions.toml
```diff
@@ -60,8 +60,11 @@ sentryAndroidOkhttp = { group = "io.sentry", name = "sentry-android-okhttp", ver
 sentrySpringBootJakarta = { group = "io.sentry", name = "sentry-spring-boot-starter-jakarta", version.ref = "sentry" }
 
 # test
-mockitoKotlin = { group = "org.mockito.kotlin", name = "mockito-kotlin", version = "5.4.0" }
 arscLib = { group = "io.github.reandroid", name = "ARSCLib", version = "1.1.4" }
+mockitoKotlin = { group = "org.mockito.kotlin", name = "mockito-kotlin", version = "5.4.0" }
```
runningcode, line 64: why was this line moved?

CHANGELOG.md
```diff
@@ -24,6 +28,10 @@
   - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#8432)
   - [diff](https://github.com/getsentry/sentry-java/compare/8.43.1...8.43.2)
 
+### Dependencies
+
+- Bump mockito-kotlin from `com.nhaarman.mockitokotlin2:2.2.0` to `org.mockito.kotlin:5.4.0` ([#1286](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1286))
```
runningcode: Is this part of this PR?

plugin-build/build.gradle.kts
```diff
@@ -55,12 +55,17 @@ dependencies {
   testImplementation(libs.asmCommons)
 
   // we need these dependencies for tests, because the bytecode verifier also analyzes superclasses
+  testImplementationAar(libs.roomRuntimeAndroid)
+  testImplementationAar(libs.room3RuntimeAndroid)
+  testImplementationAar(libs.sentryAndroid)
+  testImplementationAar(libs.sentryAndroidOkhttp)
```
runningcode: would it make sense to keep `sentryAndroidOkhttp` next to `sentryOkhttp` like it was before to minimize the diff?

plugin-build/src/test/resources/testFixtures/instrumentation/androidxRoom/RoomDatabase$Builder.class
runningcode, line 1: If we depend on room already can we load the class files from the artifact instead of checking in the binaries?
```
testImplementationAar(libs.roomRuntimeAndroid)
testImplementationAar(libs.room3RuntimeAndroid)
```

plugin-build/src/test/kotlin/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,14 @@
+package io.sentry.sqlite
+
+import androidx.sqlite.SQLiteDriver
+
+/**
+ * Minimal stub of `io.sentry.sqlite.SentrySQLiteDriver` so ASM can resolve the `INVOKESTATIC
+ * io/sentry/sqlite/SentrySQLiteDriver.create` emitted by [SetDriverMethodVisitor]. The `@JvmStatic
+ * create(SQLiteDriver)` shape here mirrors the SDK contract the visitor depends on.
+ */
+class SentrySQLiteDriver private constructor(private val delegate: SQLiteDriver) : SQLiteDriver {
```
runningcode: Is this needed because we can't have aar dependencies in the gradle plugin?

runningcode (commented):
Thanks this looks good! I just don't think I'm enough of an expert in how we instrument classes to approve this.
I left some questions for my better understanding of the patterns we are using.

gradle/libs.versions.toml
0xadam-brown: Just to alphabetize (in line with the other sections in this file...mostly). I can move it back if you prefer, of course.

plugin-build/src/test/resources/testFixtures/instrumentation/androidxRoom/RoomDatabase$Builder.class
0xadam-brown: Yes, and good question: we could load via `ClassLoader.getResourceAsStream` rather than pinning bytecode. Your approach would be a win, too, because we'd keep testing against the latest as Dependabot bumps us. 

That said, this matches existing instrumentation tests. Happy to switch if folks want to move generally...

CHANGELOG.md
0xadam-brown: Thanks, and no. It's an artifact from when I merged in `main`. Removed.

plugin-build/build.gradle.kts
0xadam-brown: Reverted!

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+### Features
+
+- Auto-instrument SQLiteDriver for Room users ([#1285](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1285))
+  - Gated on `sentry-android-sqlite` >= 8.45.0 and the existing `tracingInstrumentation` `DATABASE` feature
+  - For users of the `androidx.sqlite.driver.SupportSQLiteDriver` bridge, auto-instrumentation wraps only the `SupportSQLiteOpenHelper` consumed by the bridge and not the bridge itself (avoids duplicate spans)
+
+### Dependencies
+
+- Bump Android SDK from v8.43.2 to v8.45.0 ([#TODO](https://github.com/getsentry/sentry-android-gradle-plugin/pull/TODO))
+  - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#TODO)
```
0xadam-brown: TODO: Update URLs once we publish Android SDK 8.45.0.

gradle/libs.versions.toml
```diff
@@ -4,7 +4,7 @@ agp = "8.10.1"
 asm = "9.4" # // compatibility matrix -> https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/instrumentation/InstrumentationContext#apiversion
 ktfmt = "0.51"
 sqlite = "2.1.0"
-sentry = "8.43.2"
+sentry = "8.44.1"
```
0xadam-brown: TODO: Bump to 8.45.0 once we publish the Android SDK.

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/Versions.kt
```diff
@@ -33,6 +33,7 @@ internal object SentryVersions {
   internal val VERSION_LOGCAT = SemVer(6, 17, 0)
   internal val VERSION_APP_START = SemVer(7, 1, 0)
   internal val VERSION_SQLITE = SemVer(6, 21, 0)
+  internal val VERSION_SQLITE_DRIVER = SemVer(8, 44, 1)
```
0xadam-brown: TODO: Bump to 8.45.0 once we publish the Android SDK.

plugin-build/src/test/kotlin/io/sentry/android/gradle/util/SentryModulesServiceTest.kt
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  @Test
+  fun `isSQLiteDriverInstrEnabled is false when sentry-android-sqlite is one patch below VERSION_SQLITE_DRIVER`() {
+    val belowThreshold =
+      SemVer(
+        SentryVersions.VERSION_SQLITE_DRIVER.major,
+        SentryVersions.VERSION_SQLITE_DRIVER.minor,
+        SentryVersions.VERSION_SQLITE_DRIVER.patch - 1,
+      )
```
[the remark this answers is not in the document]
0xadam-brown: TODO: Address once we bump the Android SDK to 8.45.0.

plugin-build/src/test/kotlin/io/sentry/sqlite/SentrySQLiteDriver.kt
0xadam-brown: It was there b/c `SentrySQLiteDriver` hadn't been published at the time I first posted this PR. We've published it now, so this has been removed.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/visitor/SetDriverMethodVisitor.kt
```diff
@@ -0,0 +1,35 @@
...
+  override fun onMethodEnter() {
+    loadArg(0)
+    invokeStatic(Type.getType(SENTRY_SQLITE_DRIVER_TYPE), Method(CREATE, SENTRY_CREATE_DESCRIPTOR))
+    storeArg(0)
+  }
+
+  companion object {
+    internal const val CREATE = "create"
+    internal const val SENTRY_CREATE_DESCRIPTOR =
+      "($SQLITE_DRIVER_TYPE_DESCRIPTOR)$SQLITE_DRIVER_TYPE_DESCRIPTOR"
```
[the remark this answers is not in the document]
0xadam-brown: This is a reality of SAGP having to depend on the signature of `SentrySQLiteDriver.create()`. I've added a comment to SAGP code noting as much, and I have a branch that does the same in the reverse direction from `SentrySQLiteDriver`. 

Verified current compatibility via javap on the compiled class (and Android sample app behavior, which wires us up E2E):

```
public static final androidx.sqlite.SQLiteDriver create(androidx.sqlite.SQLiteDriver);
    descriptor: (Landroidx/sqlite/SQLiteDriver;)Landroidx/sqlite/SQLiteDriver;
``

This matches the visitor descriptor and the .api file. Runtime may return a SentrySQLiteDriver instance, but JVM linkage uses the declared return type — SQLiteDriver — not the concrete class.

The bot's scenario would only apply if someone changed the SDK to:
```
fun create(delegate: SQLiteDriver): SentrySQLiteDriver
```

That'd be a deliberate cross-repo change.

File: CHANGELOG.md [hunk 19]
```diff
@@ -1,5 +1,21 @@
...
+## Unreleased
+
+### Features
+
+- Auto-instrument SQLiteDriver for Room users ([#1285](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1285))
+  - Gated on `sentry-android-sqlite` >= 8.45.0 and the existing `tracingInstrumentation` `DATABASE` feature (enabled by default)
+  - For users of the `androidx.sqlite.driver.SupportSQLiteDriver` bridge, auto-instrumentation wraps only the `SupportSQLiteOpenHelper` consumed by the bridge and not the bridge itself (avoids duplicate spans)
+
+## 6.12.0
+
```
```suggestion
```

plugin-build/src/test/resources/testFixtures/instrumentation/androidxRoom/RoomDatabase$Builder.class
romtsn: that's clever, I hadn't thought about this when implementing the test suite! That definitely sounds like a much better approach which would give us forwards-compat testing basically for free. That said, it's probably better to do it separately for all fixtures at once. If there's no time for that now, feel free to just create an issue

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/visitor/SetDriverMethodVisitor.kt
```diff
@@ -0,0 +1,37 @@
...
+    originalVisitor,
+    instrumentableContext.access,
+    instrumentableContext.name,
+    instrumentableContext.descriptor,
+  ) {
+
+  override fun onMethodEnter() {
+    loadArg(0)
+    invokeStatic(Type.getType(SENTRY_SQLITE_DRIVER_TYPE), Method(CREATE, SENTRY_CREATE_DESCRIPTOR))
+    storeArg(0)
```
romtsn, line 28: `l`: arg0 is the driver right? maybe drop a short comment saying it and that it's gonna be used as an argument to pass down to the setDriver method? (was just trying to wrap my head around how it works)

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/AndroidXSQLiteDriver.kt
```diff
@@ -0,0 +1,90 @@
...
+ * ```
+ *
+ * `SentrySQLiteDriver` protects against duplicate wrappings, allowing the visitor to wrap
+ * `SQLiteDriver` unconditionally.
+ *
+ * Coverage is limited to Room because SQLDelight
+ * [doesn't support `SQLiteDriver`](https://github.com/sqldelight/sqldelight/issues/6072) (it uses
+ * `SupportSQLiteOpenHelper`, which we auto-instrument via
+ * [AndroidXSQLiteOpenHelper][io.sentry.android.gradle.instrumentation.androidx.sqlite.AndroidXSQLiteOpenHelper]).
+ * To keep our implementation simple and build times fast, developers who use `SQLiteDriver`
```
romtsn, line 43: `l`: maybe worth opening an issue gathering interest around instrumenting the driver across the entire codebase (not only in Room)? Customers interested in that could upvote/comment in future

romtsn (approved):
nicely done, hats off!

CHANGELOG.md
0xadam-brown: Good catch! Removed in fad9e97b719b67a19cfda121673d006e5be6bf09.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/visitor/SetDriverMethodVisitor.kt
0xadam-brown: Added a comment in fad9e97b719b67a19cfda121673d006e5be6bf09 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/AndroidXSQLiteDriver.kt
0xadam-brown: Great idea. Created https://github.com/getsentry/sentry-android-gradle-plugin/issues/1338.

plugin-build/src/test/resources/testFixtures/instrumentation/androidxRoom/RoomDatabase$Builder.class
0xadam-brown: Good deal. Addressed the easy wins in https://github.com/getsentry/sentry-android-gradle-plugin/pull/1341.

CHANGELOG.md
0xadam-brown: Updated in b7c1a715b7c8eb4ebe1b778b993ef0323c343ead.

gradle/libs.versions.toml
0xadam-brown: Updated in. b7c1a715b7c8eb4ebe1b778b993ef0323c343ead.

plugin-build/src/main/kotlin/io/sentry/android/gradle/util/Versions.kt
0xadam-brown: Updated in b7c1a715b7c8eb4ebe1b778b993ef0323c343ead.

PR #1322, opened 2026-06-23
ci(matrix): fetch googlesource files via git instead of HTTP

### LINKED ISSUE #1240: Switch test matrix fetch to git-based source
CI jobs intermittently return 503 errors when fetching the test matrix, requiring multiple retries before succeeding. Root cause is suspected to be Google rate limiting the fetch endpoint. Switching to a git-based fetch would remove the dependency on Google's uptime.

* [Example failing job](<https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/26500548437/job/78040415517?pr=1231>)
* Reported by Nelson Osacky; Roman Zavarnitsyn suspects Google rate limiting as the root cause.

Action taken on behalf of Nelson Osacky.

runningcode:
## Summary

The `Generate Compat Matrix` step intermittently fails when `android.googlesource.com`'s gitiles `?format=TEXT` endpoint returns a 503, which blocks all downstream test-matrix jobs ([example failure](https://github.com/getsentry/sentry-android-gradle-plugin/actions/runs/28007449480/job/82892611645)):

```
java.io.IOException: Server returned HTTP response code: 503 for URL:
  https://android.googlesource.com/platform/tools/adt/idea/+/refs/heads/mirror-goog-studio-main/
  build-common/src/com/android/tools/idea/gradle/util/CompatibleGradleVersion.kt?format=TEXT
```

Closes #1240

## Changes

Replaces the `fetchGooglesourceText(url)` helper (which used `URL(url).readText()` + base64-decode) with `fetchGooglesourceViaGit()`, which does a depth-1, blob-less sparse clone of just the single file needed and reads it from disk.

Affected sources:
- `CompatibleGradleVersion.kt` (AGP → Gradle compat table)
- `SdkConstants.java` (`GRADLE_LATEST_VERSION` constant)

No other external HTTP calls in the script are changed.

## Verification

- Confirmed 503 as the root cause from the failed run logs
- New helper mirrors the same file content, just via `git clone --depth=1 --no-checkout --filter=blob:none --sparse` instead of HTTP
- `git` is available on all `ubuntu-latest` runners; no new dependencies

#skip-changelog

---
[View Session in Sentry](https://sentry.sentry.io/traces/?project=4510944073809921&query=gen_ai.conversation.id%3A%22slack%3AC089VFRB8N9%3A1782198317.321199%22)

### DIFF
--- a/scripts/generate-compat-matrix.main.kts
+++ b/scripts/generate-compat-matrix.main.kts
@@ -17,8 +17,8 @@ import io.github.z4kn4fein.semver.Version
 import io.github.z4kn4fein.semver.toVersion
 import java.io.File
 import java.net.URL
-import kotlin.io.path.createTempDirectory
 import javax.xml.parsers.DocumentBuilderFactory
+import kotlin.io.path.createTempDirectory
 import org.jsoup.Jsoup
 import org.w3c.dom.Element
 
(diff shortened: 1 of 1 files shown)


romtsn (approved):
preapproving

PR #1328, opened 2026-06-23
build(samples): Unpin NDK version in android-ndk sample

runningcode:
## Summary

The `android-ndk` sample pins `ndkVersion = "24.0.8215888"`. When Mend Renovate resolves the composite build in its hosted runner (`/home/ubuntu/.android-sdk`), configuring this sample forces AGP to install that exact NDK package. Its license is not accepted in Renovate's sandbox, so AGP throws `LicenceNotAcceptedException`, Renovate's dependency resolution fails, and it surfaces as **"package lookup failures"** on the [Dependency Dashboard](https://github.com/getsentry/sentry-android-gradle-plugin/issues/1308) (and the *"Some dependencies could not be looked up"* banner on bump PRs such as #1309).

Dropping the explicit pin lets AGP fall back to the NDK that is already installed and licensed in the environment, so Renovate can configure the build without tripping the license check.

#skip-changelog

### DIFF
--- a/examples/android-ndk/build.gradle
+++ b/examples/android-ndk/build.gradle
@@ -5,7 +5,6 @@ plugins {
 
 android {
     compileSdk = LibsVersion.SDK_VERSION
-    ndkVersion = "24.0.8215888"
 
     defaultConfig {
         minSdkVersion LibsVersion.MIN_SDK_VERSION


0xadam-brown (approved):
Thanks! Looks like this was originally pinned so contributors on Apple silicon wouldn't get a slow/buggy Rosetta-emulated NDK (see the commit msg to 552d6b5fb323cc4fbe8c8272b4474b4da0f2bb7b). Every NDK release since the pinned version has had first-class M1 support, so we should be good to go 👍

runningcode:
Thanks for looking that up!

romtsn (approved):

PR #1341, opened 2026-06-23
test: Replace binaries with bytecode loaded from test deps

0xadam-brown:
## :scroll: Description

Follow-up to the convo [here](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1285#discussion_r3435146833) about checked-in bytecode fixtures. For the easy cases, tests now load real library classes from test dependencies instead of committed .class files.

### Why not the rest?

The remaining fixtures are harder in that we can't just fetch a dep from Maven + delete an existing binary. Instead, we'd have to be more clever about how we extract or regenerate them, each of which would be more involved than a quick-win, sometimes without much payoff. 

Let's us avoid (potentially fragile and always static) binary check-ins and instead rely on evergreen dependencies.

Ran test suite after swapping out binaries.

## :pencil: Checklist

- [x] No breaking changes


runningcode (approved):
Very nice! I think this is much easier to maintain than checking in `.class` files!

runningcode:
What do you mean by this?
> Instead, we'd have to pin deps or a fixture regen pipeline

We do want to pin our dependencies as part of a security audit. I'm not sure what the fixture regen pipeline means

romtsn (approved):

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/InstrumentationBytecodeTestUtil.kt
```diff
@@ -0,0 +1,39 @@
+package io.sentry.android.gradle.instrumentation
+
+internal object InstrumentationBytecodeTestUtil {
+
+  /**
+   * Instrumentation bytecode loaded from published AARs on the test classpath:
+   * - AndroidX SQLite framework classes: `androidx.sqlite:sqlite-framework`
+   * - `TypefaceCompatUtil`: `androidx.core:core`
+   */
+  private val FIXTURE_TO_RESOURCE =
```
romtsn: wondering if we need to maintain this mapping or rather change VisitorTest parameters to fit the new structure? I guess simply passing a FQN should be enough when loading from an aar

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/SQLiteDriverBytecodeTestUtil.kt
```diff
@@ -11,29 +10,32 @@ import org.objectweb.asm.tree.MethodNode
...
    *
    * `VisitorTest` needs matching Room runtime AARs (and coroutines) on the test classpath so ASM
    * can resolve types referenced by the real bytecode.
    */
-  private val CLASS_NAME_TO_FIXTURE =
+  private val CLASS_NAME_TO_RESOURCE =
     mapOf(
-      "androidx.room.RoomDatabase\$Builder" to "RoomDatabase\$Builder",
-      "androidx.room3.RoomDatabase\$Builder" to "RoomDatabase3\$Builder",
+      "androidx.room.RoomDatabase\$Builder" to "androidx/room/RoomDatabase\$Builder.class",
```
romtsn: also not sure if this one is needed, we could just append `.class` when we load, right?

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/InstrumentationBytecodeTestUtil.kt
0xadam-brown: Good callout – updated in dd74f6fa3e5f4b382e9cc9c301ea8bc522975228.

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/androidx/sqlite/driver/SQLiteDriverBytecodeTestUtil.kt
0xadam-brown: Indeed – updated in dd74f6fa3e5f4b382e9cc9c301ea8bc522975228.

plugin-build/src/test/kotlin/io/sentry/android/gradle/instrumentation/VisitorTest.kt
```diff
@@ -172,7 +192,12 @@ class VisitorTest(
         kspTracksDaoTestParameters("insertAll"),
         kspTracksDaoTestParameters("update"),
         arrayOf("fileIO", "SQLiteCopyOpenHelper", WrappingInstrumentable(), null),
-        arrayOf("fileIO", "TypefaceCompatUtil", WrappingInstrumentable(), null),
+        arrayOf(
+          "androidx.core.graphics.TypefaceCompatUtil",
+          "TypefaceCompatUtil",
+          WrappingInstrumentable(),
+          TestClassContext("androidx.core.graphics.TypefaceCompatUtil"),
```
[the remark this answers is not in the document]
0xadam-brown: False positive. TypefaceCompatUtil is in core-1.15.0.aar's classes.jar despite being @RestrictTo(LIBRARY). VisitorTest passes with classpath loading – verified locally.

0xadam-brown:
> What do you mean by this?
> 
> > Instead, we'd have to pin deps or a fixture regen pipeline
> 
> We do want to pin our dependencies as part of a security audit. I'm not sure what the fixture regen pipeline means

Ah, sorry for the confusion. (I've updated the PR description.) I just meant that additional test updates wouldn't be a simple removal of the check-in binary + introduce a corresponding Maven coordinate. Instead, we'd have to be smarter and design a way to regenerate binaries from coordinates and/or pin specific versions for testing purposes.

PR #1349, opened 2026-07-01
fix(auto-install): Use BOM version for Sentry installs

adinauer:
Use the Sentry BOM version as the source of truth for auto-installed SDK dependencies.

Previously SAGP treated `sentry-bom` like a concrete SDK dependency when choosing whether to install the core SDK. That avoided adding `sentry-android`, but it also meant BOM-only projects could not use the BOM version for the dependencies SAGP installs.

This separates version detection from SDK presence detection. Concrete Sentry SDK dependencies still prevent SAGP from adding the core SDK again, while `sentry-bom` and `sentry-opentelemetry-bom` provide the version used for auto-installed dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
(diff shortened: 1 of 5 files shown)


runningcode (approved):
Great improvement!

lbloder (approved):

PR #1350, opened 2026-07-02
feat(auto-install): Fail the build on OpenTelemetry version downgrades

adinauer:
Adds a JVM-only `verifySentryOpenTelemetryVersions` Gradle task that fails the build when the OpenTelemetry versions resolved on the runtime classpath are downgraded below what Sentry's `sentry-opentelemetry-*` artifacts require.

## Why

Sentry's OpenTelemetry artifacts are built against a specific set of OpenTelemetry versions. When another dependency management mechanism — most commonly Spring Boot's `io.spring.dependency-management` — forces OpenTelemetry below those versions (which happens silently, without the user declaring any OTel version), running against the mismatched versions throws `ClassNotFoundException` / `NoSuchMethodError` at runtime. This surfaces the problem at build time with actionable guidance instead.

## How it works

- Walks the resolved `runtimeClasspath` graph and, for every OpenTelemetry module a `sentry-opentelemetry-*` artifact depends on, compares the requested vs. resolved version. Gradle preserves the originally-requested version on each edge even when a resolution rule (e.g. `io.spring.dependency-management`) forces it lower, so the downgrade is detectable.
- The failure message adapts to the setup: it suggests importing the Sentry OpenTelemetry BOM through `io.spring.dependency-management`'s `imports { mavenBom(...) }` when that plugin is applied (a `platform()` constraint loses to it), otherwise a plain `platform(...)` dependency.
- Gated by a cheap declared-dependency check so it's skipped (no runtime classpath resolution) for projects that don't use Sentry OpenTelemetry.
- Opt out via `sentry.autoInstallation.verifyOpenTelemetryVersions = false` (default `true`).

## Context

Stacked on #1349 (base branch), which adds `sentry-opentelemetry-bom` awareness to auto-install.

This replaces an earlier approach that auto-imported the BOM: that only works with Gradle-native BOM support and is silently overridden by `io.spring.dependency-management`, so detecting-and-failing is the more honest behavior.

## Open items from SDK-team review

- Docs URL in the failure message is a placeholder (`.../platforms/java/opentelemetry/troubleshooting/`) and must resolve to a live page before release.
- Default-on behavior is a product call: defaulting a build-breaking check to `true` will surprise existing Spring Boot users on upgrade. Either keep default-on with a clearly-flagged behavioral-change changelog entry, or ship opt-in first and flip later based on telemetry.


plugin-build/src/main/kotlin/io/sentry/jvm/gradle/SentryJvmPlugin.kt
```diff
@@ -130,6 +131,24 @@ constructor(private val buildEvents: BuildEventListenerRegistryInternal) : Plugi
...
+      // Gate compilation so any build that compiles, packages, runs or tests (assemble, bootJar,
+      // bootRun, check, ...) fails fast on an OpenTelemetry version mismatch.
+      project.tasks.named("classes").configure { it.dependsOn(verifyOpenTelemetryVersionsTask) }
     }
   }
+
+  companion object {
+    // TODO: finalize the troubleshooting docs URL before release.
+    internal const val OPENTELEMETRY_VERSION_MISMATCH_DOCS_URL =
+      "https://docs.sentry.io/platforms/java/opentelemetry/troubleshooting/"
```
adinauer, line 151: TODO docs don't exist yet

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Features
+
+- Fail the build when OpenTelemetry is downgraded below the version the Sentry OpenTelemetry integration requires ([#1350](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1350))
+  - The `sentry-opentelemetry-*` artifacts are built against specific OpenTelemetry versions. When another dependency management mechanism (most commonly Spring Boot's `io.spring.dependency-management`) forces OpenTelemetry below the version Sentry's integration requires, running against those downgraded versions can cause `ClassNotFoundException` / `NoSuchMethodError` at runtime. The new `verifySentryOpenTelemetryVersions` task detects this downgrade and fails the build early with guidance on how to fix it.
+  - Opt out with `sentry.autoInstallation.verifyOpenTelemetryVersions = false`
```
adinauer:
```suggestion
  - You may disable this check by setting `sentry.autoInstallation.verifyOpenTelemetryVersions = false`
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+ * below what the Sentry OpenTelemetry integration requires.
+ *
+ * Sentry's `sentry-opentelemetry-*` artifacts declare the exact OpenTelemetry versions they were
+ * built and tested against. When another dependency management mechanism (most commonly
+ * `io.spring.dependency-management` / the Spring Boot BOM) forces a lower OpenTelemetry version,
+ * running against those downgraded versions can cause `ClassNotFoundException` /
+ * `NoSuchMethodError` at runtime. We detect that downgrade here and fail fast with actionable
+ * guidance instead of letting it blow up at runtime.
+ */
+abstract class SentryOpenTelemetryVersionCheckTask : DefaultTask() {
```
runningcode, line 30: This task is missing either a `@DisableCachingByDefault` or a `@CacheableTask` annotation. Check the build failure for details.

In this case since we have no outputs it should probably be `@DisableCachingByDefault`. Otherwise we can write a simple text file in a task specific output directory and make this task cacheable.

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryOpenTelemetryVersionCheckTest.kt
```diff
@@ -0,0 +1,332 @@
...
+
+import java.io.File
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import org.gradle.testkit.runner.TaskOutcome
+import org.gradle.util.GradleVersion
+import org.junit.Test
+
+class SentryOpenTelemetryVersionCheckTest :
```
runningcode, line 9: Can we add a configuration cache test?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+      val requested: String,
+      val resolved: String,
+    )
+
+    /**
+     * Walks the resolved dependency graph and returns, for every OpenTelemetry module that a
+     * `sentry-opentelemetry-*` artifact depends on, the ones whose resolved version is lower than
+     * the version Sentry requested.
+     */
+    internal fun collectDowngrades(root: ResolvedComponentResult): List<VersionMismatch> {
```
runningcode: do you think it would be possible to split this pure logic from the gradle task and then unit test that separately?
I'm not saying we need to add more tests, but maybe we can move some of the tests from integration tests to just be plain unit tests which are significantly faster than integration tests that spin up a whole gradle build

plugin-build/src/test/kotlin/io/sentry/android/gradle/integration/SentryOpenTelemetryVersionCheckTest.kt
```diff
@@ -0,0 +1,332 @@
...
+        .trimIndent()
+    )
+
+    val result = runner.appendArguments("app:verifySentryOpenTelemetryVersions").buildAndFail()
+
+    assertEquals(TaskOutcome.FAILED, result.task(verifyTask)?.outcome)
+    assertTrue {
+      "OpenTelemetry was downgraded below the version its integration requires" in result.output
+    }
+    assertTrue { "io.opentelemetry:opentelemetry-sdk" in result.output }
```
runningcode: Can we use Google Truth instead of `kotlin.test` error messages and debugging is a lot easier especially when matching strings.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  companion object {
+    private const val SENTRY_GROUP = "io.sentry"
+    private const val SENTRY_OPENTELEMETRY_ARTIFACT_PREFIX = "sentry-opentelemetry-"
+    private const val OTEL_GROUP = "io.opentelemetry"
+    private const val SPRING_DEPENDENCY_MANAGEMENT_PLUGIN_ID = "io.spring.dependency-management"
+
+    internal data class VersionMismatch(
```
runningcode: nit: In this case we only ever hold downgrades so this might be a better name:
```suggestion
    internal data class VersionDowngrade(
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+          }
+          stack.addLast(dependency.selected)
+        }
+      }
+      return mismatches.values.toList()
+    }
+
+    private fun ResolvedComponentResult.isSentryOpenTelemetryArtifact(): Boolean {
+      val moduleVersion = moduleVersion ?: return false
+      return moduleVersion.group == SENTRY_GROUP &&
```
runningcode: nit: this logic is repeated on line 229. would it be worth making a function?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+          // io.spring.dependency-management enforces its managed versions via a resolution strategy
+          // that overrides Gradle platform() constraints, so the BOM has to be imported through the
+          // plugin's own API to win.
+          """
+          |To fix this, import the Sentry OpenTelemetry BOM through io.spring.dependency-management so
+          |its versions win dependency resolution:
+          |
+          |  dependencyManagement {
+          |    imports {
+          |      mavenBom("io.sentry:sentry-opentelemetry-bom:<sentryVersion>")
```
runningcode: would it make sense to fill in <sentryVersion> from the value in `AutoInstallState.sentryVersion` ?

and also for 174?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+          |  }
+          """
+            .trimMargin()
+        } else {
+          """
+          |To fix this, add the Sentry OpenTelemetry BOM as a platform dependency so its versions win
+          |dependency resolution:
+          |
+          |  dependencies {
+          |    implementation platform("io.sentry:sentry-opentelemetry-bom:<sentryVersion>")
```
runningcode: this is groovy syntax (also on line 163)

I think this works on both groovy and kotlin:
```suggestion
          |    implementation(platform("io.sentry:sentry-opentelemetry-bom:<sentryVersion>"))
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+  companion object {
+    private const val SENTRY_GROUP = "io.sentry"
+    private const val SENTRY_OPENTELEMETRY_ARTIFACT_PREFIX = "sentry-opentelemetry-"
+    private const val OTEL_GROUP = "io.opentelemetry"
+    private const val SPRING_DEPENDENCY_MANAGEMENT_PLUGIN_ID = "io.spring.dependency-management"
+
+    internal data class VersionMismatch(
+      val module: String,
+      val requested: String,
+      val resolved: String,
```
runningcode: We can get more information from Gradle's dependency management engine in order to improve the error message.

```suggestion
      val resolved: String,
      val requestedBy: String,
      val reason: String,
```

Set `requestedBy` from `component.moduleVersion` (which you already know is the Sentry OTel artifact), and pull reason from the selection reason of the selected node:

```
private fun ResolvedComponentResult.downgradeReason(): String? =
  selectionReason.descriptions
    .lastOrNull {
      it.cause == ComponentSelectionCause.CONSTRAINT ||
        it.cause == ComponentSelectionCause.FORCED ||
        it.cause == ComponentSelectionCause.CONFLICT_RESOLUTION
    }
    ?.description
```

Let me know if you need more info here!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
```diff
@@ -0,0 +1,238 @@
...
+          |    implementation platform("io.sentry:sentry-opentelemetry-bom:<sentryVersion>")
+          |  }
+          """
+            .trimMargin()
+        }
+      return """
+        |Sentry detected that OpenTelemetry was downgraded below the version its integration requires.
+        |
+        |The Sentry OpenTelemetry integration was built against specific OpenTelemetry versions,
+        |but the following were downgraded by another dependency management mechanism:
```
runningcode: You can also add that users can run this command or check a build scan for more details:

```./gradlew :app:dependencyInsight --configuration runtimeClasspath --dependency io.opentelemetry:opentelemetry-sdk```

File: plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/AutoInstallExtension.kt [hunk 13]
```diff
@@ -18,4 +18,12 @@ open class AutoInstallExtension @Inject constructor(objects: ObjectFactory) {
    */
   val sentryVersion: Property<String> =
     objects.property(String::class.java).convention(SENTRY_SDK_VERSION)
+
+  /**
+   * Fails the build when the OpenTelemetry versions resolved on the runtime classpath were
```
runningcode: Can we frame comment around what the toggle controls? For example:


```suggestion
   Whether to verify that the OpenTelemetry versions resolved on the runtime classpath satisfy what the Sentry OpenTelemetry integration requires, failing the build if any were downgraded.
   Defaults to `true`.
```

runningcode (commented):
Thanks for putting this together. Overall looks good. I left some comments, let me know if you'd like to discuss any of these.

plugin-build/src/main/kotlin/io/sentry/jvm/gradle/SentryJvmPlugin.kt
adinauer: PR is up: https://github.com/getsentry/sentry-docs/pull/18647

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
adinauer: I'm afraid of getting the command wrong and adding even more confusion. Also not fully sure if this works for Spring Boot so I'd rather skip this for now.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryOpenTelemetryVersionCheckTask.kt
adinauer: Skipping the output file for now. We can always add that later if needed.

runningcode (approved):
Looks great! Thanks for the cleanup and adding the configuration cache test. 

I realized we should separate integrationTests from unitTests into separate source directories/tasks but that is an already existing organization issue.

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/AutoInstallExtension.kt
```diff
@@ -18,4 +18,12 @@ open class AutoInstallExtension @Inject constructor(objects: ObjectFactory) {
    */
   val sentryVersion: Property<String> =
     objects.property(String::class.java).convention(SENTRY_SDK_VERSION)
+
+  /**
+   * Whether to verify that the OpenTelemetry versions resolved on the runtime classpath satisfy
+   * what the Sentry OpenTelemetry integration requires, failing the build if any were downgraded.
+   * Defaults to `true`.
+   */
+  val verifyOpenTelemetryVersions: Property<Boolean> =
```
adinauer: Since this isn't tied to auto install, the option shouldn't live under auto install.

PR #1373, opened 2026-07-23
build(deps): Bump sentry-cli to 3.6.2

runningcode:
Bumps the bundled `sentry-cli` from 3.6.1 to 3.6.2.

## Why the repo needs reconfiguring

sentry-cli 3.6.2 [no longer combines a URL and a non-embedded auth token from different configuration sources](https://github.com/getsentry/sentry-cli/pull/3378): when the URL and token come from different sources, one is ignored with a warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## Unreleased
+
+### Breaking Changes
+
+- This breaking change is only for customers using self-hosted Sentry together with a user auth token (not an org auth token) and the url and auth token are configured separately:
+  - If so, this breaking change applies to you in order to patch a security flaw. [Please read this](https://github.com/getsentry/sentry-cli/issues/3380#issuecomment-5059013026) for further details.
+
+### Dependencies
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,9 +8,9 @@
 
 ### Dependencies
 
-- Bump CLI from v3.6.0 to v3.6.1 ([#1367](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1367))
-  - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#361)
-  - [diff](https://github.com/getsentry/sentry-cli/compare/3.6.0...3.6.1)
+- Bump CLI from v3.6.0 to v3.6.2 ([#1367](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1367), [#1373](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1373))
```
runningcode: Should we call out this breaking change more explicitly here? The fact that this is a minor bump on the CLI but is a breaking change might be confusing.

romtsn: yeah, let's have a breaking change section (or behavior change something like that) and reference Daniel's comment on the sentry-cli PR

romtsn (approved):
pre-approving, but let's update the changelog too

PR #1375, opened 2026-07-30
perf(instrumentation): Resolve SDK class availability at build time

romtsn:
Resolve the availability of optional Sentry SDK classes from the app's dependency graph and inject the results into `LoadClass` during bytecode instrumentation. The optimization is independent from tracing instrumentation, enabled by default, and can be disabled with `sentry.sdkOptimization.enabled = false`.

The visitor safely ignores older SDK versions without the availability field. Applications using the supporting SDK avoid reflective availability probes during startup, while unknown class names still fall back to reflection.

Supporting SDK change: [getsentry/sentry-java#5875](https://github.com/getsentry/sentry-java/pull/5875)

On a Pixel 2 XL, map construction and lookups were about 160× faster than the equivalent absent-heavy 11-class reflection batch. In the absent-heavy startup macrobenchmark, SDK initialization was 1.08% faster and full startup was 0.88% faster, though variance exceeded the measured delta. The representative 10-present/1-absent case showed no measurable improvement.

Verified with Spotless and focused visitor/plugin tests.

Refs [JAVA-654](https://linear.app/getsentry/issue/JAVA-654/move-reflection-checks-to-compile-time)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,12 @@
 - This breaking change is only for customers using self-hosted Sentry together with a user auth token (not an org auth token) and the url and auth token are configured separately:
   - If so, this breaking change applies to you in order to patch a security flaw. [Please read this](https://github.com/getsentry/sentry-cli/issues/3380#issuecomment-5059013026) for further details.
 
-### Features
+### Performance
 
-- Resolve optional Sentry SDK class availability at build time to reduce SDK initialization overhead ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
+- Eliminate the reflection cost of known optional Sentry SDK class-availability checks by resolving them at build time ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
+  - On a Pixel 2 XL, constructing and querying the availability map was about 160x faster than an absent-heavy batch of 11 reflection checks.
+  - In an absent-heavy startup benchmark, SDK initialization was 1.08% faster and full startup was 0.88% faster. These differences were within benchmark variance.
+  - This optimization is enabled by default. If it causes problems, disable it with `sentry.runtimeOptimizations.enabled = false`.
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 - Support ProGuard mapping tasks when R8 is enabled with the AGP app `optimization.enable` DSL ([#1376](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1376))
 
+### Performance
+
+- Eliminate the reflection cost of known optional Sentry SDK class-availability checks by resolving them at build time ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
+  - On a Pixel 2 XL, constructing and querying the availability map was about 160x faster than an absent-heavy batch of 11 reflection checks.
+  - In an absent-heavy startup benchmark, SDK initialization was 1.08% faster and full startup was 0.88% faster. These differences were within benchmark variance.
+  - This optimization is enabled by default. If it causes problems, disable it with `sentry.runtimeOptimizations.enabled = false`.
+
 ### Dependencies
 
 - Bump Android SDK from v8.51.0 to v8.52.0 ([#1378](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1378))
@@ -18,13 +25,6 @@
 
 - This breaking change is only for customers using self-hosted Sentry together with a user auth token (not an org auth token) and the url and auth token are configured separately:
   - If so, this breaking change applies to you in order to patch a security flaw. [Please read this](https://github.com/getsentry/sentry-cli/issues/3380#issuecomment-5059013026) for further details.
-
-### Performance
-
-- Eliminate the reflection cost of known optional Sentry SDK class-availability checks by resolving them at build time ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
-  - On a Pixel 2 XL, constructing and querying the availability map was about 160x faster than an absent-heavy batch of 11 reflection checks.
-  - In an absent-heavy startup benchmark, SDK initialization was 1.08% faster and full startup was 0.88% faster. These differences were within benchmark variance.
-  - This optimization is enabled by default. If it causes problems, disable it with `sentry.runtimeOptimizations.enabled = false`.
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,7 @@
 
 ### Performance
 
-- Eliminate the reflection cost of known optional Sentry SDK class-availability checks by resolving them at build time ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
-  - On a Pixel 2 XL, constructing and querying the availability map was about 160x faster than an absent-heavy batch of 11 reflection checks.
-  - In an absent-heavy startup benchmark, SDK initialization was 1.08% faster and full startup was 0.88% faster. These differences were within benchmark variance.
+- Eliminate reflection for known optional Sentry SDK class-availability checks, reducing SDK initialization time by about 1% in an absent-heavy startup benchmark ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
   - This optimization is enabled by default. If it causes problems, disable it with `sentry.runtimeOptimizations.enabled = false`.
 
 ### Dependencies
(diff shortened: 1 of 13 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -177,35 +178,45 @@ fun ApplicationAndroidComponentsExtension.configure(
         }
       }
 
-      if (extension.tracingInstrumentation.enabled.get()) {
-        /**
-         * We detect sentry-android SDK version using configurations.incoming.afterResolve. This is
-         * guaranteed to be executed BEFORE any of the build tasks/transforms are started.
-         *
-         * After detecting the sdk state, we use Gradle's shared build service to persist the state
```
runningcode, line 185: good to delete this. this is false, build services are not kept across builds.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/LoadClassClassVisitor.kt
```diff
@@ -0,0 +1,118 @@
+package io.sentry.android.gradle.instrumentation
+
+import org.objectweb.asm.ClassVisitor
+import org.objectweb.asm.FieldVisitor
+import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Opcodes
+
+internal class LoadClassClassVisitor(
```
runningcode, line 27: I think a small javadoc on this one would be helpful

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/RuntimeOptimizationsExtension.kt
```diff
@@ -0,0 +1,10 @@
+package io.sentry.android.gradle.extensions
+
+import javax.inject.Inject
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+
+open class SdkOptimizationExtension @Inject constructor(objects: ObjectFactory) {
+  /** Enables build-time optimizations of the Sentry SDK. Defaults to true. */
+  val enabled: Property<Boolean> = objects.property(Boolean::class.java).convention(true)
```
runningcode, line 11: aren't we going to have different optimizations? like properties, manifest and this one? Will they all be gated behind the same flag?

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/RuntimeOptimizationsExtension.kt
```diff
@@ -0,0 +1,10 @@
+package io.sentry.android.gradle.extensions
+
+import javax.inject.Inject
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+
```
runningcode, line 6: Should we name this `StartupTimeOptimizations`?

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/SdkOptimizationExtension.kt
```diff
@@ -0,0 +1,10 @@
+package io.sentry.android.gradle.extensions
+
+import javax.inject.Inject
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+
+open class SdkOptimizationExtension @Inject constructor(objects: ObjectFactory) {
+  /** Enables build-time optimizations of the Sentry SDK. Defaults to true. */
```
runningcode:
```suggestion
  /** Enables run-time optimizations of the Sentry SDK at the cost of build time. Defaults to true. */
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
```diff
@@ -177,35 +178,45 @@ fun ApplicationAndroidComponentsExtension.configure(
...
-         * between builds and also during a single build, because transforms are run in parallel.
-         */
-        val sentryModulesService =
+      val sdkOptimizationEnabled = extension.sdkOptimization.enabled.get()
```
runningcode: like above, i think we should rename this to `runtimeOptimizations` or something similar

runningcode (approved):
Looks good!

One more thing, should we detect an incompatible version of the sentry-java sdk and then show an error?

plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
romtsn: renamed to `runtimeOptimizations` in https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375/commits/bb60a243c53c03e008437570eb26c4a16badfed5

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/LoadClassClassVisitor.kt
romtsn: added in https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375/commits/d39f714f1df52461cbbba89a909e47d2b9fb6963

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/RuntimeOptimizationsExtension.kt
romtsn: i'd keep it at `runtime` because these may apply beyond startup as well (e.g. package info/in-app stuff)

plugin-build/src/main/kotlin/io/sentry/android/gradle/extensions/RuntimeOptimizationsExtension.kt
romtsn: yup that's right -- but I've seen this flag as a kill-switch if something goes wrong rather than a way to control the optimizations granularly. I think 99% of the time people would want all of these optimizations to be turned on, unless something goes horribly wrong (that we have a kill switch for).

romtsn:
> Looks good!
> 
> One more thing, should we detect an incompatible version of the sentry-java sdk and then show an error?

we usually don't do that, since we check for the runtime sdk version and gracefully fallback, but I've added an info log anyway 👍

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
 
 - Support ProGuard mapping tasks when R8 is enabled with the AGP app `optimization.enable` DSL ([#1376](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1376))
 
+### Performance
+
+- Eliminate the reflection cost of known optional Sentry SDK class-availability checks by resolving them at build time ([#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375))
+  - On a Pixel 2 XL, constructing and querying the availability map was about 160x faster than an absent-heavy batch of 11 reflection checks.
```
runningcode: I think we don't need all these details for the changelog, people can click the PR description to get these details.
I would remove line 12 at least and maybe just fold line 13 in to line 11 to say sdk initialization is 1% faster.

runningcode (approved):
LGTM!

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SentrySdkOptimizationClassVisitorFactory.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private fun module(group: String, name: String): ModuleIdentifier =
+      DefaultModuleIdentifier.newId(group, name)
+  }
+}
+
+internal fun resolveClassAvailability(modules: Set<ModuleIdentifier>): Map<String, Boolean> =
+  SentrySdkOptimizationClassVisitorFactory.CLASS_MODULES.mapValues { (_, owners) ->
+    owners.any { it in modules }
+  }
```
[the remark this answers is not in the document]
romtsn: i think `allComponents` gives us all dependencies anyway.

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SentrySdkOptimizationClassVisitorFactory.kt
```diff
@@ -0,0 +1,74 @@
...
+      instrumentationContext.apiVersion.get(),
+      nextClassVisitor,
+      parameters.get().classAvailability.get(),
+    )
+  }
+
+  // Empty availability means the runtime classpath is unknown. Skip this transformation so
+  // LoadClass falls back to reflection.
+  override fun isInstrumentable(classData: ClassData): Boolean =
+    classData.className == LOAD_CLASS_NAME && parameters.get().classAvailability.get().isNotEmpty()
```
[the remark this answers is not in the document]
romtsn: resolveClassAvailability() is not always called:

  1. Missing runtime configuration returns an absent Provider.
  2. Mapping an absent provider remains absent.
  3. .orElse(emptyMap()) supplies a genuinely empty map.
  4. isNotEmpty() is then false, preserving reflection.

  When configuration succeeds with no optional modules, the non-empty all-false map is intentional: those classes are considered known absent, so their reflection checks can be eliminated.

PR #1393, opened 2026-08-11
ref(variants): Use AndroidVariant83 wherever AGP 8.3+ is available

romtsn:
Follow-up to #1382. On AGP 8.3+ every `SentryVariant` consumer now gets `AndroidVariant83`, whose `isDebuggable` reads the public `Variant.debuggable` API. Before, only the proguard-mappings path used it; the native-symbols path — the only consumer that actually branches on `isDebuggable` — always got `AndroidVariant74`, which casts to the internal `ApplicationVariantImpl` and needs the analytics unwrap from #1382 to survive wrapped variants. With this change, the impl cast only matters for AGP 7.4–8.2 users; on 8.3+ the public API is wrapper-safe by construction.

Concretely, a new `Variant.toSentryVariant()` factory picks `AndroidVariant83` when `AgpVersions.isAGP83` and replaces the direct `AndroidVariant74(...)` constructions (telemetry, source bundles, proguard mappings, distribution, snapshots, and the native-symbols/sources site). `configureUploadAppTasks` keeps constructing `AndroidVariant74` because it needs its class-specific `bundle`/`apk` members and never reads `isDebuggable`.

The single-use `assetsWiredWithDirectories` helper is inlined at its only call site via the public `variant.artifacts` API, so the asset-transform wiring no longer depends on the concrete variant class either.

`assembleProvider`/`installProvider` still go through the 74 delegate's `unwrapImpl()` — there is no public replacement for `taskContainer`. Once the minimum supported AGP reaches 8.3, `AndroidVariant74.isDebuggable` and the factory can be folded away entirely.

#skip-changelog


runningcode (approved):
Looks great! The non-whitespace diff makes this much easier to review!

PR #1404, opened 2026-08-13
perf(instrumentation): [Generated metadata 1] Generate class availability

romtsn:
Generate an app-owned Java source containing SDK integration class availability from the variant runtime classpath. The ASM transform now emits a stable symbolic call to that generated class and no longer carries variant-specific inputs.

This moves dependency resolution into a cacheable source-generation task, avoiding runtime classpath resolution during Gradle configuration and keeping the transform isolated and configuration-cache compatible.

The SDK-side availability field shipped in [sentry-java#5875](https://github.com/getsentry/sentry-java/pull/5875). Older SDK versions are detected and left unchanged, retaining reflection as the fallback.

### DIFF
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/LoadClassClassVisitor.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/LoadClassClassVisitor.kt
@@ -17,18 +17,14 @@ import org.objectweb.asm.Opcodes
  *
  * After instrumentation the generated static initializer is equivalent to:
  * ```java
- * classAvailability = new HashMap<>();
- * classAvailability.put("timber.log.Timber", true);
+ * classAvailability = SentryGeneratedBuildTimeOptions.getClassAvailability();
  * ```
  *
  * Known entries avoid reflection; omitted class names still fall back to it. SDK versions without
  * the field are left unchanged.
  */
-internal class LoadClassClassVisitor(
-  apiVersion: Int,
-  nextClassVisitor: ClassVisitor,
-  private val classAvailability: Map<String, Boolean>,
-) : ClassVisitor(apiVersion, nextClassVisitor) {
+internal class LoadClassClassVisitor(apiVersion: Int, nextClassVisitor: ClassVisitor) :
+  ClassVisitor(apiVersion, nextClassVisitor) {
   private var hasAvailabilityField = false
   private var hasStaticInitializer = false
 
@@ -94,48 +90,27 @@ internal class LoadClassClassVisitor(
   }
 
   private fun injectClassAvailability(visitor: MethodVisitor) {
-    visitor.visitTypeInsn(Opcodes.NEW, HASH_MAP_NAME)
-    visitor.visitInsn(Opcodes.DUP)
-    visitor.visitMethodInsn(Opcodes.INVOKESPECIAL, HASH_MAP_NAME, "<init>", "()V", false)
+    visitor.visitMethodInsn(
+      Opcodes.INVOKESTATIC,
+      GENERATED_OPTIONS_INTERNAL_NAME,
+      "getClassAvailability",
+      "()Ljava/util/Map;",
+      false,
+    )
     visitor.visitFieldInsn(
       Opcodes.PUTSTATIC,
       LOAD_CLASS_INTERNAL_NAME,
       AVAILABILITY_FIELD,
       MAP_DESCRIPTOR,
     )
-
-    classAvailability.forEach { (className, available) ->
-      visitor.visitFieldInsn(
-        Opcodes.GETSTATIC,
-        LOAD_CLASS_INTERNAL_NAME,
-        AVAILABILITY_FIELD,
-        MAP_DESCRIPTOR,
-      )
-      visitor.visitLdcInsn(className)
-      visitor.visitInsn(if (available) Opcodes.ICONST_1 else Opcodes.ICONST_0)
-      visitor.visitMethodInsn(
-        Opcodes.INVOKESTATIC,
-        "java/lang/Boolean",
-        "valueOf",
-        "(Z)Ljava/lang/Boolean;",
-        false,
-      )
-      visitor.visitMethodInsn(
-        Opcodes.INVOKEINTERFACE,
-        "java/util/Map",
-        "put",
-        "(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;",
-        true,
-      )
-      visitor.visitInsn(Opcodes.POP)
-    }
   }
 
   private companion object {
     const val LOAD_CLASS_INTERNAL_NAME = "io/sentry/util/LoadClass"
+    const val GENERATED_OPTIONS_INTERNAL_NAME =
+      "io/sentry/android/core/SentryGeneratedBuildTimeOptions"
     const val AVAILABILITY_FIELD = "classAvailability"
     const val MAP_DESCRIPTOR = "Ljava/util/Map;"
-    const val HASH_MAP_NAME = "java/util/HashMap"
     const val STATIC_INITIALIZER = "<clinit>"
     const val VOID_METHOD_DESCRIPTOR = "()V"
   }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SentrySdkOptimizationClassVisitorFactory.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SentrySdkOptimizationClassVisitorFactory.kt
@@ -7,32 +7,20 @@ import com.android.build.api.instrumentation.InstrumentationParameters
 import io.sentry.android.gradle.util.SentryModules
 import org.gradle.api.artifacts.ModuleIdentifier
 import org.gradle.api.internal.artifacts.DefaultModuleIdentifier
-import org.gradle.api.provider.MapProperty
-import org.gradle.api.tasks.Input
 import org.objectweb.asm.ClassVisitor
 
 abstract class SentrySdkOptimizationClassVisitorFactory :
-  AsmClassVisitorFactory<SentrySdkOptimizationClassVisitorFactory.SdkOptimizationParameters> {
-
-  interface SdkOptimizationParameters : InstrumentationParameters {
-    @get:Input val classAvailability: MapProperty<String, Boolean>
-  }
+  AsmClassVisitorFactory<InstrumentationParameters.None> {
 
   override fun createClassVisitor(
     classContext: ClassContext,
     nextClassVisitor: ClassVisitor,
   ): ClassVisitor {
-    return LoadClassClassVisitor(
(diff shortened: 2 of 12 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/LoadClassClassVisitor.kt
```diff
@@ -94,48 +90,27 @@ internal class LoadClassClassVisitor(
...
     )
-
-    classAvailability.forEach { (className, available) ->
-      visitor.visitFieldInsn(
-        Opcodes.GETSTATIC,
-        LOAD_CLASS_INTERNAL_NAME,
-        AVAILABILITY_FIELD,
-        MAP_DESCRIPTOR,
-      )
-      visitor.visitLdcInsn(className)
```
runningcode, line 114: what did lines 114-131 do and why don't we need them anymore?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -0,0 +1,94 @@
...
+      moduleIds
+        .get()
+        .mapNotNull { coordinate ->
+          val parts = coordinate.split(':', limit = 2)
+          if (parts.size == 2) DefaultModuleIdentifier.newId(parts[0], parts[1]) else null
+        }
+        .toSet()
+    val availability = resolveClassAvailability(modules)
+    val sourceFile = File(output.get().asFile, GENERATED_CLASS_PATH)
+    sourceFile.parentFile.mkdirs()
```
runningcode, line 30: if this is properly marked as an output directory we don't need to do this.

plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/CollectSourcesTask.kt
```diff
@@ -74,10 +74,16 @@ internal class SourceCollector {
     sourceDirs.forEach { sourceDir ->
       if (sourceDir.exists()) {
         SentryPlugin.logger.debug { "Collecting sources in ${sourceDir.absolutePath}" }
-        sourceDir.walk().forEach { sourceFile ->
+        for (sourceFile in sourceDir.walk()) {
           val relativePath = sourceFile.toRelativeString(sourceDir)
           val targetFile = outDir.resolve(File(relativePath))
           if (sourceFile.isFile) {
+            if (
+              relativePath.replace(File.separatorChar, '/') ==
```
runningcode: i feel like there might be edge cases here with this replacement.
would calling `.normalize()` on the `sourceFile` and then calling `toRelativeString` work?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -0,0 +1,94 @@
...
+    val modules =
+      moduleIds
+        .get()
+        .mapNotNull { coordinate ->
+          val parts = coordinate.split(':', limit = 2)
+          if (parts.size == 2) DefaultModuleIdentifier.newId(parts[0], parts[1]) else null
+        }
+        .toSet()
+    val availability = resolveClassAvailability(modules)
+    val sourceFile = File(output.get().asFile, GENERATED_CLASS_PATH)
```
runningcode: I think this should work and is more gradle idiomatic
```suggestion
    val sourceFile = output.file(GENERATED_CLASS_PATH).get().asFile
```

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -0,0 +1,94 @@
...
+          )
+          return null
+        }
+
+      return project.tasks.register(
+        "generateSentryBuildTimeOptions$taskSuffix",
+        GenerateSentryBuildTimeOptionsTask::class.java,
+      ) { task ->
+        task.moduleIds.set(
+          configurationProvider.map { configuration ->
```
runningcode, line 80: i can't tell if this eagerly resolves the configuration. if so, we should avoid that as resolving the configuration would mean resolving all the dependencies at configuration time.
also we should check if this is configuration cache compatible since I know that the `configuration` object can't be serialized in to the configuration cache.
i think itd be worth adding an integration test for this!

plugin-build/src/test/kotlin/io/sentry/android/core/SentryGeneratedBuildTimeOptions.kt
```diff
@@ -0,0 +1,7 @@
+package io.sentry.android.core
+
+object SentryGeneratedBuildTimeOptions {
```
runningcode, line 3: since we're generating java, should this file be in java as well?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -0,0 +1,94 @@
...
+        }
+      }
+      """
+        .trimIndent() + "\n"
+    )
+  }
+
+  companion object {
+    private const val GENERATED_CLASS_PATH =
+      "io/sentry/android/core/SentryGeneratedBuildTimeOptions.java"
```
runningcode, line 58: just a note: mixing kotlin and java sources in the same module drastically slows down compilation speed. not saying we should do anything about it but just noting it here.

runningcode (approved):
looks good! main concern would be the configuration cache issue if it is an issue!

plugin-build/src/main/kotlin/io/sentry/android/gradle/sourcecontext/CollectSourcesTask.kt
romtsn: good catch! fixed it by comparing two normalized paths to `sourceDir`. since this is a direct file comparison we don't need toRelativeString

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
romtsn: thanks, applied!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
romtsn: i tested this - it's marked as @OutputDirectory, however, that one only creates the root dir automatically but not the subdirs, so we still need `mkdirs()` here

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/LoadClassClassVisitor.kt
romtsn: It used to instrument like this:
```java
static Map<String, Object> classAvailability;

static {
  classAvailability = new HashMap();
  classAvailability.put(..., ...)
  ...
}
```
now it's just this
```java
static Map<String, Object> classAvailability = /* this is injected by SAGP */ SentryGeneratedBuildTimeOptions.getClassAvailability(); 
```
after we've switched to generating `SentryGeneratedBuildTimeOptions` in the plugin.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
romtsn: `configurations.named()` and the following `map` remain lazy, and our unit test verifies registration does not resolve the runtime classpath. I’ve also made the configuration-cache integration test explicitly assert this task runs and is up-to-date when the cache is reused.

plugin-build/src/test/kotlin/io/sentry/android/core/SentryGeneratedBuildTimeOptions.kt
romtsn: i think it's fine to keep it in Kotlin since the generated bytecode matches that of java anyway (thanks to `@JvmStatic`)

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
romtsn: that's a great point I haven't thought of! I guess there's no an easy enough way for us to identify if the module contains only kotlin or only java files and generate respectively. I guess i'd keep it in mind until someone raises a real build time overhead concern

plugin-build/src/main/kotlin/io/sentry/android/gradle/instrumentation/SentrySdkOptimizationClassVisitorFactory.kt
```diff
@@ -7,32 +7,20 @@ import com.android.build.api.instrumentation.InstrumentationParameters
...
-      parameters.get().classAvailability.get(),
-    )
+    return LoadClassClassVisitor(instrumentationContext.apiVersion.get(), nextClassVisitor)
   }
 
-  // Empty availability means the runtime classpath is unknown. Skip this transformation so
-  // LoadClass falls back to reflection.
   override fun isInstrumentable(classData: ClassData): Boolean =
-    classData.className == LOAD_CLASS_NAME && parameters.get().classAvailability.get().isNotEmpty()
+    classData.className == LOAD_CLASS_NAME
```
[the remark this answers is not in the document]
romtsn: The injected INVOKESTATIC is present before R8 performs liveness analysis, so R8 sees a direct reference to the generated class. It may inline and remove the class, but it cannot remove the target while leaving the call. The minified release integration build also exercises this path, so no keep rule should be necessary.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
runningcode: nice!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
runningcode: ah so only `output` is created but not `sourceFile` ? that makes sense.

PR #1405, opened 2026-08-13
perf(instrumentation): [Generated metadata 2] Generate manifest metadata

romtsn:
Generate typed Sentry metadata from each variant final merged manifest in the app-owned build-time options class.

The SDK ASM transform remains parameterless and emits only a stable symbolic call. Resource-backed or otherwise unresolved values return null so the SDK retains its PackageManager fallback. The matching SDK consumer is getsentry/sentry-java#5969.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Performance
 
-- Generate resolved Sentry manifest metadata at build time, reducing SDK initialization time by about 6.5% in an all-options benchmark ([#1405](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1405))
+- Resolve Sentry manifest metadata at build time to reduce Android SDK initialization overhead ([#1405](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1405))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Avoid resolving the runtime classpath during Gradle configuration when SDK runtime optimizations are enabled ([#1404](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1404))
 
+### Performance
+
+- Generate resolved Sentry manifest metadata at build time, reducing SDK initialization time by about 6.5% in an all-options benchmark ([#1405](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1405))
+
 ### Dependencies
 
 - Bump ComposablePreviewScanner from v0.9.2 to v0.9.3 ([#1408](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1408))
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
@@ -1,0 +1,65 @@
+package io.sentry.android.gradle
+
+import java.io.File
+import javax.xml.parsers.DocumentBuilderFactory
+import org.w3c.dom.Element
+
+internal object ManifestMetadataParser {
+  fun parse(manifest: File): Map<String, Any>? =
+    runCatching {
+        val document =
+          manifest.inputStream().buffered().use {
+            DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(it)
+          }
+        val application = document.getElementsByTagName("application").item(0) ?: return emptyMap()
+        val metadata = linkedMapOf<String, Any>()
+
+        for (index in 0 until application.childNodes.length) {
+          val element = application.childNodes.item(index) as? Element ?: continue
+          if (element.tagName != "meta-data") continue
+
+          val name = element.getAttribute("android:name")
+          if (!name.startsWith("io.sentry.")) continue
+          val value = element.getAttribute("android:value")
+          if (
+            element.hasAttribute("android:resource") ||
+              !element.hasAttribute("android:value") ||
+              value.startsWith("@") ||
+              value.contains("${'$'}{")
+          ) {
+            SentryPlugin.logger.info(
+              "Sentry manifest metadata was not optimized because $name could not be resolved at build time."
+            )
+            return null
+          }
+          metadata[name] = inferType(value)
+        }
+        metadata
+      }
+      .onFailure {
+        SentryPlugin.logger.info(
+          "Sentry manifest metadata could not be parsed for optimization.",
+          it,
+        )
+      }
+      .getOrNull()
+
+  internal fun inferType(value: String): Any =
+    when (value) {
+      "true" -> true
+      "false" -> false
+      else -> parseInteger(value) ?: value.toFloatOrNull()?.takeIf { it.isFinite() } ?: value
+    }
+
+  private fun parseInteger(value: String): Int? =
+    if (value.matches(DECIMAL_INTEGER)) {
+      value.toIntOrNull()
+    } else if (value.matches(HEX_INTEGER)) {
+      value.removePrefix("+").let { Integer.decode(it) }
+    } else {
+      null
+    }
+
+  private val DECIMAL_INTEGER = Regex("[+-]?\\d+")
+  private val HEX_INTEGER = Regex("[+-]?0[xX][0-9a-fA-F]+")
+}
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
@@ -5,14 +5,14 @@ import javax.xml.parsers.DocumentBuilderFactory
 import org.w3c.dom.Element
 
 internal object ManifestMetadataParser {
-  fun parse(manifest: File): Map<String, Any>? =
+  fun parse(manifest: File): Map<String, String>? =
     runCatching {
         val document =
           manifest.inputStream().buffered().use {
             DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(it)
           }
         val application = document.getElementsByTagName("application").item(0) ?: return emptyMap()
-        val metadata = linkedMapOf<String, Any>()
+        val metadata = linkedMapOf<String, String>()
 
         for (index in 0 until application.childNodes.length) {
           val element = application.childNodes.item(index) as? Element ?: continue
@@ -32,7 +32,7 @@ internal object ManifestMetadataParser {
             )
             return null
           }
-          metadata[name] = inferType(value)
+          metadata[name] = value
         }
         metadata
       }
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
@@ -4,8 +4,10 @@ import io.sentry.android.gradle.ManifestMetadataParser
 import io.sentry.android.gradle.instrumentation.resolveClassAvailability
 import org.gradle.api.Project
 import org.gradle.api.UnknownDomainObjectException
+import org.gradle.api.file.RegularFile
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.internal.artifacts.DefaultModuleIdentifier
+import org.gradle.api.provider.Provider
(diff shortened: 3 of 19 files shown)


plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -27,6 +38,7 @@ abstract class GenerateSentryBuildTimeOptionsTask : DirectoryOutputTask() {
         }
         .toSet()
     val availability = resolveClassAvailability(modules)
+    val metadata = ManifestMetadataParser.parse(mergedManifest.get().asFile)
```
runningcode: if I understand the architecture correctly it seems like we have 3/4 concerns that i would separate this in to 3 tasks. this improves the cacheability and separates responsibilities:

1. resolving the class availability
2 parsing the manifest
3 merging the two / generating the source file

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -58,10 +86,37 @@ abstract class GenerateSentryBuildTimeOptionsTask : DirectoryOutputTask() {
     private const val GENERATED_CLASS_PATH =
       "io/sentry/android/core/SentryGeneratedBuildTimeOptions.java"
 
+    private fun String.javaStringLiteral(): String = buildString {
```
runningcode, line 113: fun! 😢

plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
```diff
@@ -0,0 +1,65 @@
...
+
+import java.io.File
+import javax.xml.parsers.DocumentBuilderFactory
+import org.w3c.dom.Element
+
+internal object ManifestMetadataParser {
+  fun parse(manifest: File): Map<String, Any>? =
+    runCatching {
+        val document =
+          manifest.inputStream().buffered().use {
```
runningcode, line 11: This whole class seems like something that is like 3 lines of groovy.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Avoid resolving the runtime classpath during Gradle configuration when SDK runtime optimizations are enabled ([#1404](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1404))
 
+### Performance
+
+- Generate resolved Sentry manifest metadata at build time, reducing SDK initialization time by about 6.5% in an all-options benchmark ([#1405](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1405))
```
runningcode: i think we shouldn't cite such precise specific performance percentages in our release notes

runningcode (commented):
looks great! main concern is the separate of responsibilities. and we should also add a configuration cache test for this!

CHANGELOG.md
romtsn: dropped!

plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
romtsn: True that, but I think groovy doesn't handle what we do here -- unresolved resources/placeholders, fallback behavior and type-inference (I think groovy would parse all values as strings). So I guess i'd keep the hand-rolled one

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/GenerateSentryBuildTimeOptionsTask.kt
romtsn: hm, I'm thinking if an intermediate file I/O and task scheduling that we would introduce by splitting this up would cost more than the work avoided by cacheability? It's map lookup + XML parsing which I guess should be rather quick?

romtsn:
> we should also add a configuration cache test for this!

thanks for the review - this one has been addressed in #1404 and should cover this too :)

plugin-build/src/main/kotlin/io/sentry/android/gradle/ManifestMetadataParser.kt
```diff
@@ -0,0 +1,65 @@
...
+      "true" -> true
+      "false" -> false
+      else -> parseInteger(value) ?: value.toFloatOrNull()?.takeIf { it.isFinite() } ?: value
+    }
+
+  private fun parseInteger(value: String): Int? =
+    if (value.matches(DECIMAL_INTEGER)) {
+      value.toIntOrNull()
+    } else if (value.matches(HEX_INTEGER)) {
+      value.removePrefix("+").let { Integer.decode(it) }
```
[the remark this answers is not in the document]
romtsn: This exception is handled by the outer runCatching and intentionally returns null, triggering the runtime Bundle fallback. Treating an overflowing hex literal as a String could differ from Android’s typed metadata and would be unsafe because the generated map is authoritative. Supporting unsigned 32-bit hex literals could be a future optimization, but this does not cause a runtime failure.

runningcode (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -0,0 +1,184 @@
...
+            '"' -> "\\\""
+            '\n' -> "\\n"
+            '\r' -> "\\r"
+            '\t' -> "\\t"
+            else -> character
+          }
+        )
+      }
+      append('"')
+    }
```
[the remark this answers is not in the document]
romtsn: This isn’t reachable: XML 1.0 rejects &#x08; during manifest parsing. Also, javac accepts other raw control characters inside string literals; the characters that require escaping are already handled. No change needed.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -20,24 +18,41 @@ import org.gradle.api.tasks.TaskProvider
...
-      moduleIds
-        .get()
-        .mapNotNull { coordinate ->
-          val parts = coordinate.split(':', limit = 2)
-          if (parts.size == 2) DefaultModuleIdentifier.newId(parts[0], parts[1]) else null
-        }
-        .toSet()
-    val availability = resolveClassAvailability(modules)
-    val metadata = ManifestMetadataParser.parse(mergedManifest.get().asFile)
+    val json = JsonSlurper()
```
runningcode, line 36: yess! the `JsonSlurper()`!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -20,24 +18,41 @@ import org.gradle.api.tasks.TaskProvider
...
+      } else {
+        runCatching {
+            (parsedMetadata as Map<*, *>)
+              .map { (key, value) ->
+                key as String to ManifestMetadataParser.inferType(value as String)
+              }
+              .toMap()
+          }
+          .onFailure {
+            logger.info(
```
runningcode, line 54: I wonder if we should just throw an exception here so at least we get it reported to us.
An exception that tells the users how to disable this feature so that their build will work next time.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
```diff
@@ -20,24 +18,41 @@ import org.gradle.api.tasks.TaskProvider
 @CacheableTask
 abstract class GenerateSentryBuildTimeOptionsTask : DirectoryOutputTask() {
```
runningcode, line 22: now that I think about it, should this just be a regular `DefaultTask` that outputs a file? this way we have a single file as an output (instead of a directory) and gradle will create the file's directory for us?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/ParseSentryManifestMetadataTask.kt
```diff
@@ -0,0 +1,32 @@
...
+  abstract val mergedManifest: RegularFileProperty
+
+  @get:OutputFile abstract val outputFile: RegularFileProperty
+
+  @TaskAction
+  fun parse() {
+    val metadata = ManifestMetadataParser.parse(mergedManifest.get().asFile)
+    val json = JsonOutput.toJson(metadata?.toSortedMap())
+    outputFile.get().asFile.apply {
+      parentFile.mkdirs()
```
runningcode: is this needed? can we go through this together? i feel like im misunderstanding something.

there's a real footgun caching issue here. since we're touching a directory outside of the declared outputs it could disable caching for other unrelated tasks with the message "gradle doesn't know how this file was created" or something like that.

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/ResolveSentryClassAvailabilityTask.kt
```diff
@@ -0,0 +1,37 @@
...
+      moduleIds
+        .get()
+        .mapNotNull { coordinate ->
+          val parts = coordinate.split(':', limit = 2)
+          if (parts.size == 2) DefaultModuleIdentifier.newId(parts[0], parts[1]) else null
+        }
+        .toSet()
+    val json = JsonOutput.toJson(resolveClassAvailability(modules).toSortedMap())
+    outputFile.get().asFile.apply {
+      parentFile.mkdirs()
```
runningcode: same here, why do we need this? i think im missing something

runningcode (commented):
left some comments below

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
romtsn: hm, not sure we can do that because we're using  AGP's `addGeneratedSourceDirectory` and this one requires an `@OutputDirectory`

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
romtsn: hm, I think I'd rather keep it as-is, because at runtime we'd still fallback to parsing the `Bundle` so there's only the perf regression that will have been introduced with this and breaking the build for a failed perf improvement sounds too harsh?

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
runningcode: ohh right. makes sense then!

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
runningcode: yeah, i just think we don't have good reporting mechanisms on errors but this is fine to me!

runningcode (approved):

plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/optimization/GenerateSentryBuildTimeOptionsTask.kt
romtsn: added reporting via SentryTelemetryService in https://github.com/getsentry/sentry-android-gradle-plugin/pull/1405/changes/64b4803c775936e23933d0ff738a52814b573b74

PR #1409, opened 2026-08-20
feat(instrumentation): allow disabling instrumentation for specific variants

### LINKED ISSUE #1407: Allow tracing instrumentation to be disabled per variant, independently of ignoredVariants / ignoredFlavors
### Problem Statement

`tracingInstrumentation.enabled` is a single project-wide property. There's no way to enable instrumentation for some variants and not others.

This matters because the instrumentation is registered with InstrumentationScope.ALL, so every dependency class is routed through the ASM transform and the transform's output. The practical effect is that incremental dex merging stops working, which is a significant hit to local build times on a large app.

Today the only variant-level control is ignoredVariants / ignoredFlavors / ignoredBuildTypes. The problem is that these settings are all-or-nothing: they also disable Proguard mapping upload, source context, size analysis and the rest of the plugin for that variant. There's no way to say "keep mapping upload on this variant, skip the bytecode transform."

### Solution Brainstorm

Variant filtering inside the tracingInstrumentation block, mirroring the existing top-level naming so the DSL stays consistent:

```
sentry {
  tracingInstrumentation {
    enabled = true
    ignoredVariants = setOf("devDebug", "internalDebug")
    // or ignoredFlavors / ignoredBuildTypes
  }
}
```

ansman:
## :scroll: Description

Add `ignoredVariants`, `ignoredBuildTypes`, and `ignoredFlavors` to both `tracingInstrumentation` and `runtimeOptimizations`.

```kotlin
sentry {
    tracingInstrumentation {
        ignoredBuildTypes.add("debug")
    }
    runtimeOptimizations {
        ignoredBuildTypes.add("debug")
    }
}
```

All exclusions default to empty, preserving existing behavior. Other plugin features, including mapping uploads, remain enabled for excluded variants.

Bytecode instrumentation disables incremental compilation, increasing build times during development. Existing variant exclusions disable the entire Sentry plugin, including unrelated functionality such as mapping uploads. This change allows instrumentation and runtime optimizations to be disabled independently for selected variants while preserving other plugin features and release behavior.

Fixes #1407.

```shell
./gradlew :plugin-build:spotlessApply :plugin-build:test :plugin-build:integrationTest --tests io.sentry.android.gradle.integration.SentryPluginVariantTest
```

All 429 unit tests and 13 variant integration tests pass. Integration coverage includes exclusions by variant, build type, and flavor; preserved mapping uploads; unaffected variants; and disabling both instrumentation visitors together.

I also tested it in our project and confirmed that the bytecode transforms did not run for disabled variants.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Decide if debug builds should be excluded by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Allow tracing instrumentation and runtime optimizations to be disabled for specific variants, build types, or flavors without disabling other Sentry plugin features ([#1409](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1409))
+
 ## 6.19.0
 
 ### Fixes
--- a/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
+++ b/plugin-build/src/main/kotlin/io/sentry/android/gradle/AndroidComponentsConfig.kt
@@ -179,8 +179,22 @@ fun ApplicationAndroidComponentsExtension.configure(
         }
       }
 
-      val runtimeOptimizationsEnabled = extension.runtimeOptimizations.enabled.get()
-      val tracingInstrumentationEnabled = extension.tracingInstrumentation.enabled.get()
+      val runtimeOptimizations = extension.runtimeOptimizations
+      val runtimeOptimizationsEnabled =
+        variant.isInstrumentationEnabled(
+          runtimeOptimizations.enabled.get(),
+          runtimeOptimizations.ignoredVariants,
+          runtimeOptimizations.ignoredBuildTypes,
+          runtimeOptimizations.ignoredFlavors,
+        )
+      val tracingInstrumentation = extension.tracingInstrumentation
+      val tracingInstrumentationEnabled =
+        variant.isInstrumentationEnabled(
+          tracingInstrumentation.enabled.get(),
+          tracingInstrumentation.ignoredVariants,
+          tracingInstrumentation.ignoredBuildTypes,
+          tracingInstrumentation.ignoredFlavors,
+        )
       // Both visitor factories need the resolved dependency graph.
       val modulesService =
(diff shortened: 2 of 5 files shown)


0xadam-brown (approved):
Backward compatible + a sane extension of our current API + worth exposing ASM off-switches. Lgtm – and thanks 💯

0xadam-brown:
...removed my approval as it looks like folks are discussing the precise API internally atm. We'll get back with you shortly...

romtsn:
hey @ansman thanks a bunch for your contribution -- we've discussed this internally and have come up with an inverse approach where you'd specify the variant name as the sub-DSL of `sentry` and configure each variant respectively (#1420):
```kts
sentry {
  variants {
    create("fullDebug") {
      tracingInstrumentation { enabled = false }
      runtimeOptimizations { enabled = false }
    }
    create("qaDebug") {
      tracingInstrumentation { enabled = false }
      runtimeOptimizations { enabled = false }
    }
  }
}

// or if you want a bit of sugar on top to disable it for all `debug` variants
androidComponents.beforeVariants(selector().withBuildType("debug")) { vb ->
  sentry.variants.maybeCreate(vb.name).with {
    tracingInstrumentation { enabled = false }
    runtimeOptimizations { enabled = false }
  }
}
```

let us know if it's not gonna work for your case, we can revisit it later! For now I'm gonna supersede it

romtsn:
oh and btw, it only exposes bytecode stuff now (`tracingInstrumentation` and `runtimeOptimizations` that is), let us know if you'd want more things to have configured that way.

ansman:
@romtsn That works fine too, though I'm not exactly sure what the sentry variants would be used for. I think doing something like this would feel more natural to me:
```kotlin
androidComponents.onVariants(selector().withBuildType("debug")) { variant ->
  variant.sentry {
    tracingInstrumentation { enabled = false }
    runtimeOptimizations { enabled = false }
  }
}
```

Personally I just wanted to disable the bytecode transforms because they prohibit incremental compilation.

PR #1413, opened 2026-08-21
build(develocity): Make build scan publishing opt-in locally

### LINKED ISSUE #1398: Please stop accepting build-scan ToS for everyone
### Description

As mentioned on https://docs.develocity.ai/gradle/4.5/gradle-plugin/#connecting-to-scans-gradle-com, you should

> Be careful not to commit agreement to the terms of service into a project that may be built by others.

Llike it is currently, private data of everyone cloning your sources and running any build is uploaded to the public Build Scan service without asking for their consent.
It is advisable to change the configuration so that it is not automatically agreed for anyone running the build, but for example only while running on CI.

runningcode:
## :scroll: Description
Fixes GRADLE-120. This doesn't accept TOS without opting in via a gradle flag.
CI builds will all accept and publish scans.

For local builds, you can add `gradleTOUAgree` to your `gradle.properties` in your Gradle home to opt in.

2 other changes

1. don't use background build scan publication on CI because it can get cut off by the worker
2. don't publish IP address

Fixes GRADLE-120 (https://linear.app/getsentry/issue/GRADLE-120/please-stop-accepting-build-scan-tos-for-everyone)
Fixes #1398

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -15,10 +15,21 @@ plugins {
   id("com.gradle.common-custom-user-data-gradle-plugin") version "2.8.0"
 }
 
+val isCI = System.getenv("CI") != null
+val gradleTOUAgree = "true" == providers.gradleProperty("gradleTOUAgree").orNull
+
 develocity {
   buildScan {
     termsOfUseUrl.set("https://gradle.com/help/legal-terms-of-use")
-    termsOfUseAgree.set("yes")
+    termsOfUseAgree.set(if (isCI || gradleTOUAgree) "yes" else "no")
+    uploadInBackground = !isCI
+    // we can't pass the condition directly inside the onlyIf because of a CC bug in Gradle 8.X (this is fixed in 9.X)
+    if (!isCI && !gradleTOUAgree) {
+      publishing.onlyIf { false }
+    }
+    obfuscation {
+      ipAddresses { addresses -> addresses.map { _ -> "0.0.0.0" } }
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


settings.gradle.kts
```diff
@@ -15,10 +15,21 @@ plugins {
   id("com.gradle.common-custom-user-data-gradle-plugin") version "2.8.0"
 }
 
+val isCI = System.getenv("CI") != null
```
0xadam-brown, line 18: m: Looks like `isCi` will be true if my env has `CI=false`

settings.gradle.kts
```diff
@@ -30,7 +41,7 @@ dependencyResolutionManagement {
   repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
 }
 
-rootProject.name = ("sentry-android-gradle-plugin-composite-build")
+rootProject.name = "sentry-android-gradle-plugin-composite-build"
 
 include(":examples:android-gradle")
```
0xadam-brown, line 47: Worth a CHANGELOG entry? 

Also, AFAICT, it looks like before this PR I could run `./gradlew ... --scan` and a build scan would be published, but after I have to run `./gradlew --scan -PgradleTOUAgree=true` (or be on CI) for the scan to be published. If that's right, worth including a note in the CHANGELOG entry.

0xadam-brown (commented):
Thx for this 💯! A couple quick comments.

settings.gradle.kts
runningcode: no i don't think this is worth a changelog because this doesn't affect consumers of our Gradle plugin

settings.gradle.kts
runningcode: yes but our CI is only github actions so this always works.

0xadam-brown (approved):

PR #1433, opened 2026-09-01
build(verification): Verify plugin-build dependencies with PGP

runningcode:
## What

Switches `plugin-build` from per-artifact SHA-256 checksums to PGP key trust, the scheme
`sentry-kotlin-compiler-plugin` already uses and recommended in [this article](https://www.benediktritter.de/blog/2026/06/01/gradle-dependency-verification-best-practices/).

Also adds a job to enforce this by running `plugin-build` to verify dependencies on CI.

The rest of the cleanup is just fixing the docs now that we use the same PGP key trust on both the `sentry-kotlin-compiler-plugin` and the `plugin-build`. One important thing to note is that the `plugin-build` has a lockfile for while the `sentry-kotlin-compiler-plugin` has `failOnNonReproducibleBuild`. I can make that more consistent in a follow up.


.github/workflows/integration-tests-sentry-cli.yml
```diff
@@ -39,6 +39,4 @@ jobs:
       - name: Start server and run integration test for sentry-cli commands
         run: |
           test/integration-test-server-start.sh &
-          # Verification is disabled here because the standalone `-p plugin-build` build
-          # resolves buildscript-classpath artifacts not captured in the metadata.
-          ./gradlew -p plugin-build integrationTest --tests "*IntegrationTest" --dependency-verification=off
+          ./gradlew -p plugin-build integrationTest --tests "*IntegrationTest"
```
runningcode, line 42: this is just exactly as this command was before introducing the dependency verification so I removed the comment. we don't need it now that we trust the authors PGP signature.

.github/workflows/test-matrix-agp-gradle.yaml
```diff
@@ -71,9 +71,7 @@ jobs:
           cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
 
       - name: Build the Release variants and integrationTest
```
runningcode, line 73: again, this is exactly how this was before we added dependency verification so I removed the comment. we don't need it now that we trust the PGP key.

plugin-build/build.gradle.kts
```diff
@@ -84,14 +84,20 @@ if (!isVersionOverrideBuild) {
   dependencyLocking { lockAllConfigurations() }
 }
 
-tasks.register("resolveAndLockAll") {
+// Resolves every configuration so verification covers the whole graph, not just what the requested
+// tasks need. CI verifies through this without writing anything; mirrors the compiler plugin's.
+tasks.register("resolveAll") {
```
runningcode, line 89: this is the same thing as before just split in to two different tasks.

scripts/relock-plugin-build.sh
```diff
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
```
runningcode, line 1: i'm thinking of removing this script since its just a single line of gradle.

0xadam-brown (approved):
Thanks for the consistency win (among others) 💯

PR #1430, opened 2026-09-01
build(verification): Run compiler plugin verification in CI (GRADLE-116)

runningcode:
Follow-ups to #1423.

1. Fix spotless which was after the PR was created but after it was merged.
2. Add a `verify-compiler-plugin-dependencies` job that exercises the build with the dependency verification.
3. Narrow ktfmt key from all of `com.facebook` to `ktfmt` and `ktfmt-parent`
4. Add changelog entry

### Verification

Should pass CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1423, opened 2026-08-27
build(verification): Verify compiler plugin dependencies with PGP (GRADLE-116)

runningcode:
Closes GRADLE-116. Fixes #1365, fixes #1366.

`sentry-kotlin-compiler-plugin` resolved its dependencies with no verification at all. This enables it, following [Benedikt Ritter's best practices](https://www.benediktritter.de/blog/2026/06/01/gradle-dependency-verification-best-practices/): PGP signatures rather than checksums, so a routine version bump leaves the metadata untouched and a diff here is rare enough to get read.

- 37 trusted keys; 6 SHA-256 checksums left as fallback for artifacts nobody signs (Plugin Portal marker POMs, a few parent POMs)
- Armored keyring so it diffs as text, key servers disabled
- `failOnNonReproducibleResolution()` — the graph was already fully pinned

Regeneration is a no-op against the committed file:

```bash
./gradlew -p sentry-kotlin-compiler-plugin resolveAll spotlessCheck \
  --write-verification-metadata pgp,sha256 --export-keys
```

`spotlessCheck` is in that command because Spotless resolves ktfmt itself, as a task input — it lands in no declared configuration, so `resolveAll` alone leaves those 8 artifacts out of the metadata.

### Worth a look in review

Two scopes are deliberately narrower than what Gradle's bootstrap infers. Regeneration preserves them; the file header says not to widen them back.

| Key | Gradle infers | Narrowed to |
| --- | --- | --- |
| JetBrains Compose | all of `org.jetbrains` — would vouch for Kotlin itself | compose groups, `androidx.lifecycle`, `skiko` |
| Error Prone | all of `com.google` — would cover Guava, Gson, AutoService | `errorprone`, `googlejavaformat` |

Verification is scoped to the build tree root, so an included build's own config is ignored — `preMerge` does *not* verify this build. No PR job exercises it; the first one that does is the snapshot publish, which builds the compiler plugin from its own directory.

`plugin-build`'s existing checksum metadata is untouched; converting it the same way can be a follow-up.

Build-only change: #skip-changelog

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1434, opened 2026-09-02
build(release): Generate the changelog automatically with Craft

runningcode:
## :scroll: Description

Craft has been configured with `changelogPolicy: auto` for years, but auto mode only generates a section when the `## Unreleased` section is empty. Every PR hand-wrote its entry there (enforced by Danger), so the generation path never actually ran.

This switches us to the flow [sentry-dart](https://github.com/getsentry/sentry-dart/blob/main/.github/release.yml) uses: **the PR title is the changelog entry**, and Craft categorizes it at release time.

- **`.github/release.yml`** (new) — categories matching the sections this changelog already uses (Breaking Changes / Security / Features / Fixes / Performance / Dependencies / Internal Changes), each with a `semver` field so `craft prepare auto` can derive the version bump. Renovate and Dependabot are excluded, since Renovate only bumps this repo's own build tooling; the bumps users care about (CLI, Android SDK, ComposablePreviewScanner) come from `getsentry/github-workflows/updater` as `github-actions[bot]` and land under Dependencies.
- **`.craft.yml`** — migrate off the deprecated `changelogPolicy` key to `changelog.policy`, and raise `minVersion` to `2.30.1`, which is exactly the craft SHA both workflows already pin. Craft 2.21+ bumps versions from the publish targets when no `preReleaseCommand` is set, and the `maven` target can't bump a Gradle build, so `scripts/bump-version.sh` is now named explicitly rather than relied on as a default.
- **`update-deps.yml`** — `changelog-entry: false` on all three updaters, and **`CHANGELOG.md`** — drop the stale `## Unreleased` section. Either one left in place keeps that section non-empty and silently suppresses generation for the next release. Those four entries are regenerated from their PR titles.
- **`danger.yml`** — removed. Its changelog rule hard-fails any `feat:`/`fix:`/`perf:`/`security:` PR without a manual entry, and the shared action has no input to disable that one check.

Hand-written entries are a recurring source of merge conflicts, rebase fixups (`docs(changelog): Add missing entry…`, `chore(changelog): …`), and PRs that get blocked on a one-line edit.

Replayed the last 60 merged PRs through Craft's own categorization logic with this config. Sample of the generated output:

```
### Features
- (instrumentation) Add reverse variant DSL for feature overrides
- (snapshots) Emit canvas_theme in snapshot sidecar (EME-1241)

### Fixes
- (proguard) Unwrap analytics variant for optimization

### Performance
#### Instrumentation
- [Generated metadata 2] Generate manifest metadata
- [Generated metadata 1] Generate class availability

### Dependencies
- Update CLI to v3.7.0
- Update Android SDK to v8.54.0
```

20 of the 60 were excluded (all Renovate), and 2 landed in Craft's catch-all `Other` section because their titles had no conventional-commit prefix — nothing is dropped silently.

Also verified against craft's source at the pinned SHA `cd1e829` that it is exactly `2.30.1`, that its schema accepts the `changelog` object form, and that `preReleaseCommand` still routes to `scripts/bump-version.sh`.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

- Removing Danger also drops its "pin actions to a commit SHA" check. Worth restoring separately with [zizmor](https://github.com/zizmorcore/zizmor) or `actionlint` if we want that enforced.
- #1423 / #1430 will land under *Internal Changes* rather than *Security* next release, since they're `build(verification):` commits. Applying the existing `Security` label to them puts them back — label matching takes precedence over the title pattern.


0xadam-brown (approved):
Lgtm 👍 

My clanker pointed out that the following from `renovate.json` has now gone stale and can be removed:

```
{
      "description": "Deps bundled into the published plugin are user-facing, so use a fix: title to make Danger require a CHANGELOG entry. Everything else stays build(deps): and is skipped.",
      "matchPackageNames": [
        "org.ow2.asm:asm-util",
        "org.ow2.asm:asm-commons"
      ],
      "semanticCommitType": "fix"
    }
```

PR #1442, opened 2026-09-09
chore(deps): align bundled ASM to 9.8

0xadam-brown:
## :scroll: Description
<!--- Describe your changes in detail -->

Bumps the plugin's own asm/asm-util/asm-commons implementation deps from 9.4 to 9.8 so runtimeClasspath matches the 9.8 already resolved on compileClasspath via AGP 8.13.2. ASM api-versioning is backward compatible and instrumentation visitors use the AGP-supplied apiVersion, so a newer bundled ASM only widens the supported range.


<!--- Why is this change required? What problem does it solve? -->
<!--- If it fixes an open issue, please link to the issue here. -->

Eliminates a mismatch where we use 9.4 at runtime but compile against a later version and potentially use APIs the earlier version doesn't support. We hadn't done that yet; this helps prevent us from doing so going forward...

...for now, at any rate. They could come unglued again. I have another branch with a Gradle task that ensures they stay in lockstep, but it seemed a bit heavy-handed. Chime in if you think otherwise and I can post a follow-on PR.

Ran existing regression tests. 

And had my clanker build the sentry-java Android sample app using SAGP with these changes and run some golden flows:

Bytecode proof (dexdump of the shipped APK). Each key auto-instrumentation actually injected its hook:

| Feature | Target class | Injected edge |                                                                                                                                                                                                                                                                                                                                                                                 
|---|---|---|                                                                                                                                                                                                                                                                                                                                                                                                              
| OKHTTP | `okhttp3.RealCall` / `OkHttpClient` | `new SentryOkHttpInterceptor`, `new SentryOkHttpEventListener` |                                                                                                                                                                                                                                                                                                          
| DATABASE | `FrameworkSQLiteOpenHelperFactory`, `AutoClosingRoomOpenHelperFactory` | `invoke-static SentrySupportSQLiteOpenHelper.create(...)` |                                                                                                                                                                                                                                                                          
| COMPOSE | `androidx.navigation.compose.NavHostControllerKt.rememberNavController` | `invoke-static SentryNavigationIntegrationKt.withSentryObservableEffect(...)` |   

## :pencil: Checklist
<!--- Put an `x` in the boxes that apply -->
- [x] No breaking changes


## :crystal_ball: Next steps

#skip-changelog


runningcode:
Yeah I agree that we should bump this, however the issue is that 9.8 might not be compatible with the AGP version that consumers are using since we support AGP 7.4. This change could silently bump consumers using AGP 7.4 or even AGP 8.9. We don't even run any tests against AGP 7.4.

See the discussion here: https://github.com/getsentry/sentry-android-gradle-plugin/pull/1282#issuecomment-4659358892

0xadam-brown:
> Yeah I agree that we should bump this, however the issue is that 9.8 might not be compatible with the AGP version that consumers are using since we support AGP 7.4. This change could silently bump consumers using AGP 7.4 or even AGP 8.9. We don't even run any tests against AGP 7.4.
> 
> See the discussion here: [#1282 (comment)](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1282#issuecomment-4659358892)

Ah, thanks much for that context. I'll close then 👍

PR #1452, opened 2026-09-18
feat(telemetry): Track Kotlin Gradle plugin version

runningcode:
Adds a `KGP_VERSION` tag to build telemetry so we can see which Kotlin Gradle plugin versions our users are on, alongside the existing `AGP_VERSION` and `GRADLE_VERSION` tags.

### Verification

KGP is `compileOnly` and so absent from the test classpath, which makes the unit test exactly the "no Kotlin plugin" case — it fails with `NoClassDefFoundError` if the catch is ever narrowed to `Exception`.

### Drive-by cleanups

- Removed the unused `cliExecutable` val in `SentryJvmPlugin`
- Renamed `endTask`'s parameter to `nullableSpan` to silence a name-shadowing warning

### Note

Telemetry starts from `gradle.taskGraph.whenReady`, and those callbacks do not run on a configuration-cache hit, so this tag — like all the others — is only reported from builds that actually ran configuration. Pre-existing, and worth knowing when reading the resulting data.


plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -231,8 +233,8 @@ abstract class SentryTelemetryService : BuildService<None>, BuildOperationListen
     }
   }
 
-  fun endTask(span: ISpan?, task: Task) {
-    span?.let { span ->
+  fun endTask(nullableSpan: ISpan?, task: Task) {
```
runningcode, line 236: This is completely unrelated but just fixes the `span` being shadowed here.

plugin-build/src/main/kotlin/io/sentry/jvm/gradle/SentryJvmPlugin.kt
```diff
@@ -49,7 +48,6 @@ constructor(private val buildEvents: BuildEventListenerRegistryInternal) : Plugi
 
       val javaVariant = JavaVariant(project, javaExtension)
       val outputPaths = OutputPaths(project, "java")
-      val cliExecutable = project.cliExecutableProvider()
```
runningcode, line 52: Also unrelated cleanup of an unused variable.

plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt
```diff
@@ -74,6 +74,6 @@ constructor(private val buildEvents: BuildEventListenerRegistryInternal) : Plugi
     internal const val SENTRY_SDK_VERSION = BuildConfig.SdkVersion
 
     // a single unified logger used by instrumentation
-    internal val logger by lazy { LoggerFactory.getLogger(SentryPlugin::class.java) }
+    internal val logger by lazy { Logging.getLogger(SentryPlugin::class.java) }
```
runningcode, line 77: This fixes our logger to use the proper Gradle logging API rather than slf4j.

plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt
```diff
@@ -92,6 +93,7 @@ abstract class SentryTelemetryService : BuildService<None>, BuildOperationListen
             try {
               options.setTag("AGP_VERSION", AgpVersions.CURRENT.toString())
             } catch (t: Throwable) {}
```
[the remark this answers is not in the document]
runningcode: nope. there's no documentation that we need to handle this on the `kotlinPluginVersion` method.

markushi (approved):
Nice one!

PR #1453, opened 2026-09-18
fix(agp): Stop warning when AGP drops the optimization accessor

runningcode:
## Background
`CanMinifyCode.isMinifyEnabled` is what we usually check to determine if minification is enabled.
However, in some versions that flag was not set correctly which is why we check `com.android.build.api.component.impl.features.OptimizationCreationConfigImpl.getApplicationOptimizationEnabled()` instead.

## What did Google change on us?
Google decided to fix their issue and correctly report `CanMinifyCode.isMinifyEnabled`. So now we can trust it. The issue is that they removed `.getApplicationOptimizationEnabled()` at the same time. They did this in patch versions too. So far I've noticed that 9.5.0-alpha06 and 9.4.1 and 9.3.3 are affected.

Here's a little chart my clanker came up with:

| accessor | `codeShrinkerEnabled()`, which backs `isMinifyEnabled` |
|---|---|
| present | **excludes** `optimization { enable }` → reflection needed |
| absent | **includes** it → public API is authoritative |

## The fix
We just catch the `NoSuchMethodException`

Also, `Test Matrix - AGP 9.5.0-alpha06` job has bee red on `main` since the 2026-09-18 nightly and a customer reported this: https://github.com/getsentry/sentry-android-gradle-plugin/issues/1455


plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP74Compat.kt
```diff
@@ -112,16 +112,22 @@ private fun Variant.isApplicationOptimizationEnabled(): Boolean {
...
       unwrappedVariant.javaClass.getMethod("getOptimizationCreationConfig").invoke(unwrappedVariant)
     optimizationCreationConfig.javaClass
       .getMethod("getApplicationOptimizationEnabled")
       .invoke(optimizationCreationConfig) as Boolean
+  } catch (_: NoSuchMethodException) {
+    // AGP 9.3.3, 9.4.1 and 9.5.0-alpha06 folded optimization.enable into the code shrinker flag
+    // behind CanMinifyCode.isMinifyEnabled and dropped this accessor. The method is missing only
+    // when CanMinifyCode.isMinifyEnabled includes that flag so we can trust it. Patch releases
+    // mid-line, hence probing for the method rather than checking a version.
+    false
```
0xadam-brown, line 130: Worth a test? (I'll defer to your judgment...)

0xadam-brown (approved):
An optional comment; otherwise lgtm 👍

plugin-build/src/main/kotlin/io/sentry/android/gradle/AGP74Compat.kt
runningcode: yes our test matrix was already failing because of this so we're covered. here's a link to the build scan of the failing test: https://scans.gradle.com/s/l75op5t5z5lbu/tests/task/:plugin-build:integrationTest/details/io.sentry.android.gradle.integration.SentryPluginTest/creates%20proguard%20mapping%20tasks%20when%20app%20optimization%20is%20enabled?focused-exception-line=0-11&top-execution=1

